PR #16, opened 2015-09-01
Fix some spelling in README.md

fahhem:


jettisonjoe:
Thanks for the contribution! Just a heads-up, you might want to avoid any actual substantial pull requests for the next week or so while we finish the in-progress changes in the 'refactor' branch and merge them into master. This will include some fairly major changes like separating declarative from executive data, renaming "parameters" to "measurements", etc.

PR #23, opened 2015-09-15
Adding logger and unittest

amyxchen:


openhtf/__init__.py
```diff
@@ -23,9 +23,11 @@
 import signal
 import socket
 import sys
-
+import pdb
```
amyxchen: Forget to remove this, remove later

openhtf/phasemanager.py
```diff
@@ -44,17 +44,9 @@
...
-
-_LOG = logging.getLogger('htf.phasemanager')
-
-# Only use 'is' checks, as that does pointer comparison for strings. That makes
-# this the same as an object(), but useful when printed.
-DIDNT_FINISH = 'DIDNT_FINISH'
-
-configuration.Declare(
-    'blacklist_phases', 'Phase names to skip', default_value=[])
+                      'blacklist_phases', 'Phase names to skip', default_value=[])
```
amyxchen: The first merge was not right, I have fixed my merge, forget to push to my fork afterwards.

openhtf/user_input.py
```diff
@@ -87,6 +87,7 @@ def DisplayPrompt(self, message, text_input=False):
       console_prompt.Stop()
       self._prompt = None
       if self._response is None:
+        self._prompt = None
```
amyxchen: this line seems not right,  not in the original, I will remove it.

openhtf/util/unittest_util.py
```diff
@@ -0,0 +1,87 @@
...
+
+  @classmethod
+  def tearDownClass(cls):
+    print "tearDown done"
+
+  """
+  @mock.patch('configuration.ConfigModel.loaded')
+  def mock_loaded(self, mock_property):
+    mock_property.return_value = true
+  """
```
amyxchen: remove the comments here.

openhtf/__init__.py
```diff
@@ -37,11 +38,35 @@
 
 FLAGS = gflags.FLAGS
 FLAGS(sys.argv)
-
+gflags.DEFINE_string('log_dir', '/tmp', 'dir for log files')
+gflags.DEFINE_enum('log_output', 'console', ['console','file','both'], 'output option')
```
jettisonjoe: For controlling stdout verbosity I suggest using the -v verbosity convention where:
default: send only critical and error level messages to stdout
-v: send warning and above level messages to stdout
-vv: send info and above level messages to stdout
-vvv: send everything to stdout

openhtf/__init__.py
```diff
@@ -37,11 +38,35 @@
...
+gflags.DEFINE_enum('level', 'info', ['debug', 'info', 'warning', 'error', 'critical'], 'log level')
```
jettisonjoe: Consider calling this something like ''log_level' instead of just 'level' for more clarity.

openhtf/__init__.py
```diff
@@ -37,11 +38,35 @@
...
+
+def Setup_logger():
+    htflogger = logging.getLogger('HTFTest')
```
jettisonjoe: I think we're going to want to use the htflogger (possibly with a re-name for clarity) so we get log messages in our test record in addition to the normal logfile and stdout/stderror.

openhtf/unittest_openhtf.py
```diff
@@ -0,0 +1,122 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
```
jettisonjoe: I think we should consider creating a directory 'openhtf/test' and putting our unittest stuff in there.

openhtf/unittest_openhtf.py
```diff
@@ -0,0 +1,122 @@
...
+  test_result = unittest.TextTestRunner().run(suite)
+  now =  datetime.datetime.now().strftime("%m-%d-%d_%H:%M:%S")
+  classname = os.path.dirname(os.path.abspath(__file__)).split('/')[-1]
+
+  log = TestOpenhtf.testlog_dir + '/'+ classname.strip() + '_' + now
+
+  with open(log, 'w') as f:
+    f.write("====== Test Result =====\n")
+    f.write(str(test_result))
+    f.write("\n\n====== FAILE ======\n")
```
jettisonjoe: Should this be 'FAIL' rather than 'FAILE'?

openhtf/__init__.py
```diff
@@ -37,11 +38,35 @@
...
+
+    cur_time = datetime.utcnow().strftime('%Y-%m-%d-%H:%M:%S.%f')[:-3]
+    fh = logging.FileHandler(FLAGS.log_dir + '/test_' + cur_time +'.log', delay=True)
+    fh.setFormatter(formatter)
+    htflogger.addHandler(fh)
+
+    sh = logging.StreamHandler()
+    sh.setFormatter(formatter)
+    htflogger.addHandler(sh)
+
```
jettisonjoe: Add test_record handler (is currently in util/htflogger, but probably move it to test_record).

openhtf/__init__.py
```diff
@@ -201,7 +226,7 @@ def PhaseTwo(test):
                     FLAGS.http_port,
                     os.getpid()).SaveToFile(FLAGS.rundir)
   
-    logging.info('Executing test: %s', self.filename)
+    _LOG.info('Executing test: %s', self.filename)
```
jettisonjoe: Now that I'm looking at this, it should probably be something more like "Starting test execution" since this log line applies to multiple test runs.

openhtf/capabilities/usb/adb_message.py
```diff
@@ -54,8 +54,7 @@
                       'Set to True to save all incoming and outgoing '
                       'AdbMessages and print them on Close().')
 FLAGS = gflags.FLAGS
-_LOG = logging.getLogger('adb_protocol')
```
grybmadsci: Let's all chat about the various sources of logs we have and establish a taxonomy of sorts for them, I think it's more complicated than simply changing everything to use the same 'HTFTest' logger.  Notably, some logs go into the TestRecord and some don't, so at the very least we need to differentiate those.

openhtf/util/configuration.py
```diff
@@ -46,6 +46,7 @@
     'This value will override any existing config value at config load time '
     'and will be a string')
 
+_LOG = logging.getLogger('HTFTest')
```
grybmadsci: Things in util/ aren't really core framework, and probably should use either a different logger, or just the root logger.  That is, util/configuration.py is actually a standalone utility file, not intended for use only by OpenHTF (though it's unlikely it'll actually be used by anything else).

openhtf/util/log_persister.py
```diff
@@ -27,6 +27,7 @@
 gflags.DEFINE_multistring('output_dir', ['./'],
```
grybmadsci: I'm fairly sure this entire file is deprecated.

openhtf/testrunadapter.py
```diff
@@ -35,7 +35,7 @@
 from openhtf.util import htflogger
```
grybmadsci: this file is deprecated and should just be removed.

openhtf/phasemanager.py
```diff
@@ -139,7 +139,8 @@ def __init__(self, cell_config, test, test_state, capabilities):
     self._config = cell_config
     self._phases = list(test.phases)
     self._test_state = test_state
-    self._logger = test_state.logger
```
grybmadsci: This is a change we should discuss in person - by not using the test_state.logger, you're changing the behavior so that logs here don't get output to the test record anymore.  We should do an audit and decide which log messages make sense to go to the test record and which should just be console output.  Or perhaps we have a flag to set the log level that gets output to the test record?

PR #27, opened 2015-09-23
Switch to using mutablerecords for TestRecord and ConfigurationDeclation

fahhem:


openhtf/io/test_record.py
```diff
@@ -26,38 +29,19 @@ class InvalidMeasurementDimensions(Exception):
...
-        {}, None, test_code,       # metadata, version, code
-        [], [])                    # phases and log_records
-    self.metadata['filename'] = test_filename
-    self.metadata['docstring'] = docstring
-    return self
-
-  @classmethod
-  def FromFrame(cls, frame):
-    """Create a TestRecord, initialized from the given stack frame.
+TestMetadata = mutablerecords.Record(
```
grybmadsci: This was very intentionally a dict and not a fixed data structure, and there was a lot of in-person design discussion around that (and many other) decision(s).  Happy to have additional input and more discussion around these, but please bring up nontrivial changes like this rather than just submitting a pull request containing it (where it could easily get overlooked among all the other changes).

openhtf/conf.py
```diff
@@ -46,6 +47,12 @@
...
     'and will be a string')
 
+ConfigurationDeclaration = (
+    mutablerecords.Record(
+        'ConfigurationDeclaration',
+        ['name'],
+        {'description': None, 'default_value': None, 'optional': True}))
+
 
 class ConfigurationDeclarationError(Exception):
```
grybmadsci: Does this get raised anywhere else in this file? If not, let's just get ride of the exception type too.  I see you've removed at least one place it was raised (and the associated check went away in the process, not sure that particularly matters, but it is a notable change that you can now have non-optional configuration keys with default values).

openhtf/exe/test_state.py
```diff
@@ -116,17 +116,13 @@ def GetFinishedRecord(self):
...
-    else:
-      self.record.metadata['config'] = conf.Config()
-
     if not self.record.dut_id:
       raise BlankDutIdError(
           'Blank or missing DUT ID, HTF requires a non-blank ID.')
 
-    return self.record._replace(
-        end_time_millis=util.TimeMillis(), outcome=self._state)
+    self.record.end_time_millis = util.TimeMillis()
```
grybmadsci: I was a little concerned about this convention in the first place - GetFinishedRecord() was only ever intended to be called once at the end of the lifetime of this instance, though it's not enforced.   At the time, self.record was intentionally not updated to reflect the end_time and outcome to avoid any weird behavior where someone called GetFinishedRecord() and then went back and continued doing something that used self.record (which could result, for example, in things like having log lines past the end_time_millis of the test_record).

Not sure what the best answer here is, really, but it's worth a conversation, I think.  My thoughts are that we should set self.record to None when GetFinishedRecord() is called (and maybe rename that to something like FinalizeRecord()), and then do some checks for self.record being None and raise a specific exception.  That way we can catch as soon as possible if something wonky happens in the usage of a TestState instance.  Thoughts?

openhtf/io/test_record.py
fahhem: This seemed pretty trivial, and can easily be undone in this PR by replacing 'TestMetadata' below with 'dict'

openhtf/exe/test_state.py
fahhem: I'd do something like this:

FinalizeRecord() -> sets end_time_millis, raises if it was already set
GetFinalRecord() -> returns record, raises if end_time_millis is not set

The TestExecutor would then Finalize then Get the record

openhtf/conf.py
fahhem: It's not used elsewhere, I'll knock it out of this one.

That brings up something I found recently, voluptuous, another data-validation library (I can find them with ease now for some reason). It is really easy to specify schemas and it even supports this sort of mutually-exclusive set of keys in a dict. I'm considering adding support for it as a replacement for the plain list and dict here.

openhtf/exe/test_state.py
jettisonjoe: I like having record populated while the test is running because then it can be repeatedly serialized and served by the frontend API instead of having the frontend API build all the same stuff manually. Why not just have GetFinishedRecord() raise if end_time_millis isn't already set?

jettisonjoe:
Resolved all conflicts and merged, because I need these mutablerecords in order to be able to populate and subsequently update the list of phases for the frontend API.

PR #35, opened 2015-10-17
Frontend and frontend API refactor.

jettisonjoe:


jettisonjoe:
I should probably go ahead and clean up the following to reflect that there's no more huge npm-dependent frontend:
  * setup.py
  * requirements.txt
  * CONTRIBUTING.md

jettisonjoe:
I think we can remote the protos and proto build logic as well.

PR #45, opened 2015-10-22
unit test and class string

amyxchen:


test/run_unittests.py
```diff
@@ -47,16 +46,15 @@ def main(argv):
   passing = True
 
   for testdir in testdirs:
-    if testdir != "" and testdir.strip()[0].isalpha():
-      print " -------- test dir: %s --------" % testdir.strip()
-      mod = __import__(testdir.strip(), fromlist=['testcase_runner'])
+    test = testdir.strip()
+    if test != "" and test[0].isalpha():
```
jettisonjoe: We should probably homogenize these to single quotes instead of double as a style thing.

PR #47, opened 2015-10-30
Add serail plug and updated requirement.txt to add required pyserial …

MasterChiefV:
…package


jettisonjoe:
Thanks for the PR. Will review and comment soon.

openhtf/plugs/uart_device.py
```diff
@@ -0,0 +1,134 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""SerialPlug used for serial communication
+Requires the following keys in .yaml configuration file AND declared.
+
+.yaml file
```
jettisonjoe: Since config can be hierarchical, perhaps we should establish a convention here. For example:

plugs:
    <plug name>:
        <config key 1>
        <config key 2>
        ...

Any thoughts from the other openhtf devs?

openhtf/plugs/uart_device.py
```diff
@@ -0,0 +1,134 @@
...
+  @conf.InjectPositionalArgs
+  def __init__(self):
+    logging.info('Instantiating %s with these settings:' % (type(self).__name__))
+    logging.info(' port: %s' %conf.Config().serial_port)
+    logging.info(' baudrate: %s' %conf.Config().serial_baudrate)
+    logging.info(' parity: %s' %conf.Config().serial_parity)
+    logging.info(' stopbits: %s' %conf.Config().serial_stopbits)
+    logging.info(' bytesize: %s' %conf.Config().serial_bytesize)
+    logging.info(' rtscts: %s' %conf.Config().serial_rtscts)
+    logging.info(' timeout: %s' %conf.Config().serial_timeout)
```
jettisonjoe: These look like maybe they should be debug-level lines to me.

openhtf/plugs/uart_device.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return out
+
+  def TearDown(self):
+    # This method is optional.  If implemented, it will be called at the end
+    # of the test.
+    logging.info('Tearing down %s!' % type(self).__name__)
+    if (self._portOpened):
+        logging.info('closing serial port!!!')
+        self._portHandler.close()
```
jettisonjoe: Should have a newline at end of file.

openhtf/plugs/uart_device.py
```diff
@@ -0,0 +1,134 @@
...
+      return False
+
+  def SendRawCmd(self, cmd):
+    # convert ascii cmd to raw bytes, send over uart, and convert raw byte results back to ascii for display
+    logging.info('serial input: ' + cmd)
+    inputbinary = binascii.unhexlify(cmd)
+    self._portHandler.write(inputbinary)
+    out = ''
+
+    time.sleep(0.05)
```
jettisonjoe: Is sleeping the recommended way to do this? Seems a little unclean to me.

jettisonjoe:
A general comment: Pretty much all of your classes and methods should have docstrings, not just the module itself.

openhtf/plugs/uart_device.py
MasterChiefV: Strongly agree.  In fact I tried this earlier but couldn't get it to work in the hierarchical format.  Got it working now. 

Will do the following:
plug name1:
  config key1

plug name2:
 config key1
 config key2

openhtf/plugs/uart_device.py
```diff
@@ -0,0 +1,134 @@
...
+
+declared in your project
+  conf.Declare('dut_interface', 'specify interface type: serial or adb for now')
+  conf.Declare('serial_port', 'port name of serial device. OS dependent')
+  conf.Declare('serial_baudrate')
+  conf.Declare('serial_parity')
+  conf.Declare('serial_stopbits')
+  conf.Declare('serial_bytesize')
+  conf.Declare('serial_rtscts')
+  conf.Declare('serial_timeout')
```
MasterChiefV: done.

openhtf/plugs/uart_device.py
MasterChiefV: Started with this way from examples I found online.
I am switching to polling with a timeout in the next version so a fixed delay can be avoided.

openhtf/plugs/uart_device.py
fahhem: Yes. I agree on the config format. I'd say the framework should pass them in from the right place, too, but we can do that in a separate commit.

jettisonjoe:
See the [google python style guide on 'Comments'](https://google.github.io/styleguide/pyguide.html#Comments) for doc string formatting guidelines.

openhtf/plugs/uart.py
```diff
@@ -0,0 +1,295 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
```
jettisonjoe: I would consider making the filename match the plug name. If it's SerialPlug, maybe it should be serial.py.

openhtf/plugs/uart.py
```diff
@@ -0,0 +1,295 @@
...
+class SerialPlugExceptions(Exception):
+  """Raised when failing to connect to serial device."""
+
+
+class SerialPlug(plugs.BasePlug):
+  """A Plug for serial communication."""
+
+  def __init__(self):
+    """Setup child logger and initialize instance variables."""
+    self.logger = logging.getLogger('openhtf.' + __name__)
```
jettisonjoe: ```'openhtf.' + ``` is not needed here. Since test authors are going to be importing this as a submodule of openhtf, __name__ should already contain it.

openhtf/plugs/uart.py
```diff
@@ -0,0 +1,295 @@
...
+    serial_baudrate: 19600
+    serial_parity: N              # PARITY_NONE, PARITY_EVEN, PARITY_ODD,
+                                    PARITY_MARK,
+                                    PARITY_SPACE = 'N', 'E', 'O', 'M', 'S'
+    serial_stopbits: 1            # STOPBITS_ONE, STOPBITS_ONE_POINT_FIVE,
+                                    STOPBITS_TWO = (1, 1.5, 2)
+    serial_bytesize: 8            # FIVEBITS, SIXBITS, SEVENBITS,
+                                    EIGHTBITS = (5, 6, 7, 8)
+    serial_rtscts: 0              # boolean type
+    serial_timeout: None
```
jettisonjoe: Thinking a bit more about it, if these are config keys only, then a given test script it limited to a single serial device it can talk to. What happens if I want to talk to a DUT via serial as well as control a conveyor belt via serial? Approaches I can think of might include having OpenPort be able to take args to specify the serial settings then return some kind of connection object, or just letting test authors decorate the phase multiple times and use each instance however they want.

jettisonjoe:
As a general comment, I do wanna say that this is an awesome start and I like the obvious diligence that went into it. Just gotta get it dialed in a bit.

openhtf/plugs/uart.py
MasterChiefV: I agree with you for a standalone file that the file name should match what's defined inside.

That said our motivation for the current naming convention of uart/serialplug is to match the existing usb/adbplug such that the file or folder name is the physical interface used (uart and usb) and the plug name specifies the protocol (serial and adb).  This way we can add additional uart based protocol later if necessary.

What do you think?

openhtf/plugs/uart.py
MasterChiefV: yup. __name__ = openhtf.plugs.uart.

openhtf/plugs/uart.py
MasterChiefV: Yes totally agree.  In fact I have this exact requirement of control two serial devices; a dut and a fixture.  Returning a handler to each instance of serial device seems like a reasonable solution.  I will look into this.

openhtf/plugs/uart.py
jettisonjoe: I see. That scheme makes good sense. Let's go with it for now and see how it shapes up.

openhtf/plugs/uart.py
```diff
@@ -0,0 +1,331 @@
...
+import binascii
+import logging
+import time
+
+import openhtf.conf as conf
+import openhtf.plugs as plugs
+from openhtf.util import timeouts
+import serial
+
+conf.Declare('SerialPlug')
```
jettisonjoe: I think we've decided to standardize config key names to use underscore_syntax instead of CamelCase.

googlerhuili:
This plug is refined and ready for merge. Please review. Thanks.

jettisonjoe:
@MasterChiefV, we talked in person a little while back about the timeline for this, and how we wanted to focus on some (internal) milestones before making updates to this PR. Can we close this PR and open a fresh one when you're ready and have time to work on the open-source plug again?

PR #54, opened 2015-11-14
Switch tests to use py.test

fahhem:
Also some additional cleanups of the test files, names, etc.


jettisonjoe:
I like where you're going with this PR and I want to accept most of it. But I also like having the test directory structure mirror the package directory structure rather than being flattened. Our codebase isn't yet big enough to have name collisions when flattened, but the sw test system shouldn't impose naming restrictions (even gentle ones) on the package itself. Plus Amy and I agree that we plan to add a lot more test cases and it will be easier to find test cases if the directory structure reflects the package structure as our test coverage fills out. Thoughts?

fahhem:
Moved the files back into directories. I only flattened them because test/test_exe/test_exe.py seemed redundant. Now hopefully test/exe/test_exe.py is fine.

fahhem:
PTAL, I'd like to continue working but I already have 3 PRs in-flight and github doesn't handle dependent PRs

jettisonjoe:
Looks good, thanks for the help! Merging now.

Issue #48, opened 2015-11-06
Add mutablerecords==0.2.4 to setup.py to allow proper wheel installation


wallacbe:
Add mutablerecords==0.2.4 to setup.py to allow proper wheel installation

PR #55, opened 2015-11-14
Cleanup and update deps

fahhem:
(Just the last commit, the rest are from #54)

Now that it's all inside setup.py, you should run `python setup.py install` before other things. However, to install with -e as before, use `python setup.py develop`, and `python setup.py test` will install test requirements as well.


fahhem:
Fixes #48

jettisonjoe:
Looks like this needs a rebase now that your first PR was merged.

fahhem:
Rebased

jettisonjoe:
Cool. Merging.

PR #58, opened 2015-11-24
Rework measurement, remove use of data descriptors.

grybmadsci:


openhtf/util/measurements.py
```diff
@@ -284,38 +294,34 @@ class Collection(object):  # pylint: disable=too-few-public-methods
...
       return item[0], item[1].GetValue()
-    return itertools.imap(_GetMeasValue, self._values.iteritems())
+    return itertools.imap(_GetMeasurementValue, self._values.iteritems())
 
   def __setattr__(self, name, value):
     self._AssertValidKey(name)
     record = self._values.setdefault(
-        name, MeasuredValue.ForDeclaration(self._declarations[name]))
+        name, MeasuredValue.ForMeasurement(self._measurements[name]))
     record.SetValue(value)
```
jettisonjoe: We could validate here and have PASS/FAIL on the fly if we want to enforce set-once as discussed, right?

jettisonjoe:
I dig it.

fahhem:
What's the reasoning for removing descriptors? The changes I see are that the type is now implicit in the validator (removing .String() and .Number()) but there is now a second implementation of all the validators. Confing uses a small wrapper around colander to get a similar API with none of the code: https://github.com/chainreactionmfg/confing

openhtf/util/measurements.py
grybmadsci: Yep, that's the idea.  I didn't want to shift to that just yet, but it would be easy to update here later.

grybmadsci:
The reasoning is twofold.  For one,w e're trying to minimize syntactic overhead for test authors and better integrate the data description aspect into the output.  With an out of the box data descriptor mechanism, we end up having to write adapter code to serialize it into our desired output format (which is why we used to have three classes around measurements instead of two).

Secondly, in the context of HTF specifically, we have weird data types (multidimensional data) and don't care about lots of other data types (int vs float, for example).  Our use case - validating pass/fail criteria - is just a weird use of data descriptors.  We're not trying to detect invalid values, we're trying to detect valid values that indicate a hardware problem.

jettisonjoe:
I'm gonna go ahead and merge this because I need to continue working on the frontend today and specifically on pass/fail display.

PR #59, opened 2015-11-25
Multicast station discovery service.

jettisonjoe:


jettisonjoe:
Okay, madsci, the frontend is back to where it was before I added the multicast user agent, so this PR is legitimately ready for review now. The planned frontend updates will come in a later PR once I merge in your measurements changes.

openhtf/exe/test_state.py
```diff
@@ -50,24 +48,24 @@ class TestState(object):
     config: The config being used for this test.
     test: openhtf.Test instance describing the test to run.
   """
-  State = Enum(
+  Status = Enum(
```
fahhem: There's two enums called Status, it seems one is more a TestExecutionStatus and one is TestResultStatus (or without Test prefix)?

openhtf/exe/test_state.py
```diff
@@ -50,24 +48,24 @@ class TestState(object):
...
       'RUNNING', 'ERROR', 'TIMEOUT', 'ABORTED', 'WAITING', 'FAIL', 'PASS',
```
fahhem: Looks like ABORTED isn't used?

openhtf/io/frontend/__main__.py
```diff
@@ -133,12 +200,21 @@ def main(argv):
     print('%s\nUsage: %s ARGS\n%s' % (e, sys.argv[0], FLAGS))
     sys.exit(1)
 
-  if not os.path.isdir(FLAGS.rundir):
-    print('ERROR: OpenHTF run directory does not exist', FLAGS.rundir,
-          file=sys.stderr)
-    sys.exit(1)
+  logs.setup_logger()
+
+  frontend_server = HTTPServer(WSGIContainer(app))
```
fahhem: I'm curious about the motivation behind Tornado versus flask's builtin webserver or some other one?

openhtf/io/frontend/templates/station.html
```diff
@@ -30,7 +30,7 @@
             document.getElementById(id).innerHTML = new_doc; });
   }
   function poll(interval_ms) {
```
fahhem: The current setup around poll looks like it's essentially a setInterval, so the equivalent is this:

```js
function poll() {
  ...
}
function setup() {
  setInterval(poll, {{interval}});
}
```

However, this can lead to DoS'ing a slow server (when the server isn't broken but takes longer than interval_ms to respond). Instead, scheduling a new poll only when the last one returned would be better (optionally subtracting the time the poll took to make the interval closer to the poll_interval_ms flag value):

```js
function poll() {
  refresh('test');
  ajax(..., function(id) {
    ...
    setTimeout(poll, {{interval}});
  });
}
function setup() {
  poll();
}
```

openhtf/io/http_api.py
```diff
@@ -34,16 +37,62 @@
...
     executor: An openhtf.exe.TestExecutor object.
   """
+  KILL_TIMEOUT_S = 1  # Seconds to wait between service kill attempts.
+
   def __init__(self, executor):
     super(Server, self).__init__()
+    self.servers = [HTTPServer(executor), MulticastAgent()]
+
+  def Start(self):
+    """Give the server a style-compliant Start method."""
```
fahhem: Which style is this? Is it Google's or something else?

Can you put this Start inside KillableThread (or ExceptionSafeThread) so this isn't a problem anymore?

openhtf/io/http_api.py
```diff
@@ -76,14 +116,93 @@ def do_POST(self):  # pylint: disable=invalid-name
...
+    """Stop the HTTP server."""
     self.Kill()
 
   def _ThreadProc(self):
     """Start up a raw HTTPServer based on our HTTPHandler definition."""
     server = BaseHTTPServer.HTTPServer(('', FLAGS.http_port), self.HTTPHandler)
     server.serve_forever()
+
+
+class MulticastAgent(util.threads.KillableThread):
```
fahhem: If I got this right, MulticastAgent is either a server or a client depending on which function is called? Can you make them two separate classes that share a socket-creation function (the only shared functionality, it seems)?

fahhem:
Can't wait to see a multi-cell station in openhtf!

openhtf/exe/test_state.py
jettisonjoe: With the overall upgrade of http output to include more robust framework state as well as test state, we needed a representative of framework state with access to the right data. The test executor became that representative. So what you call test execution status is more like framework status, whereas  test status refers to the current test run only.

openhtf/exe/test_state.py
jettisonjoe: Good point. I should wire that up to the SIGINT handler like I had planned.

openhtf/io/frontend/__main__.py
jettisonjoe: Flask's built-in server is only intended as a dev server. Their own documentation advises against using it in prod. Previously we used Rocket as our web server, and using Flask's built-in dev server was a temporary thing while I investigated other options. I think tornado looks like a better overall framework fit for our needs and adding it here is a step towards replacing flask entirely.

openhtf/io/frontend/templates/station.html
jettisonjoe: This was hastily-produced code and my bad. I'm actually halfway through overhauling it and the next iteration should look a lot better. I'll include it in this PR when it's ready.

openhtf/io/http_api.py
jettisonjoe: Yes, I would love to do this in KillableThread instead of all over the place. It's openhtf style (I think it's originally Google style -- as defined by our pylintrc) to differentiate methods (from other attributes in some cases, and from functions in other cases) by capitalizing them.

openhtf/io/http_api.py
jettisonjoe: There is shared data as well (the string used for pinging). I could be convinced either way on this one. Putting the Ping method into the server class was an arbitrary decision on my part.

jettisonjoe:
multi-cell station is a bit of a misnomer. This PR takes us toward viewing multiple stations from the same frontend, but they are still separate stations (as defined in our README.md file). Re-implementing pseudo multi-cell support by grouping stations together in the frontend is a planned feature, but not being worked on yet. ;-)

jettisonjoe:
I should also mention that having a recommended plug design pattern for shared test equipment is probably also important for most "multi-cell" scenarios, and we haven't established that yet either.

jettisonjoe:
@grybmadsci, here's what the move-multicast-out-of-http_api approach looks like, with flags declared twice in both places where they're used, but both pulling the same defaults from module variables in the multicast module. Ping() is now a more general module-level function send(), and MulticastAgent is now just a MulticastListener. Listener outsources its response generation to a callback function, and the callback defined in http_api's Server class now uses JSON instead of sending a raw port number. Merge conflicts aside for now, how does this approach look? Any better?

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+  """
+  def __init__(self):
+    super(StationStore, self).__init__()
+    self._live = False
+    self.hostname = socket.gethostname()
+    self.stations = {}
+
+  @staticmethod
+  def MessageFramework(host, port, method='GET', message=None):
+    """Interact with a running instance of openhtf via its HTTP API.
```
grybmadsci: can we get a 'for example' in this docstring as to how you might use this?

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+    Args:
+      host: The IP address of the host running openhtf.
+      port: The TCP port of openhtf's HTTP API on host.
+      method: 'GET' or 'POST'.
+      message: Prompt response for POSTs. Ignored for GETs.
+    """
+    result = None
+    conn = httplib.HTTPConnection(host, int(port))
+    args = [method, '/']
+    if message and method == 'POST':
```
grybmadsci: maybe log something that 'message' is getting ignored if it's passed but the method isn't POST

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+      args.append(message)
+    try:
+      conn.request(*args)
+      if method == 'GET':
+        response = conn.getresponse()
+        if response.status != 200:
+          result = None
+        else:
+          result = json.loads(response.read())
```
grybmadsci: json.loads can also raise if the response is malformed - up to you whether you want to handle that here or propagate up, just an FYI.

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+
+  Attributes:
+    station_id: Station's ID string.
+    state: A dictionary representation of station state with two keys:
+      framework: A dictionary representation of a TestExecutor object.
+      test: A dictionary representation of a TestState object.
+    """
+
+
+class StationStore(threads.KillableThread):
```
grybmadsci: killablethread is kind of messy at best, consider using ExceptionSafeThread instead

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+      except (KeyError, ValueError):
+        _LOG.warning('Ignoring unrecognized discovery response from %s: %s' % (
+            host, response))
+
+  def _ThreadProc(self):
+    """Continuously scan for new stations and add them to the store."""
+    self._live = True
+    while self._live:
+      self._Discover()
+      time.sleep(FLAGS.discovery_interval_s)
```
grybmadsci: if you want to get fancy, you can use a threading event and wait() with a timeout, and if it gets set that's your 'thread killed' signal - ie instead of self._live = False in Stop(), you do self._killed_event.set().  this way, if you have a long discovery interval, your join() in Stop() can take a while.

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+
+  def _Track(self, host, port):
+    """Start tracking the given station."""
+    if (host, port) not in self.stations:
```
grybmadsci: you could make stations a collections.defaultdict instead, or use setdefault() here.

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+      self.stations[(host, port)] = Station()
```
grybmadsci: random syntax note, i don't think you need the () here, just [host, port] works and does the same thing.

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+    """
+    port = int(port)
+    if (host, port) not in self.stations:
+      _LOG.warning(
+          'Store was queried for an unknown station: %s:%s' % (host, port))
+      return None
+    response = self.MessageFramework(host, port)
+    self.stations[(host, port)].state = response
+    if response is not None:
+      station_id = response['framework']['station_id']
```
grybmadsci: consider how a malformed response will be handled here.

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,207 @@
...
+                      'TTL for station discovery pings.')
+
+
+_LOG = logging.getLogger(__name__)
+
+UNKNOWN_STATION_ID = 'UNKNOWN_STATION'
+
+
+class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,no-init
+    'Station', [], {'station_id': UNKNOWN_STATION_ID, 'state': None})):
```
grybmadsci: might want to track the last update time in here too and periodically re-update the state of stale stations just for kicks.

openhtf/util/multicast.py
```diff
@@ -0,0 +1,134 @@
...
+            '!4sL',
+            socket.inet_aton(self.address),
+            socket.INADDR_ANY))
+
+    while self._live:
+      try:
+        data, address = self._sock.recvfrom(MAX_MESSAGE_BYTES)
+        _LOG.debug('Received multicast message from %s: %s'% (address, data))
+        response = self._callback(data)
+        if response:
```
grybmadsci: might check for is not None here, i'm not sure, but i think sending an empty string is not the same as not responding?

openhtf/util/multicast.py
```diff
@@ -0,0 +1,134 @@
...
+import threading
+
+
+_LOG = logging.getLogger(__name__)
+
+DEFAULT_ADDRESS = '239.1.1.1'
+DEFAULT_PORT = 10000
+DEFAULT_TTL = 1
+MAX_MESSAGE_BYTES = 1024  # Maximum allowable message length in bytes.
+PING_STRING = 'OPENHTF_PING'
```
grybmadsci: this pair of strings is never actually used in this module.  in fact, they're really more coupled to the callback that gets passed on, so should probably not be in this module.

nice refactor pulling this into a separate module

openhtf/io/frontend/__main__.py
jettisonjoe: Oops, leaving that in was my bad. I changed it to threading.Thread but it got reverted at some point. Changing it back now.

openhtf/io/frontend/__main__.py
jettisonjoe: That's a great idea...I'm thinking we add it next iteration when I switch everything to long polling or hanging requests.

openhtf/io/frontend/__main__.py
jettisonjoe: Added.

openhtf/io/frontend/__main__.py

openhtf/io/frontend/__main__.py
jettisonjoe: Propagate up is probably fine here, since it will indicate a framework issue.

openhtf/io/frontend/__main__.py
jettisonjoe: Hmm, yeah cool idea. Next pass perhaps?

openhtf/io/frontend/__main__.py
jettisonjoe: Dang, Python, why you gotta make everything so easy?

openhtf/io/frontend/__main__.py
jettisonjoe: Fixed (I think).

openhtf/io/frontend/__main__.py
jettisonjoe: Added some exception handling.

openhtf/exe/__init__.py
```diff
@@ -60,6 +61,11 @@ class TestExecutor(threads.KillableThread):
 
   daemon = True
 
+  FrameworkStatus = Enum(
```
wallacbe: Change to enum34 format

FrameworkStatus = Enum('FrameworkStatus',
      ['CREATED', 'START_WAIT', 'INITIALIZING', 'EXECUTING', 'STOP_WAIT',
      'FINISHING']
  )

openhtf/exe/test_state.py
```diff
@@ -54,18 +52,18 @@ class TestState(object):
                          'WAITING', 'FAIL', 'PASS', 'CREATED'])
 
   _PHASE_RESULT_TO_STATE = {
-      phase_data.PhaseResults.CONTINUE: State.WAITING,
-      phase_data.PhaseResults.REPEAT: State.WAITING,
-      phase_data.PhaseResults.FAIL: State.FAIL,
-      phase_data.PhaseResults.TIMEOUT: State.TIMEOUT,
+      phase_data.PhaseResults.CONTINUE: Status.WAITING,
```
wallacbe: Status should be State

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,223 @@
...
+  def run(self):
+    """Continuously scan for new stations and add them to the store."""
+    self._live = True
+    while self._live:
+      self._Discover()
+      time.sleep(FLAGS.discovery_interval_s)
+
+  def _Track(self, host, port):
+    """Start tracking the given station."""
+    station = self.stations.setdefault((host, port), Station())
```
fahhem: self.stations should be a defaultdict

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,223 @@
...
+    if station.station_id is None:
```
fahhem: It won't ever be None because it's UNKNOWN_STATION_ID by default. You can put None there instead to get that behavior here

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,223 @@
...
+    Args:
+      host: The IP address of the host running openhtf.
+      port: The TCP port of openhtf's HTTP API on host.
+      method: 'GET' or 'POST'.
+      message: Prompt response for POSTs. Ignored for GETs.
+    """
+    result = None
+    conn = httplib.HTTPConnection(host, int(port))
+    args = [method, '/']
+    if message and method == 'GET':
```
fahhem: exposing 'method' here seems wrong, MessageFramework should just have reads and writes (with responses). Two functions, strawman-named ReadFromFramework and SendToFramework or CallToFramework that take the right number of args wouldn't even share much code. Maybe a shared function or contextmanager to do the connection creation (and later cache it) and close it when the context ends?

openhtf/io/frontend/__main__.py
```diff
@@ -32,97 +32,223 @@
...
+      _LOG.warning("Message provided for 'GET' request will be ignored: %s",
+                   message)
+    if message and method == 'POST':
+      args.append(message)
+    try:
+      conn.request(*args)
+      if method == 'GET':
+        response = conn.getresponse()
+        if response.status != 200:
+          result = None
```
fahhem: result is already None above, should just check == (superseded by comment above)

openhtf/io/http_api.py
```diff
@@ -22,51 +22,102 @@
...
+
   def __init__(self, executor):
     super(Server, self).__init__()
-    self.HTTPHandler.executor = executor
 
-  class HTTPHandler(BaseHTTPServer.BaseHTTPRequestHandler):
+    def multicast_response(message):
+      """Formulate a response to a station discovery ping."""
+      if message == multicast.PING_STRING:
+        return json.JSONEncoder().encode(
```
fahhem: is there a reason you don't use json.dumps instead of this?

openhtf/util/multicast.py
```diff
@@ -0,0 +1,134 @@
...
+
+  Returns: A set of all responses that arrived before the timeout expired.
+           Responses are tuples of (sender_address, message).
+  """
+  result = set()
+  sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+  sock.setsockopt(socket.IPPROTO_IP,
+                  socket.IP_MULTICAST_TTL,
+                  ttl)
+  sock.settimeout(timeout_s)
```
fahhem: I'd put these bits that's shared with the Listener above into a shared function, with a simple (strawman) name like _MakeMulticastSocket(). The settimeout() call above in run() can also be moved into the __init__ making it a candidate for inclusion in the shared function

openhtf/util/multicast.py
```diff
@@ -0,0 +1,134 @@
...
+    """Stop listening for messages."""
+    self._live = False
+    self.join()
+
+  def run(self):
+    """Listen for pings until stopped."""
+    self._live = True
+    self._sock.bind(('', self.port))
+    self._sock.settimeout(self.LISTEN_TIMEOUT_S)
+    self._sock.setsockopt(
```
fahhem: Can you comment what this does? It's not clear from the code without looking all of this up

openhtf/io/frontend/__main__.py
grybmadsci: sure, but can you TODO it, as i do believe it's really worth doing.  generally you want to avoid polls/long join()s whenever possible.

openhtf/exe/test_state.py
jettisonjoe: State within a state seemed confusing to me, but I'll leave it for now.

openhtf/exe/__init__.py
jettisonjoe: gp

openhtf/io/frontend/__main__.py
jettisonjoe: Hmm. You may be right. I'll think about this and possibly change it in a future iteration. Leaving it for now in the interests of getting this merged.

openhtf/io/frontend/__main__.py
jettisonjoe: Good eye.

openhtf/io/frontend/__main__.py
jettisonjoe: Fair. Will TODO it.

openhtf/io/frontend/__main__.py
jettisonjoe: Fixed.

openhtf/io/frontend/__main__.py
jettisonjoe: gp

openhtf/io/http_api.py
jettisonjoe: I probably copied and pasted from somewhere in our old code. Will update.

openhtf/util/multicast.py
jettisonjoe: gp. Moved to http_api, and the built-in frontend imports them from there.

openhtf/util/multicast.py
jettisonjoe: Added some clarification on multicast-specific options.

openhtf/util/multicast.py
jettisonjoe: Fair enough. Fixed.

openhtf/util/multicast.py
jettisonjoe: Yeah, this might be good, but I'm gonna defer it for now. I'll look at cleaning this up in the next iteration.

fahhem:
I think you may need to rebase? There's 1 commit that seems irrelevant?

openhtf/exe/__init__.py
```diff
@@ -60,6 +61,10 @@ class TestExecutor(threads.KillableThread):
 
   daemon = True
 
+  FrameworkStatus = Enum('FrameworkStatus',
+                         ['CREATED', 'START_WAIT', 'INITIALIZING', 'EXECUTING',
```
fahhem: What does each status mean?

fahhem:
LGTM

openhtf/io/frontend/__main__.py
grybmadsci: +1, agree with Farz's comment here, can you add a TODO to decrease the chance of this getting lost in the Tome of github comments we're collecting?

PR #63, opened 2015-12-02
Create a TestPhaseInfo to simplify function handling

fahhem:
Instead of following a linked-list of .wraps attributes, we store
information into the TestPhaseInfo and use that where necessary. This
also allows a client library to create TestPhaseInfo objects with the
correct attributes from other sources (config files, different
arrangement of functions, etc)

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -84,32 +84,56 @@ class OutputToJSON(JSONEncoder):
       f.write(self.encode(as_dict))
 
 
-def TestPhase(timeout_s=None, run_if=None):  # pylint: disable=invalid-name
-  """Decorator to wrap a test phase function with the given options.
-
-  Args:
+class TestPhaseOptions(mutablerecords.Record(
+    'TestPhaseOptions', [], {'timeout_s': None, 'run_if': None})):
+  """
+  Attributes:
     timeout_s: Timeout to use for the phase, in seconds.
     run_if: Callback that decides whether to run the phase or not.  The
       callback will be passed the phase_data the phase would be run with.
 
-  Returns:
-    A wrapper function that takes a phase_func and returns a
-        TestPhaseInfo for it with the given options set.
+  Example Usage:
+    @TestPhaseOptions(timeout_s=1)
+    def PhaseFunc(test):
+      pass
   """
 
-  def Wrap(phase_func):  # pylint: disable=invalid-name
-    """Attach the given options to the phase_func."""
+  def __call__(self, phase_func):
+    phase = TestPhaseInfo.WrapOrReturn(phase_func)
+    phase.options = self
+    return phase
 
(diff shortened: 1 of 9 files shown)


openhtf/__init__.py
```diff
@@ -84,32 +84,56 @@ def __call__(self, test_record):  # pylint: disable=invalid-name
...
+PhasePlug = mutablerecords.Record('PhasePlug', ['name', 'cls', 'update_kwargs'])
 
-    if timeout_s is not None:
-      phase_func.timeout_s = timeout_s
-    if run_if is not None:
-      phase_func.run_if = run_if
-    return phase_func
-  return Wrap
+
+class TestPhaseInfo(mutablerecords.Record(
```
grybmadsci: I'll defer to Joe on this one, but we had made the call a while back that we wanted phase functions to remain just functions.  In fact, this was explicitly changed from the way it was done in XTF (which is what you've recreated here).

Since then, we ended up tacking a few attributes onto there, so arguably it really should be a fancier class like this.  I'm somewhat 50/50 on it, so like I said I'll let Joe weigh in.

Again, though, this was an obvious design decision that was made and explicitly changed from XTF, but I didn't see any conversation around "hey guys why did you do this?" or "i think we should change this, can I go ahead and implement it?"

fahhem: Is there a mailing list for this sort of thing or do should we go via bugs? There's nothing about governance or discussion in the CONTRIBUTING.md file. In lieu of that, I talked to Joe in person regarding this.

I could see that it was an explicit change. However, I was doing other things and this started to get in my way so I changed it one piece at a time until I was done and realized I had fully converted it back. I talked to Joe in person after sending this out and he seemed potentially in favor of it, but hadn't seen it just yet.

The thing that bothered me the most was the '.wraps' loops in various places. I also think this cleaned things up and didn't change the user API at all. They're free to add decorators and such to their phases. Before, if they wanted things like plugs and measurements to bubble up, they had to add outer_func.wraps = wrapped_func, which was not documented anywhere. Documenting it as "openhtf.TestPhaseInfo.WrapOrReturn(wrapped_func).func = outer_func" versus the .wraps earlier is also less prone to typo debugging.

fahhem: Also, the only line in CONTRIBUTING.md regarding this is here:

    We welcome patches and contribution to OpenHTF. It's as easy as submitting a pull request with your changes in order to kick off the review process. But first, please read through the rest of this doc.

So I followed the one and only request available.

fahhem:
Updated to match head. This change was discussed at a weekly meeting and there was general consensus around the change in theory. If the actual code could be merged, it will help my future changes as this touches 9 core framework files.

jettisonjoe:
I'm going to merge this so you're not blocked, but I reserve the right to (maybe) rename stuff later for clarity.

PR #68, opened 2015-12-05
Remove requirement for config file, check for station_id key

wallacbe:
Allow to launch openhtf tester from command line without a config file

### DIFF
--- a/openhtf/conf.py
+++ b/openhtf/conf.py
@@ -46,8 +46,7 @@ from openhtf.util import threads
 
 FLAGS = gflags.FLAGS
 
-gflags.DEFINE_string('config',
-                     '/usr/local/openhtf_client/config/clientfoo.yaml',
+gflags.DEFINE_string('config', None,
                      'The OpenHTF configuration file for this tester')
 
 gflags.DEFINE_multistring(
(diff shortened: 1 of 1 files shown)


openhtf/conf.py
```diff
@@ -257,29 +256,31 @@ def Load(self, config_file=None, force_reload=False,
...
+            raise ConfigurationInvalidError('No data', config_file)
+          self._state.clear()
+          self._state.update(data)
 
       # Load string values from flags
       for keyval in FLAGS.config_value:
         key, val = keyval.split('=')
         self._state[key] = val
+   
+      if 'station_id' not in self._state:
```
fahhem: Is this necessary for openhtf to function now? Or only for the frontend or something?

wallacbe: If station_id key isn't present we see the following:

  File "./hello_world.py", line 91, in <module>
  File "/usr/local/google/home/wallaceb/openhtf/openhtf/__init__.py", line 204, in Execute
    os.getpid()).SaveToFile(FLAGS.rundir)
  File "/usr/local//home/openhtf/openhtf/io/rundata.py", line 71, in SaveToFile
    filename = os.path.join(directory, self.station_id)
  File "/usr/local//home/venv/lib/python2.7/posixpath.py", line 75, in join
    if b.startswith('/'):
AttributeError: 'NoneType' object has no attribute 'startswith'

jettisonjoe: The rundata module uses it as we see, and the built-in JSON output module uses it as well.

wallacbe:
Notes from meeting:
Add default option to conf Declare function
provide a default station ID

PR #78, opened 2015-12-07
Add Ethersync plug and general USB plug, change run function name

amyxchen:


fahhem:
It would be best if you opened separate PRs for each (set of) features you're adding. Also, you don't have to close and open a PR to make a change, just keep making commits to the branch on github and it'll update the PR attached to that branch. It helps to create a non-master branch in that case

openhtf/exe/phase_data.py
```diff
@@ -118,6 +118,7 @@ def RecordPhaseTiming(self, phase, test_state):
         measurement.name: measurement
         for measurement in getattr(phase, 'measurements', [])
     }
+
```
grybmadsci: can we revert this? it's just diff noise.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
...
-    **kwargs: Arguments to pass to respective handle's Open() method.
-
-  Returns:
-    Instance of UsbHandle.
-  """
-  if conf.Config().usb_server:
-    return None
+def _get_usb_serial(port_addr):
+  """Get a usb serial based on the Cambrionix unit mac address in configuration."""
+  cmd = '/usr/local/google/home/amyxchen/esuit64 -t "DEVICE INFO, %s"' % port_addr
```
grybmadsci: can't have code in here that refers to a local custom binary like this.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
...
 conf.Declare('libusb_rsa_key', 'A private key file for use by libusb auth.')
+conf.Declare('device_type', 'android, fastboot or general usb type')
+conf.Declare('usb_hub', 'local or Ethersyncy')
+
+_LOG = logging.getLogger(__name__)
+
+class GeneralUsbAttributeError(Exception):
+  """Indicates the usb class/subclass/protocol info was invalid."""
+
+class general_usb(object):
```
grybmadsci: this is just a container class, should be a namedtuple.

openhtf/plugs/usb/local_usb.py
```diff
@@ -214,7 +220,7 @@ def IterOpen(cls, name=None, interface_class=None, interface_subclass=None,
         if (port_path is not None and
             cls._DeviceToSysfsPath(device) != port_path):
           continue
-
+
```
grybmadsci: please revert this, just whitespace change.

setup.py
```diff
@@ -75,8 +75,7 @@ def finalize_options(self):
     self.test_args = []
     self.test_suite = True
 
-  def run(self):
-    test.run(self)
+  def run_tests(self):
```
grybmadsci: this doesn't look like it belongs in this PR, can we pull this out?

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
 conf.Declare('usb_server', 'USB Server IP/Hostname')
 conf.Declare('usb_server_port', 'USB Server Port',
              default_value=10000)
-
 conf.Declare('libusb_rsa_key', 'A private key file for use by libusb auth.')
+conf.Declare('device_type', 'android, fastboot or general usb type')
+conf.Declare('usb_hub', 'local or Ethersyncy')
```
grybmadsci: usb_hub is a poor name, i think you're trying to group usb devices into a single top-level conf key.  i don't think this even actually belongs in config, though, because config is intended to capture the types of things that vary between multiple instances of the same tester.  in this case, usb class/subclass/protocol/vid/pid are really baked into the device, so should be hardcoded instead.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
 conf.Declare('usb_server', 'USB Server IP/Hostname')
 conf.Declare('usb_server_port', 'USB Server Port',
              default_value=10000)
-
 conf.Declare('libusb_rsa_key', 'A private key file for use by libusb auth.')
+conf.Declare('device_type', 'android, fastboot or general usb type')
```
grybmadsci: hm, it's unclear what this means... which device's type are you specifying?  i don't think this is a config thing, i think when you want an android device, you use an android plug, and when you want a fastboot device, you use a fastboot plug...

jettisonjoe: I agree with @grybmadsci that this key shouldn't be necessary.

openhtf/plugs/usb/__init__.py
jettisonjoe: I think what we had talked about was having a ```usb``` key under any plug that uses a USB connection (like adb and fastboot) to tell them how to get their USB handles (local, ethersync, and conceivably minnow when/if we resurrect that).

openhtf/plugs/usb/__init__.py
jettisonjoe: This brings up a good point. esuit64 is the tool we have for managing EtherSync connections. Do we make having it in the executable path a requirement for this plug? What happens if the user has esuit32 instead (if such a thing exists)? It might be worth asking Cambrionix again about their python package for EtherSync since it has been a couple of months since we last heard from them about it.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
...
+    return self.usb_subclass
+  @property
+  def PROTOCOL(self):
+    return self.usb_protocol
+
+""" Config file example for device at Ethersync
+device_type: android (fastboot, general)
+usb_hub: 
+  Ethersync:
+   unit_name: EtherSyncca9167
```
jettisonjoe: Unit name can be derived from MAC address, so we shouldn't need a separate key for it.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
...
+    raise GeneralUsbAttributeError('must set device type')
+
+  return local_usb.LibUsbHandle.Open(
+         interface_class=device.CLASS,
+         interface_subclass=device.SUBCLASS,
+         interface_protocol=device.PROTOCOL,
+         serial_number=serial,
+         vendor_id=vendor_id,
+         product_id=product_id)
+
```
jettisonjoe: Instead of doing all the device type checking here, I think FastbootPlug and ADBPlug should be left how they are, but if we're using EtherSync to connect to the device, we first get the serial of the device attached to the configured EtherSync/port through the EtherSync API, then pass it through the serial_number kwarg when trying to open the USB handle.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,56 +38,153 @@ def MyPhase(test, adb):
...
+        pass
+      try:
+        serial = device['serial']
+      except KeyError:
+        pass
+
+      try:
+        device = general_usb(device['class'], device['subclass'], device['protocol'])
+      except KeyError:
+        raise GeneralUsbAttributeError('must set class\subclass\protocol for \
```
jettisonjoe: Prefer adjacent literal line joins here instead of backslash joins, like:
```
raise GeneralUsbAttributeError('foo bar'
                                                  'baz qux')
```

setup.py
amyxchen: This is only one line change to run the unit test. 

Amy

openhtf/plugs/usb/__init__.py
amyxchen: I guess "usb_hub" may be a bad name, but it is one device which has multiple USB ports for test station to connect, and it converts the network USB to be a local USB, This piece of info needs to be in config file in order to open the USB port.

Amy

openhtf/plugs/usb/__init__.py
amyxchen: hmm, this device type specify the android, fastboot or pure usb device. it is true, if it is a android/fastboot/usb device, just use AdbPlug/FastbootPlug/UsbPlug, since internally, AdbPlug/FastbootPlug are both based on the usb handle to do another level connection, I simplified the code base on each plug, and get the info from config to open the underneath usb device.

openhtf/plugs/usb/__init__.py
amyxchen: This change makes it look complicated, general idea is to open the underneath USB device at the one place instead of at each of plugs, otherwise, at both AdbPlug and FastbootPlug, need to check if it is remote or local, the same logic code needs to add both of the plugs.

jettisonjoe: Per our in-person discussion, the individual plug code seems like the correct place to introduce constraints specific to that plug. Especially given that a single test may use multiple forms of USB connection before it's done.

openhtf/plugs/usb/__init__.py
jettisonjoe: Talked in person and leaning towards a ```remoteusb``` config key. If that key isn't set, assume local USB. If it is set, it will tell us the type of remote USB connection to make (ethersync, minnodw, etc.).

openhtf/plugs/usb/__init__.py
```diff
@@ -36,9 +38,27 @@ def MyPhase(test, adb):
 conf.Declare('usb_server', 'USB Server IP/Hostname')
 conf.Declare('usb_server_port', 'USB Server Port',
              default_value=10000)
-
 conf.Declare('libusb_rsa_key', 'A private key file for use by libusb auth.')
+conf.Declare('remote_usb', 'Ethersyncy or other',default_value=None)
```
jettisonjoe: The docstring for remote_usb could be more descriptive. Something like "Type of provider of remote USB connectivity." Then you could add text to the module docstring to show an example and list the choices (right now there will only be 'ethersync'.

jettisonjoe: Also, we probably shouldn't provide a default value here.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,9 +38,27 @@ def MyPhase(test, adb):
...
+
+_LOG = logging.getLogger(__name__)
 
+class GeneralUsbAttributeError(Exception):
+  """Indicates the usb class/subclass/protocol info was invalid."""
+
+def _get_usb_serial(port_addr):
```
jettisonjoe: We should strongly consider wrapping the esuit64 tool in an EtherSync object abstraction. This would make EtherSync interactions more declarative, would be usable as a straight 'EtherSync' plug if we need to do other things besides just get the serial for making a handle (like switch the USB ports on and off), and so that when Cambrionix comes out with a native python API we can easily switch over to that under the hood.

This would also give us a clean(er) place to do error checking. For example, if the esuit64 tool doesn't exist in the path, we should be able to detect that and give a useful error message.

openhtf/plugs/usb/__init__.py
```diff
@@ -36,9 +38,27 @@ def MyPhase(test, adb):
...
+  """Get a usb serial based on the Cambrionix unit/port address in configuration."""
+  cmd = 'esuit64 -t "DEVICE INFO,%s"' % port_addr.strip()
+  info = commands.getstatusoutput(cmd)[1]
```
jettisonjoe: I believe the 'commands' module is deprecated and we should probably use 'subprocess' instead. Especially if we intend on supporting Python3 eventually.

openhtf/plugs/usb/__init__.py
```diff
@@ -52,9 +72,21 @@ def _open_usb_handle(**kwargs):
...
+    if remote_usb:
+      if remote_usb.has_key('Ethersync'):
+        device = remote_usb['Ethersync']
+        if isinstance(device, dict) and device.has_key('port_addr'):
+          serial = _get_usb_serial(device['port_addr'])
+        else:
+          raise GeneralUsbAttributeError('Ethersync needs port_addr address to be set')
+
+    kwargs['serial_number'] = serial
     return local_usb.LibUsbHandle.Open(**kwargs)
```
jettisonjoe: Shouldn't need to modify kwargs here:
```return local_usb.LibUsbHandle.Open(serial_number=serial, **kwargs)```

File: openhtf/plugs/usb/__init__.py [hunk 15]
```diff
@@ -52,9 +72,21 @@ def _open_usb_handle(**kwargs):
   Returns:
     Instance of UsbHandle.
   """
+
+  serial = None
   if conf.Config().usb_server:
     return None
   else:
+    remote_usb = conf.Config().remote_usb
```
jettisonjoe: The usb_server config key is superseded by the remote_usb key. I'm thinking we should get rid of usb_server, and assume a local connection unless remote_usb is configured.

openhtf/plugs/usb/__init__.py
```diff
@@ -52,9 +72,21 @@ def _open_usb_handle(**kwargs):
...
+    if remote_usb:
+      if remote_usb.has_key('Ethersync'):
```
jettisonjoe: I was thinking if remote_usb is set to 'ethersync', we require another top-level key called 'ethersync' that holds the ethersync configuration info. Definitely open to discussion on this, though. At the same time, we should think about standardizing our config keys on either CamelCase or underscore_notation.

openhtf/plugs/usb/__init__.py
amyxchen: I am behind this. we should get rid of usb_server

openhtf/plugs/usb/__init__.py
```diff
@@ -52,10 +58,18 @@ def _open_usb_handle(**kwargs):
...
-  if conf.Config().usb_server:
-    return None
-  else:
-    return local_usb.LibUsbHandle.Open(**kwargs)
+
+  serial = None
+  remote_usb = conf.Config().remote_usb.strip()
+  if (remote_usb == 'EtherSync'):
+    device = conf.Config()[remote_usb]
+    if device and isinstance(device, dict) and device.has_key('mac_addr') and device.has_key('plug_port'):
```
jettisonjoe: Instead of the 3-conjunct test here, you might want to just put this stuff into a 'try' block and raise the appropriate exception when you catch a KeyError.

openhtf/plugs/usb/__init__.py
```diff
@@ -52,10 +58,18 @@ def _open_usb_handle(**kwargs):
...
     Instance of UsbHandle.
   """
-  if conf.Config().usb_server:
-    return None
-  else:
-    return local_usb.LibUsbHandle.Open(**kwargs)
+
+  serial = None
+  remote_usb = conf.Config().remote_usb.strip()
+  if (remote_usb == 'EtherSync'):
```
jettisonjoe: We used to have a decently elegant pattern where we would register a class name to a config value. Just a note that we probably want to do something like that here eventually, but this if-test is fine for now, especially since EtherSync is currently our only remote USB option.

openhtf/plugs/cambrionix/__init__.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
```
jettisonjoe: We should start using '2016' for new files.

jettisonjoe:
I'm gonna go ahead and merge this so test engineers can start using it. But let's plan to revisit some of the spots we talked about later when we have time to dial it in further.

Issue #72, opened 2015-12-05
phase_data raises "RuntimeError: Validate must only be called once" with loop=True


wallacbe:
To reproduce, add loop=True to test.Execute in hello_world.py and loop it twice:

sudo python hello_world.py --config hello_world.yaml 
Provide a DUT ID in order to start the test.
test
What's the widget type?
hi
Provide a DUT ID in order to start the test.
test
INFO:hello_world.py:Hello World!
What's the widget type?
hi
INFO:hello_world.py:Example says: Example Plug Did Stuff!
2015-12-05 15:51:42,619 - ERROR - Failed to execute test.
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/__init__.py", line 138, in _ThreadProc
    self._ExecuteTest(phase_executor)
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/__init__.py", line 164, in _ExecuteTest
    for phase_result in phase_executor.ExecutePhases():
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/phasemanager.py", line 150, in ExecutePhases
    result = self._ExecuteOnePhase(self._test_state.pending_phases[0])
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/phasemanager.py", line 180, in _ExecuteOnePhase
    result_wrapper.SetResult(phase_thread.JoinOrDie())
  File "build/bdist.linux-x86_64/egg/contextlib2.py", line 56, in __exit__
    next(self.gen)
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/phase_data.py", line 147, in RecordPhaseTiming
    for name, measurement in measurement_map.iteritems()
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/exe/phase_data.py", line 147, in <dictcomp>
    for name, measurement in measurement_map.iteritems()
  File "/usr/local/lib/python2.7/dist-packages/openhtf-1.0-py2.7.egg/openhtf/util/measurements.py", line 152, in Validate
    raise RuntimeError('Validate must only be called once')
RuntimeError: Validate must only be called once
ERROR:root:Thread raised an exception: Thread-1
[...]

PR #81, opened 2015-12-09
Move outcome into MeasuredValue

fahhem:
Tested with example plugin.


fahhem:
Should fix #72

grybmadsci:
outcome was intentionally left out of MeasuredValue by design, because it's tightly coupled with the validators, which are in the Measurement instead.

i appreciate the sentiment of wanting to help, but this code was just authored a week ago and this was a minor bug that should have been simply assigned to the original author.  there's already a fix out for it.

PR #82, opened 2015-12-10
Refactor main loop and fix looping bug in measurements.

grybmadsci:
Refactored the main loop to be a bit more readable, no logic changes.

Also fixed Measurement.Validate() being called multiple times when loop testing by making a deepcopy of the Measurement object when we create the measurements.Collection().

### DIFF
--- a/openhtf/exe/__init__.py
+++ b/openhtf/exe/__init__.py
@@ -74,71 +74,6 @@ class TestExecutor(threads.KillableThread):
     """Style-compliant start method."""
     self.start()
 
-  def GetState(self):
-    """Return the current TestState object."""
-    return self._test_state
-
-  def _ThreadProc(self):
-    """Handles one whole test from start to finish.
-
-    When this finishes, the parent loops back around and calls us again.
-    """
-    while True:
-      with contextlib.ExitStack() as exit_stack, \
-          LogSleepSuppress() as suppressor:
-        _LOG.info('Starting test %s', self.test.filename)
-
-        self._current_exit_stack = exit_stack
-        exit_stack.callback(lambda: setattr(self, '_current_exit_stack', None))
-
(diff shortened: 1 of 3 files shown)


openhtf/exe/__init__.py
```diff
@@ -154,17 +89,95 @@ def Wait(self):
...
+    if not self.test.loop:
+      self._RunTestLoop.once(self)
+    else:
+      self._RunTestLoop()
+
+  @threads.Loop
+  def _RunTestLoop(self):
+    """Handles one whole test from start to finish.
+
+    When this finishes, the parent loops back around and calls us again.
```
jettisonjoe: This sentence of the docstring is no longer accurate (nor was it in the previous iteration of the code for that matter). We should probably update or remove it.

openhtf/util/validators.py
```diff
@@ -107,11 +113,15 @@ class _MatchesRegex(object):
...
   
     def __init__(self, re_module, regex):
+      self._re_module = re_module
       self._compiled = re_module.compile(regex)
       self._regex = regex
   
     def __call__(self, value):
       return self._compiled.match(str(value)) is not None
+
+    def __deepcopy__(self, memo):
```
jettisonjoe: ```memo``` is unused in the function and so should be named ```_``` or ```dummy_memo``` to pass lint.

grybmadsci: Done.

openhtf/exe/__init__.py
grybmadsci: Done.  Updated to include the semantics of @threads.Loop to save someone the extra level of indirection in doc-hunting.

jettisonjoe:
LGTM

jettisonjoe:
Merging this now, but reserving the right to discuss the placement of outcomes at greater length later.

PR #83, opened 2015-12-10
prompt timeout

amyxchen:


openhtf/io/user_input.py
```diff
@@ -37,7 +37,6 @@
                       300,
                       'User prompt timeout in seconds.')
 
-
```
jettisonjoe: 2 blank lines between top-level stuff is conventional. I would leave this line in.

openhtf/io/user_input.py
```diff
@@ -51,7 +50,6 @@ class PromptUnansweredError(Exception):
 
 Prompt = collections.namedtuple('Prompt', 'id message text_input')
 
-
```
jettisonjoe: Same here with the 2 blank lines.

openhtf/io/user_input.py
```diff
@@ -83,8 +82,8 @@ def DisplayPrompt(self, message, text_input=False):
 
       console_prompt = ConsolePrompt(
           message, functools.partial(self.Respond, self.prompt.id))
-      console_prompt.start()
-      self._cond.wait(int(FLAGS.prompt_timeout_s))
+      console_prompt.start()
```
jettisonjoe: Looks like you have some trailing spaces at the end of this line.

openhtf/io/user_input.py
```diff
@@ -63,7 +61,8 @@ def __init__(self):
     self._response = None
     self._cond = threading.Condition()
 
-  def DisplayPrompt(self, message, text_input=False):
+  def DisplayPrompt(self, message, text_input=False,
+                    timeout=int(FLAGS.prompt_timeout_s)):
```
jettisonjoe: This might be my fault from before, but the int() cast probably isn't needed here since this flag is declared as an integer. I wonder why I put it in.

openhtf/exe/triggers.py
```diff
@@ -114,5 +114,5 @@ def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
   """Make a test start trigger based on prompting the user for input."""
   def trigger():  # pylint: disable=missing-docstring
     prompt_manager = user_input.get_prompt_manager()
-    return prompt_manager.DisplayPrompt(message, text_input=text_input)
+    return prompt_manager.DisplayPrompt(message, text_input=text_input, timeout=356*24*3600)
```
jettisonjoe: I think you meant ```365``` here.

amyxchen:
All comments fixed.

openhtf/io/user_input.py
```diff
@@ -63,7 +63,8 @@ def __init__(self):
     self._response = None
     self._cond = threading.Condition()
 
-  def DisplayPrompt(self, message, text_input=False):
+  def DisplayPrompt(self, message, text_input=False,
+                    timeout=FLAGS.prompt_timeout_s):
```
grybmadsci: It's a good convention to include the units in the argument name, so 'timeout' should be 'timeout_s' if it's seconds.

grybmadsci: It's generally bad to use mutable types as default values (there's a good explanation of why here: http://docs.python-guide.org/en/latest/writing/gotchas/), better to default to None and then check for None in the code.

openhtf/io/user_input.py
```diff
@@ -84,7 +85,7 @@ def DisplayPrompt(self, message, text_input=False):
       console_prompt = ConsolePrompt(
           message, functools.partial(self.Respond, self.prompt.id))
       console_prompt.start()
-      self._cond.wait(int(FLAGS.prompt_timeout_s))
+      self._cond.wait(timeout)
```
grybmadsci: if you do all the other changes i mentioned, this line becomes:
self._cond.wait(timeout_s or FLAGS.prompt_timeout_s)

This has the handy side effect of allowing you to pass 'None' explicitly (or just leave it out), causing an indefinite timeout (usually good to be able to do this so you don't have to make up "really big value"s).

openhtf/exe/triggers.py
```diff
@@ -114,5 +114,5 @@ def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
   """Make a test start trigger based on prompting the user for input."""
   def trigger():  # pylint: disable=missing-docstring
     prompt_manager = user_input.get_prompt_manager()
-    return prompt_manager.DisplayPrompt(message, text_input=text_input)
+    return prompt_manager.DisplayPrompt(message, text_input=text_input, timeout=365*24*3600)
```
grybmadsci: If this was meant to be just a "really big value", then you should just leave out the timeout and let it default to None, with the other changes I mentioned, that will cause no timeout (it will block indefinitely).

jettisonjoe: Yeah, I guess @grybmadsci is right. Having this block indefinitely seems cleaner than having a very long timeout now that I see it.

openhtf/io/user_input.py
jettisonjoe: I don't see how that gotcha applies here, since we're not creating a new instance in the function definition (as in the example) but accessing an essentially global object that has already been defined elsewhere. The pattern of using a flag's value as a kwarg's seems reasonable to me. But perhaps I'm missing something?

openhtf/io/user_input.py
jettisonjoe: Hmm. Being able to pass ```None``` explicitly is a good argument in favor of doing it @grybmadsci's way.

openhtf/io/user_input.py
fahhem: FLAGS.prompt_timeout_s becomes a primitive (integer). However, it's value is accessed at function declaration, which is at module import, when FLAGS have not been parsed yet. You will either always get the default or whatever gflags does when it's not ready yet.

Also, I'm not sure if this is a wanted feature or not, but FLAGS values can be changed at runtime, so accessing it each time rather than once also enables getting the new value.

jettisonjoe: Aaahhh, I see now. Indeed I was being dumb. Do it @grybmadsci's way.

openhtf/io/user_input.py
```diff
@@ -84,7 +85,7 @@ def DisplayPrompt(self, message, text_input=False):
       console_prompt = ConsolePrompt(
           message, functools.partial(self.Respond, self.prompt.id))
       console_prompt.start()
-      self._cond.wait(int(FLAGS.prompt_timeout_s))
+      self._cond.wait(timeout_s or FLAGS.prompt_timeout_s)
```
jettisonjoe: I think in order for the explicit ```None``` thing to work, we have to make the flag default ```None``` instead of 300 also. I think we might have talked about this earlier but I forget, so putting it here.

jettisonjoe:
Cool, merging.

PR #84, opened 2015-12-17
Unittest

amyxchen:


setup.py
```diff
@@ -75,8 +75,7 @@ def finalize_options(self):
     self.test_args = []
     self.test_suite = True
 
-  def run(self):
-    test.run(self)
```
jettisonjoe: Is test.run not needed?

fahhem:
Ah, this is a great change! It overwrites only the running of tests, not the rest of it.

setup.py
fahhem: test.run eventually runs this run_tests function

jettisonjoe: Got it. Just found that in setuptools' source.

jettisonjoe:
Merging.

PR #86, opened 2015-12-28
Use tox to test multiple environments

fahhem:
Initially, just enables python 2.7 and 3.4. However, Python 3 support is broken for a few reasons, so this just lets us start working towards that goal.


jettisonjoe:
Cool, thanks!

Issue #88, opened 2015-12-30
    'mutablerecords==0.2.4' is listed twice in setup.py


wallacbe:
'mutablerecords==0.2.4' is listed twice in setup.py

fahhem:
Yeah, two PRs to do the same thing were merged: https://github.com/google/openhtf/blame/master/setup.py

I'll send a PR to fix it now

PR #89, opened 2015-12-31
Remove second mention of mutablerecords

fahhem:


fahhem:
Fixes #88

jettisonjoe:
Oops, should have caught that doubling when it went in. Sorry!

PR #101, opened 2016-01-22
add new test unit, change log file time stamp, add a check for combrionix

amyxchen:
this PR include:
   - A new test unit added. 
   - Change the log file time stamp format to be the same as test record. 
   - Add the combrionix check for the case:
       1. the combrionix unit's port not in use, adding command to put the port in use mode
       2. fix the bug when not using combrionix unit.

### DIFF
--- a/openhtf/plugs/cambrionix/__init__.py
+++ b/openhtf/plugs/cambrionix/__init__.py
@@ -58,10 +58,14 @@ class EtherSync(object):
     cmd = (['esuit64', '-t', arg])
     info = subprocess.check_output(cmd,stderr=subprocess.STDOUT)
     serial = None
-
     if "SERIAL" in info:
       serial_info = info.split('SERIAL:')[1]
       serial = serial_info.split('\n')[0].strip()
+      use_info = info.split('BY')[1].split(' ')[1]
+      if use_info == 'NO':
+        cmd = (['esuit64', '-t', 'AUTO USE ALL'])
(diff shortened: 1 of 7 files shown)


openhtf/plugs/cambrionix/__init__.py
```diff
@@ -58,10 +58,14 @@ def GetUSBSerial(self, port_num):
...
     serial = None
-
     if "SERIAL" in info:
       serial_info = info.split('SERIAL:')[1]
       serial = serial_info.split('\n')[0].strip()
+      use_info = info.split('BY')[1].split(' ')[1]
+      if use_info == 'NO':
+        cmd = (['esuit64', '-t', 'AUTO USE ALL'])
+        subprocess.check_output(cmd,stderr=subprocess.STDOUT)
+        time.sleep(50.0/1000.0)
```
jettisonjoe: Is there a surer way to do this than a timeout? Like something we can loop until we confirm the AUTO USE ALL has been accepted by the hub?

setup.py
```diff
@@ -54,7 +54,6 @@ def run(self):
     'Werkzeug==0.10.4',
 ]
 
-
```
jettisonjoe: This newline is intentional (2 blank lines before a top-level class definition). No need to delete it.

openhtf/plugs/cambrionix/__init__.py
amyxchen: This command has no return code for us to check, I will request for it.

PR #107, opened 2016-02-04
Implement mfg-inspector style TestRun Proto output and Upload

grybmadsci:
This adds support for writing TestRun protos to disk, or for uploading them directly to mfg-inspector.com on test completion.

Note that since the current output mechanism doesn't really have a mechanism for handling output failures, it's recommended that you output both to disk and upload, so you can manually retry the upload later if it failed (a warning message will be printed if the upload fails, at least).

Also shuffled a couple things around a little for json output.


fahhem:
I think you may need to rebase?

openhtf/io/output/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+  def __call__(self, test_record):  # pylint: disable=invalid-name
+    credentials = oauth2client.client.SignedJwtAssertionCredentials(
+        service_account_name=self.user,
+        private_key=self.keydata,
+        scope=self.SCOPE_CODE_URI,
+        user_agent='OpenHTF Guzzle Upload Client',
+        token_uri=self.TOKEN_URI)
+    credentials.set_store(self._MemStorage())
+
+    testrun = mfg_inspector.TestRunFromTestRecord(test_record)
```
fahhem: Do we want an option to write the TestRun to disk for stations that may not have immediate internet access?

openhtf/io/output/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+import logging
+import oauth2client.client
+import threading
+
+from openhtf import util
+from openhtf.io.output import json_factory
+from openhtf.io.output import mfg_inspector
+
+OutputToJSON = json_factory.OutputToJSON  # pylint: disable=invalid-name
```
fahhem: This gets used only in names.py, which could just refer to it correctly.

openhtf/io/proto/testrun.proto
```diff
@@ -0,0 +1,257 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
```
fahhem: Slightly out of date

openhtf/io/output/mfg_inspector.py
```diff
@@ -0,0 +1,276 @@
...
+    if mimetype in MIMETYPE_MAP:
+      testrun_param.type = MIMETYPE_MAP[mimetype]
+    else:
+      # pylint: disable=no-member
+      testrun_param.type = testrun_pb2.InformationParameter.BINARY
+      # pylint: enable=no-member
+
+
+def _MangleMeasurement(name, value, measurement, mangled_parameters):
+  """Flatten parameters for backwards compatibility, watch for collisions.
```
fahhem: Any interest in attaching an info_parameter for each multidim measurement as we discussed?

fahhem:
Can you add a quick unit test? Some level of coverage will help future renames/refactors

openhtf/io/output/__init__.py
grybmadsci: ah, yeah, good call, done!

openhtf/io/proto/testrun.proto
grybmadsci: heh, lots of files have a 2014 date, now that I look at it.. updated here since we're adding it, maybe if i feel ambitious i'll do another PR updating all the copyright notices.

fahhem: Oh I didn't even notice that. I meant that the file itself is actually out of date, since it's missing the MULTIDIM_JSON and related changes

openhtf/io/output/mfg_inspector.py
grybmadsci: not in this change, but you're welcome to do it once this is merged, i don't know when i'm going to have the time to, but likely not for a while.

grybmadsci:
Unit test added, that's neat that we have the setup.py 'test' command set up!

PR #110, opened 2016-02-09
Cleanup conf uses

fahhem:
Some quick cleanups from looking at the code


openhtf/__init__.py
```diff
@@ -86,7 +86,7 @@ def __call__(self, test_record):  # pylint: disable=invalid-name
 
 class TestPhaseOptions(mutablerecords.Record(
     'TestPhaseOptions', [], {'timeout_s': None, 'run_if': None})):
-  """
+  """Test Phase Options object.
```
grybmadsci: this is not a useful docstring, all you did was add spaces between the words already in the name and add 'object' after it -.-

fahhem: It was to suppress a lint error. If you have a better docstring, I'm all ~~ears~~ eyes

grybmadsci:
LGTM, that one docstring could use a minor update.

As a side note, the conf module is a holdover from XTF that maybe could use a little love eventually.  I like the cleanup you did here, but the ambiguity between the two ways to access the config (and when should it be passed around instead of always accessed directly) indicates to me that it's not as clean an implementation as it should be.

Low priority for now, just some food for thought.

fahhem:
Yeah, I would rather it get passed around. I have some other big changes that will affect these things so I didn't want to conflate the two. In the meantime, what do you think about the approach here: https://github.com/chainreactionmfg/confing

I was thinking of adding nesting, so openhtf would have 'htf' and each plug or other file would have 'htf.plug_name' or 'htf.phase_mgr'. Then a load into 'htf' with key 'plug_name' would load into the 'htf.plug_name' namespace. It solves the name-collision/long-name issues we had, and it doesn't allow any globals. The loaded config is a new object and you have to pass that around, there is no access to a 'global config' like XTF's. It would be a change in the API to the plugs, however, as they don't take the config object anywhere.

fahhem:
I fixed the merge conflicts

jettisonjoe:
Fix one merge conflict, and two new ones spring up to take its place...

fahhem:
Because someone keeps jumping the line...

fahhem:
Okay, should finally be mergeable and pass tests

openhtf/__init__.py
grybmadsci: 'Options used to override default framework behaviors handling test phases.'

grybmadsci:
I think namespacing is unnecessary complexity.  i personally don't mind the global object model.  in fact, my preference would be to change it slightly to mirror the gflags interface and not pass around the object at all (just refer directly to the keys you want, like you do with FLAGS).

In fact, we used a separate config outside of flags in the first place largely to simplify deployment, which isn't really a concern of OpenHTF.  It's probably still worth keeping the two separate, but a similar interface between them would likely be appreciated by users.

fahhem:
I believe Joe wanted to keep deployment sane with OpenHTF, if not provide a
deployment solution eventually. Namespacing was just a possibility.
A global object has its benefits (simplicity, one can grep for accesses to the
'path') and downsides (gets tempting to use for constants, can't be overridden
by callers).
This latter problem occurs when you want to use the same implementation for two
different things, such as USB or UART to two, potentially very different,
devices. If the UART plug refers to conf.uart_filename directly, then you're
stuck. If it takes in a config, then the caller (framework or part of the test)
can pass in a context-specific config object with the right uart_filename. Is
there another solution to this? At the moment, the ethersync stuff has run into
this problem and shoved itself inside both the FastbootPlug and AdbPlug to
inject itself at the right moment and replace the usb device.

On Thu, Feb 11, 2016 at 1:40 PM, grybmadsci < notifications@github.com > wrote:
I think namespacing is unnecessary complexity. i personally don't mind the
global object model. in fact, my preference would be to change it slightly to
mirror the gflags interface and not pass around the object at all (just refer
directly to the keys you want, like you do with FLAGS).

In fact, we used a separate config outside of flags in the first place largely
to simplify deployment, which isn't really a concern of OpenHTF. It's probably
still worth keeping the two separate, but a similar interface between them would
likely be appreciated by users.

openhtf/__init__.py
jettisonjoe: +1 to @grybmadsci's proposed docstring.

test/io/output/record.json
```diff
@@ -1 +1 @@
...
..., "metadata": {"docstring": "Unit tests for the openhtf.io.output module.", "filename": "output_test.py", "test_description": "Unittest test", "test_name": "TestTest", "test_version": "1.0.0"}, "outcome": "State.PASS", "outcome_details": [], "phases": [{"attachments": {"test_attachment": {"data": "This is test attachment data.", "mimetype": null}}, "code": "@measures(\n    Measurement('measurement'),\n    Measurement('dimensions').WithDimensions(UOM['HERTZ']))\ndef dimensions(test):\n  test.measurements.measurement = 1\n  for dim in range(5):\n    test.measurements.dimensions[dim] = 1 << dim\n  test.Attach('test_attachment', 'This is test attachment data.')\n", "end_time_millis": "67891", "measured_values": {"dimensions": [["0", "1"], ["1", "2"], ["2", "4"], ["3", "8"], ["4", "16"]], "measurement": "1"}, "measurements": {"dimensions": {"dimensions": [{"uom_code": "HTZ", "uom_suffix": "Hz"}], "name": "dimensions", "outcome": "PASS"}, "measurement": {"name": "measurement", "outcome": "PASS"}}, "name": "dimensions", "result": {"phase_result": "RESULT_CONTINUE", "raised_exception": "False"}, "start_time_millis": "12345"}], "start_time_millis": "12344", "station_id": "unittest_openhtf"}
```
jettisonjoe: Yeah, I can see how updating this is gonna get annoying.

fahhem:
All the discussion aside, can I get this merged?

jettisonjoe:
Yep, the new doctring looks good. Will merge this as soon as tests finish.

fahhem:
So over the weekend? :100: :)

PR #113, opened 2016-02-09
Revert "Revert "Output to mfg-inspector.com""

grybmadsci:
This reverts commit 87b888f17746c446c41da88418b1967213a7fe39.

We must go deeper.


fahhem:
Why do you need to sort the keys? Was it to make the outputs match, but then you gave up and did FromString?

grybmadsci:
Keys have to be sorted in order for the JSON to match, FromString needs to be there because the stringification of protobufs isn't guaranteed to be an exact match across versions of protoc (is my guess).

fahhem:
Okay, for the JSON to match it makes. FromString for protobufs makes sense since there are multiple valid encodings of the same data, which can change even between computers with the same version of protoc.

LGTM (might want to rebase instead of a merge?)

Issue #120, opened 2016-02-11
Testrun proto doesn't build on some linux systems.


jettisonjoe:
[snip from Brandon's email]
:~/source/openhtf$ python setup.py bdist_wheel
running bdist_wheel
running build
running build_proto
Attempting to build proto files:
./openhtf/io/proto/testrun.proto
./openhtf/io/proto/units.proto
google/protobuf/descriptor.proto: File not found.
units.proto: Import "google/protobuf/descriptor.proto" was not found or had errors.
units.proto:22:8: "google.protobuf.EnumValueOptions" is not defined.
units.proto: "google.protobuf.EnumValueOptions" is not defined.
testrun.proto: Import "units.proto" was not found or had errors.
testrun.proto:253:12: "Units.UnitCode" is not defined.
Traceback (most recent call last):
  File "setup.py", line 144, in <module>
    install_requires=requires,
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib/python2.7/dist-packages/wheel/bdist_wheel.py", line 175, in run
    self.run_command('build')
  File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib/python2.7/distutils/command/build.py", line 128, in run
    self.run_command(cmd_name)
  File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 70, in run
    ] + protos)
  File "/usr/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/usr/bin/protoc', '--proto_path', './openhtf/io/proto', '--proto_path', '/usr/include', '--python_out', './openhtf/io/proto', './openhtf/io/proto/testrun.proto', './openhtf/io/proto/units.proto']' returned non-zero exit status 1
~/source/openhtf$ protoc --version
[...]

PR #122, opened 2016-02-11
Grab the protoc prefix from pkg-config

fahhem:
One possible solution to #120

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@
 
 import glob
 import os
+import platform
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


grybmadsci:
it's preferred style to write platform-independent code whenever possible, even if other adjacent parts of the code are necessarily platform-dependent.

jettisonjoe:
One of the main reasons for this whole "find protoc" enterprise instead of including our own prebuilt protoc binary is to be multi-platform. We have a workaround users can apply using symlinks right now, so I think we should take our time here and craft a cross-platform solution.

One of my objectives is to make openhtf truly cross-platform. Let's not introduce _more_ platform specific code if we can help it.

fahhem:
Turns out I was wrong, pkg-config works on Windows and OS X. However, using os.path for /usr still doesn't make sense. If there was a default path for Windows, then I could check `platform.system()` and return the correct default path there.

setup.py
```diff
@@ -49,8 +49,15 @@ class BuildProtoCommand(Command):
...
+    try:
+      prefix = subprocess.check_output(
+          'pkg-config --variable prefix protobuf'.split()).strip()
+    except subprocess.CalledProcessError:
+      # Default to /usr?
+      prefix = '/usr'
+
+    self.protoc = os.path.join(prefix, 'bin', 'protoc')
+    self.protodir = os.path.join(prefix, 'include')
     self.indir = './openhtf/io/proto'
```
grybmadsci: these should also use os.path.join, and should use os.getcwd() instead of '.'

setup.py
```diff
@@ -49,8 +49,15 @@ class BuildProtoCommand(Command):
...
 
   def initialize_options(self):
-    self.protoc = '/usr/bin/protoc'
-    self.protodir = '/usr/include'
+    try:
+      prefix = subprocess.check_output(
+          'pkg-config --variable prefix protobuf'.split()).strip()
+    except subprocess.CalledProcessError:
+      # Default to /usr?
+      prefix = '/usr'
```
grybmadsci: at least check that the system is *nix when applying this default.  for completeness, it would be nice to provide a default on Windows, but you can just raise for now.

the requirement right now isn't to fully support windows, but rather that when run on windows, failures should be obvious (like a raise here), not obscured behind things like platform-specific path defaults and path separators.  that said, windows support is a fairly high priority after the frontend.

setup.py
fahhem: last time I made changes to 'code written a week ago' I got lambasted :)

setup.py
fahhem: Done

fahhem:
Tests have been passing for about 9 days now :)

PR #123, opened 2016-02-11
Clean up setup.py and add support for JSON encoded private keys.

grybmadsci:


example/hello_world.py
```diff
@@ -95,9 +97,13 @@ def attachments(test):
...
-  # Example of how to upload to mfg-inspector.  Replace user email and key,
-  # these are dummy values.
-  #test.AddOutputCallback(output.UploadToMfgInspector(
-  #  'foobarbaz_gaia_id@developer.gserviceaccount.com',
-  #  open('my-upload-key.p12', 'r').read()))
+  # Example of how to upload to mfg-inspector.  Replace filename with your
+  # JSON-formatted private key downloaded from Google Developers Console
+  # when you created the Service Account you intend to use, or name it
+  # 'my_private_key.json'.
+  if os.path.isfile('my_private_key.json'):
```
fahhem: are these named `.json` when downloaded or something else?

setup.py
```diff
@@ -141,5 +142,12 @@ def run_tests(self):
         'clean': CleanCommand,
         'test': PyTestCommand,
     },
-    install_requires=requires,
+    install_requires=install_requires,
+    setup_requires=[
+        'wheel==0.29.0',
+    ],
+    tests_require=[
```
fahhem: I've had trouble getting travis to play well with tests_require, which is why mock and pytest are in test_reqs.txt instead.

example/hello_world.py
grybmadsci: they are named <something>.json when downloaded.

setup.py
grybmadsci: hm, they can be in test_reqs.txt to make travis work, but manually running python setup.py test doesn't work if they aren't also included somewhere here, and tests_require is a more correct place than install_requires.  i'd prefer not to make the user manually pip install -r test_reqs.txt when setuptools provides a mechanism specifically for this purpose.

jettisonjoe:
LGTM

PR #128, opened 2016-02-12
Explain certain exceptions in build_proto

fahhem:


setup.py
```diff
@@ -61,15 +61,28 @@ def run(self):
...
           self.protoc,
           '--proto_path', self.indir,
           '--proto_path', self.protodir,
           '--python_out', self.outdir,
-          ] + protos)
+      ] + protos
+      try:
+        subprocess.check_call(cmd)
+      except OSError as e:
+        if e.errno == 2:
```
grybmadsci: prefer errno.ENOENT instead of 2.

setup.py
```diff
@@ -61,15 +61,28 @@ def run(self):
...
+          print 'You need to install the protobuf compiler.'
```
grybmadsci: 'You may need to...' it's a bit of an assumption to make that ENOENT was definitely because of a missing protoc.

grybmadsci:
LGTM, pending minor nits.

setup.py
fahhem: Done

setup.py
fahhem: Done. ENOENT means the binary didn't exist, so protoc wasn't at the location so I added that message.

setup.py
```diff
@@ -61,15 +62,30 @@ def run(self):
...
+      ] + protos
+      try:
+        subprocess.check_call(cmd)
+      except OSError as e:
+        if e.errno == errno.ENOENT:
+          print 'You may need to install the protobuf compiler.'
+          print ('On many Linux systems, this is done by installing the '
+                 '"protobuf-compiler" and "libprotobuf-dev" packages.')
+          print ('If you have it installed, then we cannot find it at '
+                 '%s' self.protoc)
```
jettisonjoe: Picking nits here, but this is a little verbose. How about something like:
```
          print ('Could not find the protobuf compiler at '
                 '%s.' % self.protoc)
          print ('On many Linux systems, this is fixed by installing the '
                 '"protobuf-compiler" and "libprotobuf-dev" packages.')
```

fahhem: Fixed

setup.py
```diff
@@ -61,15 +62,28 @@ def run(self):
...
           '--proto_path', self.protodir,
           '--python_out', self.outdir,
-          ] + protos)
+      ] + protos
+      try:
+        subprocess.check_call(cmd)
+      except OSError as e:
+        if e.errno == errno.ENOENT:
+          print 'Could not find the protobuf compiler at %s' % self.protoc
+          print ('On many Linux systems, this is done by installing the '
```
jettisonjoe: s/done/fixed/

fahhem:
Fixed. I only changed a string, so the test results from earlier should be sufficient :)

setup.py
```diff
@@ -61,15 +62,28 @@ def run(self):
...
+        if e.errno == errno.ENOENT:
+          print 'Could not find the protobuf compiler at %s' % self.protoc
+          print ('On many Linux systems, this is fixed by installing the '
+                 '"protobuf-compiler" and "libprotobuf-dev" packages.')
+        raise
+      except subprocess.CalledProcessError:
+        print 'Could not build proto files.'
+        print ('This could be due to missing helper files. On many Linux '
+               'systems, this is fixed by installing the '
+               '"libprotobuf-dev" package.')
```
grybmadsci: oh, uh, this should raise too... otherwise it silently keeps going and it looks like the build succeeded.

PR #129, opened 2016-02-12
Cleanup outcome details in inspector output

fahhem:


grybmadsci:
Nice cleanup, thanks!

Issue #131, opened 2016-02-13
Separate test initialization from execution


fahhem:
From the OpenHTF weekly meetings back in January, this was one of the things discussed. If there is still agreement on this, then I'll continue this work.

grybmadsci:
#132 went in quite a while ago, are there any more pending actions for this or can we close it?  if there are, can we get at least a high-level list of requirements to close this issue?

fahhem:
Yeah, the eventual future will be that a Test instance had a PlugManager, PhaseExecutor, TestExecutor, etc all setup in `__init__()` and ready to go in `Execute()`, which would allow a motivated user to replace one or more such pieces while keeping the interface intact to take advantage of everything else in OpenHTF.

Sorry about the ambiguity, I need to look at the code again to solidify it, which tends to end up with another PR for something else :)

fahhem:
@grybmadsci, I can't assign this to myself, can you assign it to me?

jettisonjoe:
This is done.

Issue #106, opened 2016-02-04
PhaseExecutor and PlugManager should not be mutablerecords.Record subclasses


grybmadsci:
Executor and Manager obviously imply more functionality than a POD type.  I think we should discourage using records for anything but POD types (potentially with minimal additional convenience functionality).  

Using mutablerecords outside of POD types translates to "we don't like the way Python classes work, so we redid it", and I think that's frustratingly obfuscated for any outside developers looking at the code.

That said, I think this is super low priority and more of an aesthetic change.

fahhem:
This seems done at HEAD

PR #132, opened 2016-02-13
Separating init from exec

fahhem:
Step 1 of #131


example/hello_world.py
```diff
@@ -94,10 +94,11 @@ def attachments(test):
   test.AddOutputCallback(OutputToJSON(
       './%(dut_id)s.%(start_time_millis)s.json', indent=4))
   test.AddOutputCallback(output.OutputToTestRunProto(
-      './%(dut_id)s.%(start_time_millis)s.json'))
+      './%(dut_id)s.%(start_time_millis)s.pb'))
```
grybmadsci: oh, whoops, thanks!

openhtf/__init__.py
```diff
@@ -45,6 +46,105 @@ class InvalidTestPhaseError(Exception):
...
+    server.Start()
+    executor.Start()
+    executor.Wait()
+    server.Stop()
+
+
+def SetupFramework():
+  """Sets up various bits of the framework. Only needs to be called once."""
+  try:
+    FLAGS(sys.argv)  # parse flags
```
grybmadsci: hm, i believe you're supposed to assign the result to sys.argv in order to preserve any non-gflags stuff?

openhtf/exe/phasemanager.py
```diff
@@ -162,7 +162,7 @@ def _ExecuteOnePhase(self, phase):
         'Executing phase %s with plugs %s', phase.func.__name__, data.plugs)
 
     self._test_state.running_phase = test_record.PhaseRecord(
-        phase.func.__name__, phase.source, docstring=phase.func.__doc__)
+        phase.func.__name__, phase.code_info.sourcecode, docstring=phase.func.__doc__)
```
grybmadsci: hm, probably better to plumb through all of the CodeInfo attributes, don't you think?  you could even encapsulate all three of these and store a CodeInfo directly on the PhaseRecord if you wanted.

openhtf/exe/test_state.py
```diff
@@ -67,11 +70,11 @@ def __init__(self, config, test, plugs, dut_id):
     self.record = test_record.TestRecord(
         dut_id=dut_id, station_id=config.station_id,
         metadata={
-            'code': test.code,
-            'filename': test.filename,
-            'docstring': test.docstring
+            'code': test.code_info.sourcecode,
```
grybmadsci: same comment here, you could just do 'code_info': test.code_info, i think?

openhtf/__init__.py
fahhem: This is what was there before, was it wrong before?

openhtf/exe/phasemanager.py
fahhem: Done

openhtf/exe/test_state.py
fahhem: Done

openhtf/__init__.py
grybmadsci: ah, yeah it was wrong before, just nobody noticed because it's uncommon to use other argv args, but you can see from a gflags docstring here:
https://github.com/google/python-gflags/blob/master/gflags/flagvalues.py#L75

fahhem:
Any chance I could get this in soon?

openhtf/__init__.py
```diff
@@ -45,6 +46,106 @@ class InvalidTestPhaseError(Exception):
   """Raised when an invalid method is decorated."""
 
 
+def Test(*phases, **metadata):
+  return TestInfo(phases, metadata=metadata)
+
+
+class TestInfo(mutablerecords.Record(
```
grybmadsci: This is not a record; it has an 'Execute' method.  Please use standard Python syntax for class definitions and only use mutablerecord for POD types.

fahhem: While I agree with #106 in general, I don't agree here. This is a TestInfo class, the Execute method is a convenience wrapper around two other classes that do actual work. I would prefer an `openhtf.ExecuteTest(...)` function to do that if I wanted to be a purist

fahhem: I make `TestInfo.Execute()` call `ExecuteTest()` which makes it more like a record with a helper function. The point of all this is to make the `TestInfo` class become stateless and separate execution from initialization, that was my proposal to the openhtf meeting and that was accepted. If you disagree, I'd rather we talk about that than this detail

grybmadsci: I'm okay with that separation - I don't really feel too strongly about it either way.  My point is that there are two methods on this class that have side effects beyond getting/setting its own internal state, `OutputTestRecord()` and `Execute()`.  Making `Execute()` a thin wrapper around a global-scope function doesn't change the fact that calling it has side effects.  `GetPlugTypeMap()` is a great example of what constitutes a 'helper function' on a POD type; in fact you could even make it an `@property` and call it just `plug_type_map` (oh, looking at it, it already was, not sure why you changed this).

Making `Test()` just an arg unpacker for a `TestInfo` record forces a functional user-facing interface over an object-oriented one (ie, make the user call `openhtf.ExecuteTest(foo)` instead of `foo.Execute()`, as you described).  I have a preference for the latter, I think it's more succinct and straightforward, which is why the `Test()` class was structured the way it was in the first place.  I'm okay with pulling out the initialization/state tracking into a separate record class, but the user-facing interface should remain object-oriented, and breaking convention of POD types to provide that interface isn't acceptable.

The two solutions I'm okay with are promoting this to a real class, not a record (but that basically just puts you back where it was before), or keeping a `Test` class whose only purpose is to provide a true object-oriented user interface, and it can keep a `TestInfo` to encapsulate all init/state stuff.  This is generally how functional-to-OO adapters work.

fahhem: I'm dealing with something at work right now, but I'm good with the second route (`Test` class stores a `TestInfo` record and does the `Execute()`). The point was to enable a different wrapper to create `TestInfo` instances and replace the parts of the actual execution that it wants without needing OpenHTF to support/maintain it.

fahhem: Fixed it, now it's a dynamic `Test()` with a static `TestInfo()` record inside. I'll clean up the unnecessary properties in another PR since we have enough big PR's in the queue that modifying more files in this PR is just going to make more work for one of us

example/hello_world.py
```diff
@@ -106,4 +106,5 @@ def attachments(test):
   #    test.AddOutputCallback(mfg_inspector.UploadToMfgInspector.from_json(
   #        json.load(json_file)))
 
-  test.Execute(test_start=triggers.PromptForTestStart())
+  test.test_start = triggers.PromptForTestStart()
```
jettisonjoe: It looks like you're still using the kwargs to `Execute()` for this below.

openhtf/__init__.py
```diff
@@ -45,6 +46,137 @@ class InvalidTestPhaseError(Exception):
...
+    executor = exe.TestExecutor(conf.Config(), self, test_start, test_stop)
+    server = http_api.Server(executor)
+    StopOnSigInt([server.Stop, executor.Stop])
+    server.Start()
+    executor.Start()
+    executor.Wait()
+    server.Stop()
+
+
+class TestInfo(mutablerecords.Record(
```
jettisonjoe: `TestTemplate` instead of `TestInfo`; more or less confusing? I'm hunting for a name that's a little more specific. Although the generic name isn't a showstopper for me. Just a my-2-cents thing.

openhtf/__init__.py
```diff
@@ -114,103 +248,12 @@ def __call__(self, phase_data):
...
-
-  Args:
-    *phases: The ordered list of phases to execute for this test.
-    **metadata: Any metadata that should be associated with test records.
-  """
-
-  def __init__(self, *phases, **metadata):
-    """Creates a new Test to be executed.
+def StopOnSigInt(callbacks):
+  """Handles SigInt by calling the given callbacks."""
```
jettisonjoe: `StopOnSigInt`'s docstring looks like it should go with `_Handler`, while `StopOnSigInt` itself should be more like "Create a SIGINT handler..." The point is that there's nothing in `_Handler` that ties it to executor etc. specifically.

openhtf/io/test_record.py
```diff
@@ -71,3 +74,23 @@ class PhaseRecord(  # pylint: disable=too-few-public-methods,no-init
 
   See measurements.Record.GetValues() for more information.
   """
+
+
+class CodeInfo(mutablerecords.Record(
```
jettisonjoe: Side note, there's a new requirement from my team to be able to disable sourcecode capturing on test phases. Not sure where the best place to do that check is, but wanted to get it out there.

grybmadsci: i snuck that into #168 here:
https://github.com/google/openhtf/pull/168/files#diff-84ac26137df2222d79ee6e7afa7a3cc0R40
it may need some tweaks still to behave the way you guys want, but it's in there at least.

openhtf/__init__.py
```diff
@@ -45,6 +46,137 @@ class InvalidTestPhaseError(Exception):
...
+        ((plug.name, plug.cls) for plug in phase.plugs)
+        for phase in self.phases):
+      if (plug in plug_type_map and
+          plug_type is not plug_type_map[plug]):
+        raise plugs.DuplicatePlugError(
+            'Duplicate plug with different type: %s' % plug)
+      plug_type_map[plug] = plug_type
+    return plug_type_map
+
+
```
grybmadsci: Extra whitespace line here

openhtf/__init__.py
```diff
@@ -45,6 +46,137 @@ class InvalidTestPhaseError(Exception):
...
+    executor.Start()
+    executor.Wait()
+    server.Stop()
+
+
+class TestInfo(mutablerecords.Record(
+    'Test', ['phases', 'code_info'], {'metadata': dict})):
+  """An object that represents the reusable portions of an OpenHTF test.
+
+  This object encapsulates the static test information that is set once and used
```
grybmadsci: If it's set once, does it need to be a `Record` instead of a `namedtuple`?  It looks like the only place it gets instantiated, `metadata` is always provided, so neither the mutability nor the the default value mechanism of the `Record` are used, it looks like?

grybmadsci:
LGTM after this last round of minor comments is addressed, will merge before #168 so I don't create a bunch of merge conflicts for you :)

openhtf/__init__.py
fahhem: `TestData`?

openhtf/__init__.py

openhtf/__init__.py

openhtf/__init__.py
fahhem: namedtuple makes you do stuff with `__new__` since it's crazy like that, but I switched anyway since it's not meant to be mutable after all

fahhem:
Thank you :)

openhtf/__init__.py
jettisonjoe: Well, I wouldn't say `TestData` is any more descriptive than `TestInfo`, but it does have a nice linguistic parity with `PhaseData`.

jettisonjoe:
LGTM

PR #135, opened 2016-02-18
upload test record to server

amyxchen:


openhtf/names.py
```diff
@@ -55,5 +56,6 @@
 # Classes used in writing test scripts.
 Measurement = openhtf.util.measurements.Measurement
 OutputToJSON = openhtf.io.output.json_factory.OutputToJSON
+UploadJSON = openhtf.io.output.upload_json.UploadJSON
```
fahhem: In the last meeting, we discussed the 'vision' for names.py and generally agreed it should only include modules that are vendor-agnostic. While the JSON is vendor-agnostic, upload_json is centered around uploading to the 'CM' system, so it doesn't belong in names.py, sorry!

jettisonjoe: Agreed. We should be sparing with names since we're recommending people to import * from it.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,81 @@
...
+    self.project = None
+    self.cm_server = None
+
+
+  # pylint: disable=invalid-name
+  def __call__(self, test_record):
+    super(UploadJSON, self).__call__(test_record)
+  # pylint: enable=invalid-name
+    if not self.data_server:
+      self.project = conf.Config().project
```
jettisonjoe: See the config fixes in that PR from last week. You should call Config() once then use the returned result to access keys rather than calling Config() every time you need to access a key.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,81 @@
...
+      if self.project and server:
+        self.data_server = server + self.project
+      elif server:
+        serlf.data_server = server
+      else:
+        raise ValueError('No data_server set')
+      
+      self.cm_server = conf.Config().cm_server
+      if self.cm_server:
+        from openhtf.io.output.upload_cm import UploadTestToCM as upload_cm
```
jettisonjoe: This CM-specific stuff shouldn't be in openhtf but in private repos instead. In the private repo, you can create an output module that calls this one rather than the other way around.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,81 @@
...
+          if rc == 200:
+            os.remove(f)
+            _LOG.info('Test record successfully loaded to servers.')
+          else:
+            _LOG.warning('Error on upload test_record to cm server, rc:%d', rc)
+        else:
+          os.remove(f)
+          _LOG.info('Test record successfully loaded to server.')
+      else:
+        _LOG.warning('Error on upload test_record to server rc:%d', rc_data)
```
jettisonjoe: I would log some identifying info about the files that succeed or fail to upload. Maybe filename?

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,81 @@
...
+      if self.cm_server:
+        from openhtf.io.output.upload_cm import UploadTestToCM as upload_cm
+
+    files = [f for f in os.listdir('.') if os.path.isfile(f)]                
+    for f in files:
+      resp = requests.post(self.data_server, data=open(f, 'rb'),
+                             headers=self.headers)
+      rc_data = resp.status_code
+      if rc_data == 200: 
+        if self.cm_server:
```
jettisonjoe: Again, this CM-specific stuff should be split out into a private output module which can import and use this module if necessary.

jettisonjoe: Instead of nesting everything in a giant if- block, I would raise or return on failure instead to make things more readable and save levels of indent i.e. hard-to-follow conditional logic.

openhtf/names.py
amyxchen: This upload_json is not centered to uploading to any "CM", it is a general loader to any data server. uploading test results to a server is a  necessary step to complete the whole circle of testing. It should have the same weight as OutputToJSON.

jettisonjoe: Hmm. Perhaps we should remove OutputToJSON from names as well.

fahhem: If there were a generic way to 'upload', then sure. But at the moment, your UploadJSON uploads in a specific manner, and all upload paths will be different. Be it over HTTP, raw TCP, etc and with OAuth, OAuth2, Basic Auth, etc. There are a ton of variations and trying to make a generic one will never be generic enough. Instead, we provide json output as human-readable only (not even guaranteed to be the same structure across versions, yet) and these other modules should only do the data-conversion necessary to convert the test_record data into their format and then upload.

openhtf/io/output/upload_json.py
amyxchen: This has been changed, all cm related checks are removed.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
+"""Module for uploading JSON-formatted test record files to server."""
```
grybmadsci: can we change the wording on this a little, so that instead of 'upload to server', it's 'HttpUpload'?  i think the 'to server' part is implied, but the upload mechanism is ambiguous.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+import logging
+import requests
+
+from json import JSONEncoder
+from openhtf import conf
+from openhtf.io.output.json_factory import OutputToJSON
+
+conf.Declare('project', 'project name')
+conf.Declare('data_server', 'data server')
```
grybmadsci: can we rename this to http_upload_target or something similar?

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+import os
+import logging
+import requests
+
+from json import JSONEncoder
+from openhtf import conf
+from openhtf.io.output.json_factory import OutputToJSON
+
+conf.Declare('project', 'project name')
```
grybmadsci: it looks like neither project nor data_server is ever used alone, so i think they can both get rolled into the one config option 'http_upload_target'?

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+  Args:
+    filename_pattern: A format string specifying the filename to write to,
+      will be formatted with the Test Record as a dictionary.
+    inline_attachments: Whether attachments should be included inline in the
+      output.  Set to False if you expect to have large binary attachments.
+  """
+
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
+    super(UploadJSON, self).__init__(filename_pattern, inline_attachments, **kwargs)
+    self.headers = {'Accept' : 'application/json',
```
grybmadsci: if this is a static value, better to put it on the class as HTTP_HEADERS

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+from json import JSONEncoder
+from openhtf import conf
+from openhtf.io.output.json_factory import OutputToJSON
+
+conf.Declare('project', 'project name')
+conf.Declare('data_server', 'data server')
+_LOG = logging.getLogger(__name__)
+
+class UploadJSON(OutputToJSON):
```
grybmadsci: HttpUploadJSON

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+
+  An example filename_pattern might be:
+    '/data/test_records/%(dut_id)s.%(start_time_millis)s'
+
+  To use this output mechanism:
+    test = openhtf.Test(PhaseOne, PhaseTwo)
+    test.AddOutputCallback(UploadJson(
+        '/data/test_records/%(dut_id)s.%(start_time_millis)s'))
+  
+    in config.yaml:
```
grybmadsci: 'In configuration (probably a .yaml file):'

config doesn't *have* to come from a yaml file, and that file doesn't have to be named config.yaml, so this may be misleading to some users.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+       data_server:  http://your_data_server_entry
+  Args:
+    filename_pattern: A format string specifying the filename to write to,
+      will be formatted with the Test Record as a dictionary.
+    inline_attachments: Whether attachments should be included inline in the
+      output.  Set to False if you expect to have large binary attachments.
+  """
+
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
```
grybmadsci:
```
@conf.InjectPositionalArgs
def __init__(self, http_upload_target, filename_pattern=None, inline_attachments=True, **kwargs):
  super(UploadJSON, self).__init__(filename_pattern, inline_attachments, **kwargs)
  self.upload_target = http_upload_target
```
And then you can do this below, instead of instantiating the conf.Config() object:
```
...
requests.post(self.upload_target, data=..., headers=self.HTTP_HEADERS)
...
```

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+      config = conf.Config()
+      project = config.project
+      server = config.data_server
+   
+      if server:
+        self.data_server = server + project
+      else:
+        raise ValueError('No data_server set')
+    
+    for f in self.IterFile(self.file_dir):
```
grybmadsci: this terrifies me because you're potentially attempting to upload test runs that are not the one you just generated.  i'm on the fence about subclassing OutputToJSON like this to get 'fallback to file output' behavior, but with the current state of the output callback mechanism, it's really the most sane way to do that kind of fallback.

Still, we definitely can't use the output callback as a generic upload service for any test records that might be hanging around in the directory, that's a job for a separate daemon.  Since there's no return value specified by the output callback convention, I suggest updating the OutputToJSON callback to return the filename that was just generated, and then use the return value from the super() call above to explicitly know which file to upload.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+      resp = requests.post(self.data_server, data=open(f, 'rb'),
+                             headers=self.headers)
+      if resp.status_code == 200:
+        os.remove(f)
+        _LOG.info('Test record successfully loaded to servers.')
```
grybmadsci: maybe 'Test record successfully upload to %s', self.upload_target

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+      else:
+        _LOG.warning('Error on upload test_record to server rc:%d, file:%s',
```
grybmadsci: I think you're going to get a really long log line putting this all on one line, consider splitting it into:
```
_LOG.warning('Error uploading test record to %s: %s - %s', self.upload_target, resp.status_code, resp.reason)
_LOG.warning('Test record saved to %s', record_filename)
```

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
+    super(UploadJSON, self).__init__(filename_pattern, inline_attachments, **kwargs)
+    self.headers = {'Accept' : 'application/json',
+                 'Content-Type' : 'application/json'}
+    self.data_server = None
+    self.file_dir = filename_pattern.split('%')[0]
+
+  # pylint: disable=invalid-name
+  def __call__(self, test_record):
+    super(UploadJSON, self).__call__(test_record)
```
grybmadsci: no need to invoke __call__ explicitly:
```
record_filename = super(HttpUploadJSON, self)(test_record)
```
note, you'll have to update OutputToJSON to return the filename generated.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+from openhtf.io.output.json_factory import OutputToJSON
+
+conf.Declare('project', 'project name')
+conf.Declare('data_server', 'data server')
+_LOG = logging.getLogger(__name__)
+
+class UploadJSON(OutputToJSON):
+  """Return an output callback that upload JSON Test Records to data server.
+
+  An example filename_pattern might be:
```
grybmadsci: Note in the docstring that filename_pattern determines the filename to fallback to if the upload fails.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+    test = openhtf.Test(PhaseOne, PhaseTwo)
+    test.AddOutputCallback(UploadJson(
+        '/data/test_records/%(dut_id)s.%(start_time_millis)s'))
+  
+    in config.yaml:
+       data_server:  http://your_data_server_entry
+  Args:
+    filename_pattern: A format string specifying the filename to write to,
+      will be formatted with the Test Record as a dictionary.
+    inline_attachments: Whether attachments should be included inline in the
```
grybmadsci: the inline_attachments mechanism needs work, it's beyond the scope of this PR, but can you update the comment to note that if this is False, you *lose* the attachments.

openhtf/io/output/upload_json.py
```diff
@@ -0,0 +1,72 @@
...
+      output.  Set to False if you expect to have large binary attachments.
+  """
+
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
+    super(UploadJSON, self).__init__(filename_pattern, inline_attachments, **kwargs)
```
grybmadsci: filename_pattern defaults to None for OutputToJSON so that the encode() method can be called without writing to a file, but in order to invoke __call__, it *must* be set (you'll get an AssertionError otherwise).

If there's no filename_pattern passed in to your constructor as a fallback mechanism, you should set one using tempfile to generate a temporary file or directory to use instead and pass that to the super() constructor.

openhtf/io/output/upload_json.py
amyxchen: Can you suggest me how to word the ***WARNING *** part.

openhtf/io/output/upload_json.py
amyxchen: I change the class to be HttpUploadJSON, still use JSON is to be consistent with OutputJSON. file will be http_upload.py.

openhtf/io/output/upload_json.py
amyxchen: Oh, I understand why you want the "target" here. the way to use it  is duo to the server entry point may have other info between data_server and project, if put them together, code needs to do some parsing to meet the requirement

openhtf/io/output/upload_json.py
amyxchen: sounds good.

amyxchen: will change

openhtf/io/output/upload_json.py
amyxchen: Yes

openhtf/io/output/upload_json.py
amyxchen: I will think about it and find a way to this.

example/hello_world.py
```diff
@@ -93,6 +93,7 @@ def attachments(test):
       # but you can include any metadata fields.
       test_name='MyTest', test_description='OpenHTF Example Test',
       test_version='1.0.0')
+  OutputToJSON = output.json_factory.OutputToJSON
```
jettisonjoe: Name exposition (is that a term?) like this seems better done with an `import...as` to me.

openhtf/io/output/upload_json.py
jettisonjoe: How about this:
```
***WARNING***
This module sends test records over the network as un-encrypted plaintext. It
should only be used on secure internal networks, and then only if HTTPS is not
available.
```

openhtf/io/output/upload_json.py
jettisonjoe: I wonder if 'path' would be more clear than 'target'. In fact, maybe we should take a step back and think about whether it even makes sense to break down the endpoint into parts rather than to just have an `http_upload_url` that encodes the host, port, and path. Thinking back, it seems like we're always taking a full URL for uploads and breaking it down by hand to put it into config anyway. Why not cut out the middleman? Open to discussion.

grybmadsci: Agreed, no need to complexify the configuration unnecessarily (this is our fault for setting a bad example, we've done this a lot in the past), let's roll this into a single string URL.

jettisonjoe:
@amyxchen checking in on this PR. Need any help to work on addressing the reviewer comments?

PR #136, opened 2016-02-19
Switch to UNECE units

fahhem:
Also generate units.py from units.proto so we don't require protoc for normal usage but the files don't diverge still. Since we're moving to making the protos part of an 'extra' in setup.py, I figured I'd remove one dependency while I was at it. Let me know if you'd rather go another way with the generation.


openhtf/io/proto/generate_units.py
```diff
@@ -0,0 +1,70 @@
...
+
+def unit_constructor(unit):
+  exts = unit.GetOptions().Extensions
+  return 'Unit(%s, %s)' % (
+      string_or_none(exts[units_pb2.uom_code]),
+      string_or_none(exts[units_pb2.uom_suffix]))
+
+
+def main():
+  with io.open(sys.argv[1], 'wb') as output:
```
grybmadsci: can you do a quick sanity check on len(sys.argv) and print a basic usage line on error?  doesn't have to be too verbose, just a quick one-liner

openhtf/io/proto/generate_units.py
```diff
@@ -0,0 +1,70 @@
+"""Generates util/units.py from units.proto!
```
grybmadsci: oh god, this is moderately painful... but probably the best way to do this, in fact i'm impressed at how not-hacky you managed to make this.

one thing i'd like to change, though, if you don't mind a little extra work ;)  if we're going to generate units.py, we should probably do so in the build process and not have a units.py checked in to the repo.  you could roll it into build_proto if you want to save some boilerplate (i don't think we need a build_units target specifically for just this, but entirely up to you).

jettisonjoe:
I hate to throw a wrench in this, but what do you guys think about reversing it so the .proto is built from the .py? I understand why that _seems_ weird (python > proto > python), but I would _love_ it if we would remove protoc (and the build_proto step) as a dependency from the core package, and only require it with the mfg-inspector extra.

grybmadsci:
hm, i don't think depending on protoc for the build process is all that onerous, as anyone who is building a package should be capable of installing system utilities, and we don't depend on a specific version of protoc now (which *was* pretty annoying, i think).  i don't feel super strongly about it either way. in fact, if we're going to do something like that, i'd say store the units in a CSV and have the utility generate both python and proto, which i'm surprisingly okay with despite the 'CSV' part...

jettisonjoe:
@grybmadsci and I talked about it in person over the weekend as well, and I think we're now all in agreement about keeping a CSV and generating both the .py and the .proto off of that.

ahaberlach:
Ewww.  But ok.

grybmadsci:
Hm, we could keep it in a JSON, that would be a little nicer, I think, and
equally parseable.

fahhem:
Assuming the future of this code will be internal->GitHub publishing, I'd like to stick with the proto as the source and a checked in python file (only in GitHub)

jettisonjoe:
Per our meeting 2 weeks ago the clear consensus is to _not_  pursue internal->GitHub publishing at this time, and keep GitHub as the source of truth for the project. So I think we can say that we want to keep unit definitions in a JSON and generate both the .py and the .proto off of that.

jettisonjoe:
So folks, should we do something with this PR?

fahhem:
The internal format is moving away from this enum system, partly because proto3 doesn't have extensions, and partly because it's provided far less value than the pain. We're moving towards taking the 2 strings directly, the display units and the UNECE string, both being optional.

With that in mind, units.proto will be going away and so units.py can just be the same as it is now. Optionally, we could store the 3 values (human-readably name, display string, and UNECE code) in a csv and generate the Python?

jettisonjoe:
Hmm, fair enough. Maybe we just leave it pretty much as-is structure-wise and just switch over to UNECE units.

jettisonjoe:
@fahhem Did you want to continue this one, or pass the torch?

jettisonjoe:
@fahhem checking in on this. I'm cool with closing this and filing an issue with the new POR if you want to let whoever gets to it first take care of it, or waiting till you have a chance to get to it.

fahhem:
So we just want to switch to UNECE units, right? We can close this and create a new issue since this PR is way off base anyway

jettisonjoe:
Cool, sounds good. Opened #298 to track the feature request, and closing this PR.

PR #139, opened 2016-02-23
First pass at windows support

grybmadsci:
Ran example test on Windows and Linux, verified error messages when usb_plugs extra not installed on Linux.

- Move adb/libusb plugs into an extra so we don't depend on M2Crypto unless we want those plugs
- Add fallback to raw_input() for commandline prompt on Windows because select() isn't supported for files.
- Update example hello_world.py to play nicer on Windows (and comment out testrun output being enabled by default).
- Move some things around so we don't by default depend on protos.
- Note: I left the dependency on protobuf in install_requires because Farz has a pending change to move that around anyway, and the pip dependency installs easily, it's the actual system library that is troublesome and platform-dependent.


example/hello_world.py
```diff
@@ -81,10 +81,21 @@ def dimensions(test):
 
 def attachments(test):
   test.Attach('test_attachment', 'This is test attachment data.')
-  with tempfile.NamedTemporaryFile() as f:
-    f.write('This is a file attachment')
-    f.flush()
-    test.AttachFromFile(f.name)
+  # Windows doesn't allow reading from an open NamedTemporaryFile, so we
+  # have to do this the hard way.  Typically in a real test, you would have
+  # invoked some external program that generated an output file.
```
fahhem: what about using delete=False, then replacing f.flush() with f.close()? might make this less crazy?

example/hello_world.py
```diff
@@ -95,15 +106,15 @@ def attachments(test):
       test_version='1.0.0')
   test.AddOutputCallback(OutputToJSON(
       './%(dut_id)s.%(start_time_millis)s.json', indent=4))
-  test.AddOutputCallback(output.OutputToTestRunProto(
-      './%(dut_id)s.%(start_time_millis)s.json'))
+  #test.AddOutputCallback(output.OutputToTestRunProto(
```
fahhem: should probably undo this commenting out?

openhtf/io/user_input.py
```diff
@@ -133,20 +143,27 @@ def Stop(self):
...
+
+        while not self._stopped:
+          inputs, _, _ = select.select([sys.stdin], [], [], 0.001)
+          for stream in inputs:
+            if stream is sys.stdin:
+              response = sys.stdin.readline().rstrip()
+              self._callback(response)
+              return
+    finally:
+      self._stopped = True
```
fahhem: Why does this need to be set? Stop() just prints something out if it wasn't (which doesn't make sense anyway)

setup.py
```diff
@@ -143,7 +141,7 @@ def run_tests(self):
 
 setup(
     name='openhtf',
-    version='1.0',
+    version='1.1',
```
fahhem: we haven't been updating this because it's not on PyPI yet. I don't think we want to update it until we've got releases happening (tags, a PyPI upload process, etc)

example/hello_world.py
grybmadsci: i would prefer the example by default not depend on the protos being there, but it's a handy reference to have there, which is why i left it in but commented out.

openhtf/io/user_input.py
grybmadsci: i'm guessing it's because Stop() can be called multiple times from different threads.  I put it here for consistency with the previous implementation.  we can clean this up in a separate PR if you want, for now i'd rather just leave the same behavior as it had.

setup.py
grybmadsci: we have releases happening in that i am building sdist tarballs and have them lying around, so the distinction between being installable on windows and not is somewhat important to me.  what's the cost of incrementing this?

example/hello_world.py
grybmadsci: done.

jettisonjoe: As discussed in meeting, let's just make a static file in example/ for this instead of generating on the fly, so as to keep this Windows nightmare code out of the example. We don't want users to follow our example and write stuff like this.

openhtf/io/user_input.py
```diff
@@ -24,14 +24,20 @@
...
 import select
 import sys
-import termios
 import threading
 import uuid
 
 import gflags
 
+if platform.system() != 'Windows':
+  import termios
```
jettisonjoe: As discussed in meeting, we wanted to consider using pyreadline as a better cross-platform prompting alternative.

setup.py
jettisonjoe: Decided in meeting to bump this _down_ until we get in pypi.

openhtf/io/user_input.py
grybmadsci: mind if we make this a separate feature request? it's a nontrivial amount of work and i'd like to get this in.

jettisonjoe: SGTM.

jettisonjoe:
LGTM. Merging. Thanks for doing this, @grybmadsci!

PR #143, opened 2016-02-24
Logging updates

grybmadsci:
I'd like to propose a few changes to the logging mechanisms we use.  A summary of the changes here:

- Change flag names to be a little more precise, and to use hyphens instead of underscore (it's more standard, even if it makes the interface to gflags a little uglier for us).
- Provide better isolation between test logs and framework logs by outputting test logs to stdout and framework logs to stderr, and don't propagate test logs to framework logs.
- Other changes are cleanup/docs updates.

I'm fielding this for discussion, happy to reconsider if there are objections.  Notably, I want to make sure to get an LGTM from wallaceb before merging because this may break some of his things if he's using these flags.  Let me know what you guys think.


wallacbe:
This shouldn't break me, I'm currently not explicitly setting flags.

To summarize my usage, In my phases I'm doing the following to get test output to stdout:

  logging.basicConfig()
  test.logger.info(' foo')

And I'm currently not seeing logging go to stdout in plugs

grybmadsci:
Ah, you shouldn't have to call .basicConfig(), in fact that's probably overriding some of the things we set up internally (and might be why it's working for you)... but it won't hurt anything to do so.

openhtf/util/logs.py
```diff
@@ -37,14 +65,15 @@
...
-                   'Logging verbosity level.')
+                   'Console log verbosity level (stderr).')
+gflags.DEFINE_boolean('quiet', False, "If True, don't output logs to stderr.")
+gflags.DEFINE_string('log-file', '', 'Filename to output logs to, if any.')
+gflags.DEFINE_enum('log-file-level', 'warning', ['debug', 'info', 'warning',
+                                                 'error', 'critical'],
+                   'Logging verbosity level for log file output.')
 
 LOGGER_PREFIX = 'openhtf'
+RECORD_LOGGER = '.'.join(LOGGER_PREFIX, 'test_record')
```
fahhem: join takes a list (broke the tests)

fahhem:
LGTM after the tests are fixed

fahhem:
might want to pull back on calling yourself an idiot in the commits? Also, the tests are still broken due to the output_test detecting small changes in the framework

PR #152, opened 2016-03-04
Fix test result visibility in the existing barebones web frontend.

jettisonjoe:
* Add test status and DUT ID to the frontend test page.
  * Add a prompt-based test stop trigger so results can be kept on the screen until the operator acknowledges them.
  * Fix a bug in test_state.py where test outcomes were displayed as 'State.<outcome>'.
  * Fix a bug in phase_data.py where attachments were cleared to 'None' instead of to {}, which broke PhaseData's _asdict method in between test runs.


openhtf/io/frontend/templates/test.html
```diff
@@ -12,6 +12,17 @@
...
+  {% if state['test']['status'] == 'PASS' %}
+    <h1 style="color:green;">
+  {% elif state['test']['status'] == 'FAIL' %}
+    <h1 style="color:red;">
+  {% else %}
+    <h1 style="color:orange;">
+  {% end %}
+  {{state['test']['status']}}</h1>
+
+{% if state['framework']['status'] == 'EXECUTING' %}
```
fahhem: in the templates, it should be possible to do state.framework.status and it'll do the right thing automatically, no?

fahhem:
The output tests caught you changing the code :)

LGTM from my perspective

openhtf/io/frontend/templates/test.html
jettisonjoe: Downside of tornado templating versus flask. Their {{}} expressions are basically python, which means the data we're passing in is python dicts. This could be avoided by just moving the raw json around instead of parsing it, but since the current frontend server does some of the logic here we parse. The new frontend will have more logic built into the client, so the server will be able to just feed it json state updates from framework instances.

jettisonjoe:
Forgot to update those pesky output tests. Done now, and passing tests.

@fahhem, thanks for the fast feedback!

amyxchen:
The changes look good.

Issue #133, opened 2016-02-16
Shutting down is slow


fahhem:
It takes ~1 second for no good reason

fahhem:
I believe the causes for slow shutdown are the `join()` calls in the server `Stop()` methods. This is due to the way the main loop is told to exit:

https://github.com/google/openhtf/blob/262181734ceeb0649c38c28c4c0860ee0e32e46a/openhtf/io/frontend/__main__.py#L159
https://github.com/google/openhtf/blob/262181734ceeb0649c38c28c4c0860ee0e32e46a/openhtf/util/multicast.py#L88

The frontend sleeps for 5 seconds by default, and multicast sleeps for 3 seconds by default, so you can only exit on the next multiple of 5 seconds from when they were started.

fahhem:
For MulticastListener, there are two options, so I'd like to know which you prefer:

* Currently, it has a timeout of 3 seconds. This can be decreased to 1 without incurring much penalty, or even changed to a flag to allow someone to change it at will. The output unittest could specify 0.1 to allow it to exit quicker.
* Another way is to switch to using `select.select()`. This would only really help if we added a socketpair that we write to in `Stop()` to cause `select.select()` to return immediately. There is a [cross-platform way][1] to make a `socketpair()` that we can use.

The latter way would allow snappy shutdowns. Any other ideas? Suggestions?

[1]: http://code.activestate.com/recipes/525487-extending-socketsocketpair-to-work-on-windows/

grybmadsci:
I believe that select() on windows works with sockets (just not files), so
switching to select() should be fine, with no need to make custom wrappers.

grybmadsci:
Ah, i see the problem, i looked at it too quickly the first time... that's
a little ugly, the fallback to TCP loopback, but i suppose it's not that
bad, i've got no objection to adding a helper to do the socketpair business
and then using that to stop the multicast listener immediately.

Any thoughts, Joe?

~madsci

jettisonjoe:
I'm down with the socketpair approach, and it seems better than an adjustable timeout.

jettisonjoe:
[...]

PR #158, opened 2016-03-10
Remove a sleep in the frontend's StationStore

fahhem:
Part of #133, this should potentially reduce shutdown times.


jettisonjoe:
`_death_event` seems a little macabre compared to like `_stopped` or something, but thanks for cleaning this up.

PR #160, opened 2016-03-10
Allow reuse of phases with different options, etc

fahhem:
Changes WrapOrReturn to WrapOrCopy and only replaces attributes of TestPhaseOptions that aren't the default.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -61,8 +61,11 @@ class TestPhaseOptions(mutablerecords.Record(
   """
 
   def __call__(self, phase_func):
-    phase = TestPhaseInfo.WrapOrReturn(phase_func)
-    phase.options = self
+    phase = TestPhaseInfo.WrapOrCopy(phase_func)
+    for attr in self.__slots__:
+      value = getattr(self, attr)
+      if value is not None:
+        setattr(phase.options, attr, value)
     return phase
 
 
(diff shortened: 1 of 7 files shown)


openhtf/__init__.py
```diff
@@ -83,10 +87,22 @@ class TestPhaseInfo(mutablerecords.Record(
...
+    
+    Returns:
+      A new TestPhaseInfo object.
+    """
     if not isinstance(func, cls):
-      func = cls(func, inspect.getsource(func))
-    return func
+      return cls(func, inspect.getsource(func))
+    # We want to copy so that a phase can be reused
+    return copy.copy(func)
```
grybmadsci: Hm, do you need to deepcopy here?  Seems like you might get some weird behavior with a shallow copy, I suspect that means all copies would share single instances of options, plugs, and measurements, right?

fahhem: deepcopy is tricky because you can't say how many levels. I want to copy the _list_ of plugs, but I don't want to copy the actual plug classes as that can do some nasty stuff depending on the class. Though you're right, this only makes a new Record with the same stuff inside. Any suggestions other than 'deepcopy and we'll see when it fails' ?

grybmadsci: ugh, good call, it's not as simple as deepcopy... hm, since you just did the isinstance check, you know that func is an instance of TestPhaseInfo, so you could explicitly copy one level deep, which I *think* is enough?

fahhem: I was thinking of a function that would do 1-level-past-Record copying, so a record with a list would get a new list with the same elements. Since all the attributes are either Records, builtin collections, or immutable types, I think this will work (barring someone messing with framework internals incorrectly). 

I can add it to mutablerecords directly since it makes sense to be in that package, something like `mutablerecords.CopyRecord()`, what do you think?

grybmadsci: yea i guess there's no cleaner way to do it.. if it were me, i'd make it a method on the record base class, next to __copy__ and __deepcopy__, and call it something clever that i can't think of right now, but here's some ideas... RecordCopy (like deepcopy, but different), CopyRecord, CopyFields (borrowing some nomenclature from namedtuple).  entirely your call, mutablerecords is your thing.

fahhem: It's opensource, not 'my thing', that's not how opensource works. I'm simply the maintainer and a contributor.

I wasn't going to put it on the `Record` base class to avoid polluting the namespace.

fahhem:
Added `CopyRecord` and updated mutablerecords to 0.2.7 (and fixed tox's annoying dependency rigging)

fahhem:
Could I get this merged?

grybmadsci:
Please don't bump PRs less than 1 business day after the previous comment, it's just spam at that point.

PR #168, opened 2016-03-12
Test state updates

grybmadsci:
Update TestState to decouple running/completed from pass/fail.  This cascaded into lots of cleanup/refactoring.  Settled on this naming convention:
-  PhaseResult, any reference to a 'result': User-facing return values from test phases.
-  PhaseOutcome, any reference to an 'outcome': Framework internal state based on what happened in a phase, including return values, exceptions raised, or timeout.

Made a few other name changes, usually to be more explicit.  There's the one change from 'running_phase' to 'running_phase_record', because I was annoyed at having no idea what kind of object 'running_phase' was, and 'running_phase_record' at least gives a clue as to what to look for to find out (PhaseRecord).  That one required a change in the frontend, though, so @jettisonjoe a heads up for the new frontend.

Also, I moved things around a bit for examples, updated some docs for example_plug, made a real hello_world.py example (because our last one was really anything but).

I tried to keep any real logic changes inside test_state.py, as that's the focus of the PR, most other stuff should be primarily refactoring only.

Tested via all_the_things.py to produce a pass, fail, error (by raising), and timeout.  We should add an integration test that does those 4 things automatically, but i'm concerned this PR is already huge and noisy, so I'd like to do that in a followup (or have someone else do it ;))


openhtf/plugs/__init__.py
```diff
@@ -117,7 +118,7 @@ def TearDown(self):
     pass
 
 
-def requires(update_kwargs=True, **plugs):
+def plug(update_kwargs=True, **plugs):
```
fahhem: missed a test with this change

openhtf/exe/test_state.py
```diff
@@ -46,27 +46,16 @@ class TestState(object):
...
-  State = Enum('State', ['RUNNING', 'ERROR', 'TIMEOUT', 'ABORTED',
-                         'WAITING', 'FAIL', 'PASS', 'CREATED'])
-
-  _PHASE_RESULT_TO_STATE = {
-      phase_data.PhaseResults.CONTINUE: State.WAITING,
-      phase_data.PhaseResults.REPEAT: State.WAITING,
-      phase_data.PhaseResults.FAIL: State.FAIL,
-      phase_data.PhaseResults.TIMEOUT: State.TIMEOUT,
-  }
+  State = Enum('State', ['CREATED', 'RUNNING', 'COMPLETED'])
```
fahhem: Joe wanted ABORTED to eventually be triggered by the signal handlers

openhtf/exe/test_state.py
```diff
@@ -95,33 +82,36 @@ def _asdict(self):
...
-    else:
-      if phase_result.phase_result not in self._PHASE_RESULT_TO_STATE:
-        raise InvalidPhaseResultError(
-            'Phase result is invalid.', phase_result.phase_result)
-      self._state = self._PHASE_RESULT_TO_STATE[phase_result.phase_result]
-
-    return self._state in self._FINISHED_STATES
+    # Handle a few cases where the test is ending prematurely.
+    if phase_outcome.raised_exception:
+      self.record.outcome = 'ERROR'
```
fahhem: no enum for these values?

openhtf/exe/phase_executor.py
```diff
@@ -0,0 +1,208 @@
...
+
+  def __str__(self):
+    return '<%s: (%s)>' % (type(self).__name__, self.name)
+  __repr__ = __str__
+
+
+class PhaseExecutor(mutablerecords.Record(
+    'PhaseExecutor', ['test_state'], {'_current_phase_thread': None})):
+  """Encompasses the execution of the phases of a test."""
+  # TODO(madsci): continue removing mutablerecords here.
```
fahhem: you have `__init__` here, why does it still inherit from a `Record`?

openhtf/exe/test_state.py
fahhem: On second thought, this definitely needs an enum to at least list the values somewhere other than in the code here. Maybe `mfg_inspector` and `OutputToJson` are okay with passing the strings along, but someone may want to map these to their own names and an enum will give them a specific target to cover

grybmadsci: Hm, yeah, talked with Joe at the time and we decided strings were the best way to go for the output record, but it's a good point that there *really* should be a place to see at-a-glance what the possible values for 'outcome' are, and a valid point that someone else may want to use different strings.  I'll put the enum in test_record.py somewhere and switch this over.

openhtf/exe/test_state.py
grybmadsci: yeah we can add that back in when we plumb that through - ABORTED was unused here and I'd rather keep this as minimal as possible and only add things in when they are supported.

openhtf/exe/phase_executor.py
grybmadsci: Ah yeah, never finished up this TODO, will do that.

openhtf/plugs/__init__.py
grybmadsci: bah, yeah i figured i'd have missed a couple things, but wanted to get it sent out before i went to sleep, will go back and fix all these types of things today.

openhtf/exe/test_state.py
grybmadsci: Okay, after looking into this more, we were doing a poor job with enums in a couple places, both measurement outcomes and test outcome.  I've converted both to use Enum, and I think it looks a whole lot cleaner now, thanks for the suggestion :)  I've also updated the output mechanisms to stringify enums to their name, so the end result is we get the same string output, but internally they are proper Enum instances.

grybmadsci:
Waiting on an LGTM from @jettisonjoe to get this and #171 in, assuming no more comments from @fahhem ?

examples/hello_world.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Import this output mechanism as it's the specific one we want to use.
+from openhtf.io.output import json_factory
+
+# Import a handful of useful names.  If you're worried about polluting
+# your namespace, you can manually import just the things you want, this
+# is just a convenience.  See names.py for an exhaustive list.
+from openhtf.names import *
+
+# TODO(madsci): Clean up conf so we don't need to do this.
```
fahhem: What does 'clean up conf' mean? This todo has no meaning to anyone else

fahhem: Oh, it's for the `LoadFromDict()` below, whoops

examples/hello_world.py
```diff
@@ -0,0 +1,83 @@
...
+  # example of the sort of minimal functionality you're likely to use.
+
+  # The test.logger attribute is a Python logger instance that is configured
+  # to log to the test record we will output at the end of the test.  This
+  # is the recommended way to do any logging within test phases (this is also
+  # how to get logs to show up in the frontend).
+  test.logger.info('Hello World!')
+
+  # As described above, this is how measurements are set.  The value that has
+  # been set can also be accessed, but can only be set once.
```
fahhem: Actually, it can be set multiple times, it just warns on subsequent sets.

openhtf/__init__.py
```diff
@@ -72,37 +90,50 @@ def __call__(self, phase_func):
...
+      A new PhaseInfo object.
     """
     if not isinstance(func, cls):
       return cls(func, inspect.getsource(func))
     # We want to copy so that a phase can be reused with different options, etc.
     return mutablerecords.CopyRecord(func)
 
+  @property
+  def name(self):
+    return self.func.__name__
```
fahhem: Do we want the ability to override this, such as for lambdas and other badly-`__name__`d callables?

openhtf/exe/__init__.py
```diff
@@ -128,11 +127,11 @@ def _RunTestLoop(self):
       exit_stack.callback(lambda: setattr(self, '_current_exit_stack', None))
 
       plug_manager = self._MakePlugManager(suppressor)
-      phase_executor = self._MakePhaseExecutor()
+      executor = self._MakePhaseExecutor()
```
fahhem: It's weird to have a variable called `executor` passed around inside a `TestExecutor()` that's actually a `PhaseExecutor()`

openhtf/exe/phase_executor.py
```diff
@@ -0,0 +1,205 @@
...
+
+_LOG = logging.getLogger(__name__)
+
+
+class InvalidPhaseResultError(Exception):
+  """Raised when a PhaseOutcome is created with an invalid phase result."""
+
+
+class PhaseOutcome(collections.namedtuple(
+    'PhaseOutcome', 'phase_result')):
```
fahhem: same line?

openhtf/exe/phase_executor.py
```diff
@@ -0,0 +1,205 @@
...
+  a convenience to test for that condition, and the is_timeout attribute can
+  similarly be used to check for the timeout case.
+
+  The only accepted values for phase_result are None (timeout), an instance
+  of Exception (phase raised), or an instance of openhtf.PhaseResult.  Any
+  other value will raise an InvalidPhaseResultError.
+  """
+  def __init__(self, phase_result):
+    if not (phase_result is None or
+            isinstance(phase_result, Exception) or
```
fahhem: `isinstance(phase_result, (Exception, openhtf.PhaseResult))`

openhtf/util/__init__.py
```diff
@@ -56,8 +57,24 @@ def TimeMillis():  # pylint: disable=invalid-name
   return int(time.time() * 1000)
 
 
-def convert_to_dict(obj, ignore_keys=tuple()):
-  """Recursively convert namedtuples to dicts."""
+def ConvertToBaseTypes(obj, ignore_keys=tuple()):
+  """Recursively convert objects into base types, mostly dicts and strings.
```
fahhem: ...dicts and primitives

Otherwise, one might think this stringifies enums, ints and floats, which it doesn't

openhtf/util/measurements.py
```diff
@@ -93,16 +95,23 @@ class MultipleValidatorsException(Exception):
   """Multiple validators were used when defining a measurement."""
 
 
+class Outcome(Enum):
+  """Enum values for measurement 'outcome' fields."""
+  PASS = 'PASS'
```
fahhem: Why do you set the value to its string version? `value.name` returns the string version when you need it.

```py
Outcome = enum.Enum('Outcome', ['PASS', 'FAIL', 'UNSET'])

**Comment by fahhem** [comment]
LGTM from me

File: examples/hello_world.py (hunk 5 above)
Comment (grybmadsci) [inline examples/hello_world.py]: yeah this was already removed in #171 i believe, if not, i'll do that there.

File: examples/hello_world.py (hunk 6 above)
Comment (grybmadsci) [inline examples/hello_world.py]: Hm, whoops.  i think this will change when we validation happens at set time to avoid re-validation and weird frontend behavior (once-failing test becoming passing), but i'll update the comment to be accurate for now.

File: openhtf/__init__.py (hunk 7 above)
Comment (grybmadsci) [inline openhtf/__init__.py]: How would you specify it for a lambda?  I feel like fixing your callable to be better named via __name__ is probably the recommended fix for that?

Comment (grybmadsci) [inline openhtf/__init__.py]: Oh, you mean by passing a name explicitly as an option or somesuch?  Eh, I think for now let's keep it simple and add that later if a compelling case comes up for it.

File: openhtf/exe/__init__.py (hunk 8 above)
Comment (grybmadsci) [inline openhtf/exe/__init__.py]: i agree, this made me sad, but phase_executor is already taken by outer scope, and i wasn't feeling terribly creative... suggestions?

File: openhtf/exe/phase_executor.py (hunk 9 above)
Comment (grybmadsci) [inline openhtf/exe/phase_executor.py]: oh yeah, thanks.

File: openhtf/exe/phase_executor.py (hunk 10 above)
Comment (grybmadsci) [inline openhtf/exe/phase_executor.py]: oh neat, done.

File: openhtf/util/__init__.py (hunk 11 above)
Comment (grybmadsci) [inline openhtf/util/__init__.py]: And ints and floats, everything is `str()`d, so i think the name is okay as-is?

File: openhtf/util/measurements.py (hunk 12 above)
Comment (grybmadsci) [inline openhtf/util/measurements.py]: Ah yeah, pylint complains about bad-name that way, but it's more concise and cleaner, so i'll go with that, good point.

File: openhtf/io/test_record.py [hunk 13]
```diff
@@ -31,7 +33,15 @@ class InvalidMeasurementDimensions(Exception):
 LogRecord = collections.namedtuple(
     'LogRecord', 'level logger_name source lineno timestamp_millis message')
 OutcomeDetails = collections.namedtuple(
-    'OutcomeDetails', 'code_type code description')
+    'OutcomeDetails', 'code description')
+
+
+class Outcome(Enum):
```
grybmadsci: whoops, should fix this to not set values to names as well.  you missed one @fahhem :P

examples/hello_world.py
```diff
@@ -0,0 +1,84 @@
...
+  conf.LoadFromDict({})
+
+  # PromptForTestStart prompts the operator for a DUT ID, a unique identifier
+  # for the DUT (Device Under Test).  OpenHTF requires that a DUT ID is set
+  # each time a test is executed.  It may be set programmatically, but the
+  # simplest way to get one is to prompt the user for it.  If test_start is
+  # not provided, the test will start immediately and assume the DUT ID will
+  # be set later (OpenHTF will raise an exception when the test completes if
+  # a DUT ID has not been set).
+  test.Execute(test_start=triggers.PromptForTestStart())
```
jettisonjoe: This is awesome. _love_ the new `hello_world.py`!

openhtf/exe/__init__.py
jettisonjoe: How about `flagship_of_lord_vader`?

openhtf/exe/test_state.py
jettisonjoe: I was sitting next to @grybmadsci when he typed this change, and I'm cool with it. I'm dragging my heals on wiring up `ABORTED`, so I should be responsible for adding it back in later.

jettisonjoe:
LGTM

Issue #169, opened 2016-03-13
mfg_inspector TypeError,   UploadTestRun() takes at most 3 arguments (4 given)


wallacbe:
Trying to upload from hello_world.py like this:
`user = '<my user>'`
`  key = '<my_key>'`
`  with open(key) as f:`
`    key_data = f.read()`
`  test.AddOutputCallback(mfg_inspector.UploadToMfgInspector(user, key_data))`
    
self.test.OutputTestRecord(self._test_state.GetFinishedRecord())
  File "/home/source/venv/local/lib/python2.7/site-packages/openhtf-0.9-py2.7.egg/openhtf/__init__.py", line 167, in OutputTestRecord
    output_cb(test_record)
  File "/home/source/venv/local/lib/python2.7/site-packages/openhtf-0.9-py2.7.egg/openhtf/io/output/mfg_inspector.py", line 386, in __call__
    self.UploadTestRun(testrun, self.DESTINATION_URL, credentials)
TypeError: UploadTestRun() takes at most 3 arguments (4 given)

wallacbe:
pulled repo from head

wallacbe:
Worked when I added 'self'
def UploadTestRun(self, testrun, destination, credentials=None):

PR #170, opened 2016-03-13
add self to UploadTestRun method definition

wallacbe:
https://github.com/google/openhtf/issues/169


openhtf/io/output/mfg_inspector.py
```diff
@@ -348,7 +348,7 @@ def from_json(cls, json_data):
                keydata=json_data['private_key'],
                token_uri=json_data['token_uri'])
 
-  def UploadTestRun(testrun, destination, credentials=None):
+  def UploadTestRun(self, testrun, destination, credentials=None):
```
grybmadsci: ha, this is a silly bug, sorry about that!  Since self isn't actually used anywhere, can you decorate this with `@staticmethod` instead of adding the self argument?  Thanks!

PR #171, opened 2016-03-14
Config updates!

grybmadsci:
See the lengthy module docstring for details, but a summary of the changes:
-  Remove various deprecated things, `StackedConfig`, value validation.
-  Change flag names to `--config-value` and `--config-file` to be more GNU-standard. @wallacbe heads up on this one, it's annoying but I'd like to do it once now and get it out of the way.
-  Simplified config value access, directly on `conf` module as attribute/dict access (`conf.foo` or `conf['foo']`), 'in' checks work directly on `conf` module in a sensible way (if a value is available, `key in conf` is True).
-  Remove `config` attribute on phase data object passed to test phases, users should directly use the `openhtf.conf` module.
-  Kept `.Declare()` semantics the same, slight refactor of the `Declaration` record type.
-  Simplified and clarified `Load()` semantics:
  -  `Load()` not required anymore, will fall back to defaults/flags values if available.
  -  Multiple `Load*()` methods for convenience, each one documented.
  -  All `Load*()` methods take optional `_override` to allow explicit control of overriding (no more `LoadMissingFromDict()`, what does that even mean?!)
  -  Added support for loading from JSON as well as YAML (just tries both, user doesn't have to specify which to use).

Unit test coming, but I wanted to get this out there for review now.  I also recognize that this is going to be a slightly painful change for tests that already exist, but it's only going to get worse the longer we put it off, and I really want to change how we interface with the config.  The `conf.Config()` being instantiated some places, passed around other places, confusing load semantics, are all pretty painful.


grybmadsci:
Oh, this is currently rebased on top of another PR, for now only look at conf.py, the diff will clean up when that other PR goes in.

openhtf/conf.py
```diff
@@ -43,250 +128,224 @@
...
+    self._functools = _functools
+    self._modules = kwargs
+    self._declarations = {}
+    self._loaded_values = {}
+    self._flag_values = {}
+    for keyval in self._flags['config-value'].value:
+      self._flag_values.setdefault(*keyval.split('=', 1))
+
+    # Everywhere that uses configuration uses this, so we just declare it here.
+    self.Declare('station_id', 'The name of this tester')
```
fahhem: This is required in places and yet its declaration is buried. I'd declare it in `openhtf/__init__.py` if possible and check if it exists earlier

openhtf/conf.py
```diff
@@ -43,250 +128,224 @@
...
-    self._state = state if state is not None else {}
-    self._declarations = declarations or _DeclaredKeys()
-    self._loaded = state is not None
-    self.lock = threading.Lock()
+    if item not in self._declarations:
+      raise self.UndeclaredKeyError('Configuration key not declared', item)
+
+    if item in self._flag_values:
+      if item in self._loaded_values:
+        self._logger.warning(
```
fahhem: So this will log on every access? What if we had a single dict that had all the 'final' values and warned for all the replaced keys when a Load function is called?

openhtf/conf.py
```diff
@@ -294,334 +353,110 @@ def LoadFromDict(self, dictionary, force_reload=False):
...
-
-    for key, value in config_dict.items():
-      if key in self._state:
+    for key, value in dictionary.iteritems():
+      # Warn in this case.  We raise if you try to access a config key that
+      # hasn't been declared, but we don't raise here so that you can use
+      # configuration files that are supersets of required configuration for
+      # any particular test station.
+      if key not in self._declarations:
+        self._logger.warning('Ignoring undeclared configuration key: %s', key)
```
fahhem: If you want to make superset-configs okay, I'd drop the spamminess of this by storing ignored keys in a list and printing them all out once afterward

grybmadsci: Ah, good idea, will do in the next push.

openhtf/conf.py
```diff
@@ -294,334 +353,110 @@ def LoadFromDict(self, dictionary, force_reload=False):
...
+                               'is a configuration key', kwarg)
+
+      # Set positional args from configuration values.
+      config_args = {name: self[name] for name in arg_names if name in self}
+
+      for overridden in set(kwargs) & set(config_args):
+        self._logger.warning('Overriding configuration value for kwarg %s (%s) '
+                             'with provided kwarg value: %s', overridden,
+                             self[overridden], kwargs[overridden])
+        del config_args[overridden]
```
fahhem: if you flipped the update method, you could elide this del, and it would make the operation clearer (configuration keys, except for what gets passed in)

openhtf/conf.py
```diff
@@ -294,334 +353,110 @@ def LoadFromDict(self, dictionary, force_reload=False):
...
+      if key in retval:
+        retval[key] = value
+    return retval
+
+  def InjectPositionalArgs(self, method):
+    """Decorator for injecting positional arguments from the configuration.
+
+    This decorator wraps the given method, so that any positional arguments are
+    passed with corresponding values from the configuration.  The name of the
+    positional argument must match the configuration key.  Keyword arguments
```
fahhem: Keyword arguments in the signature or when invoked? Why would keyword arguments that aren't being modified need to match config keys? That feels wrong to me. Also, the implementation seems to do the opposite, where it only warns when you have a kwarg passed in that _is_ in the config.

openhtf/conf.py
grybmadsci: Hm, I think I see what you're getting at.  My reasoning here was that I really do want to discourage using commandline flags to override config parameters.  It really should be used sparingly for debug purposes only, whereas the intention is for multiple Load() calls to be a perfectly reasonable thing to do.  Note that there is an `info()` message logged whenever a loaded value is overridden by a new loaded value, so I think that covers your latter point?

openhtf/conf.py
grybmadsci: good call, will switch that around.

openhtf/conf.py
fahhem: I meant that the `Load()` calls would each result in a final dict (which would simplify `__contains__` and `__getitem__`) and only warn, or log `info()`, once each with a list of the keys being overridden. If you want flags to spam, then never mind. Having a final dict may still be useful, but keeping the flag values around explicitly would allow you to treat them differently/as second-class values :)

grybmadsci: Ah, I see what you meant.  Yeah I like keeping them around separately because it does allow me to treat flag-provided values as second class citizens and badger the user for using them, and it also lets me distinguish between overriding a `default_value` (which shouldn't log a message, that's an obvious thing to do) and overriding a previously `Load()`ed value.

A snapshot view of the config is available via `_asdict()`, incidentally, but I'm just not using it to detect deltas like you're talking about.

openhtf/conf.py
grybmadsci: Hm, I think you read that backwards?  "Keyword arguments are not modified, but should not be named such that..."
As in, we don't touch keyword args no matter what, but don't name them overlapping names with config keys because that's confusing.  The warning is to call out exactly that bad practice (keyword arg named something that's in the config).  Is there something else here that I missed?

openhtf/conf.py
grybmadsci: ah, good call, i think this bounced around a few times in the XTF -> OpenHTF migration, but now it clearly belongs in `openhtf/__init__.py`.  moved.

grybmadsci: while we're looking at this, it's a little irksome that we have this single required configuration key.  if it's always required, then why isn't it in the constructor of the `Test()` object?  Should we really require this?  I believe back when Joe and I talked about it initially, we had said that this was fundamentally required for a test record to make sense (it had to run on SOME machine, after all), but I think we had talked about defaulting it to the machine's hostname or something.  Thoughts on that?  Declaring it in __init__, and in `Execute()`, after loading the conf file, if there isn't a `station_id` set, grab the hostname?  That way, if the user doesn't provide one, a sane default is used?

fahhem: That's actually what I do in some cases. I shove the hostname in for tests that aren't designated a particular station. I think taking it as an actual argument is one possibility, but it's in line with the philosophy of configs being station-specific. As an argument makes getting it from configs difficult, so I'd say it could be both: A default config key specified in `__init__.py` as well as taken by `Test.__init__()` (as `conf.InjectPositionalArgs`? That might reverse the override semantics?).

grybmadsci: The current problem with using `conf.InjectPositionalArgs` is that the conf file hasn't been loaded yet at `Test()` instantiation (it happens in `Test.Execute()` currently).  I think the override semantics would be consistent (pull from conf, unless specified in the call).

This brings up another point - loading from config file specified via flags should probably happen at `import` time, so I'll make that change.  Then we could take it as an arg to the `Test()` constructor, but I rather like the semantics of that only taking test phases and metadata keys.

If we have a fallback of hostname, then I'm okay with not providing a means to specify it in the `Test()` constructor, because it's not really required from a user standpoint (the user doesn't have to specify it in a config somewhere to get OpenHTF to work).  Toughts?  Move `--config-file` loading to `import conf` time (basically immediately before loading `--config-value` flags), and default `station_id` to machine hostname if it's not set by the time `Test.Execute()` is called?

fahhem: That seems good, why didn't we load configs at import time in the first place? We had so many problems due to that before...

Yeah, so config values are available at:

- import time due to loading a file
- `conf.Load*()`
- or lastly when checked during the test's execution.
  * `station_id` specifically gets checked in `Test.Execute()` against a default of `socket.gethostname()`

grybmadsci: Hm, I can't recall why we weren't doing the load at import time before, but I suspect it was something to do with the way the models and validators were set up?  Not sure, but as long as it works not, I like this better :)  Also let me remove stashing the flags on the `Configuration()` object and clean up `LoadFromFile()` a bit.

fahhem: I thought about it, and I think it was because we wanted to be able to validate the keys at load time and make sure both that no extra keys were defined and that all values were valid. This can be done at another time now.

I'm thinking this allows adding a layer of validation on top that will include a richer declaration API and a `LoadAndValidate()` call that will check/convert everything but the actual value usage will be the same (`conf.key`). That layer would be separate from OpenHTF (or maybe in `contrib` if we're still interested in that?) and backwards-compatible.

grybmadsci: Oh, it's also because at import time, flags typically haven't been parsed out of sys.argv yet.  Solutions to this are:
- Parse `gflags.FLAGS` from `sys.argv` upon `import` of `conf` module.
- Parse only the `--config-file` flag from `sys.argv` if it's present upon `import` of `conf` module.
- Trigger `LoadFromFile(FLAGS['config-file'].value)` from `Execute()` instead, similar to how it was before.
Thoughts?  I kind of like the second one, I think it's weird to parse all commandline args just because you `import`ed `conf`, and I like having `conf` be standalone enough that you don't have to explicitly invoke `LoadFromFile()` with the `--config-file`.  It's a tiny bit ugly because it'll mean doing a little string parsing that duplicates what `gflags` is going to do, but I think that's okay.

grybmadsci:
Started writing unit test, played around interactively with this a bit and came across some issues, so I've fixed those, still working on that unit test, though, but now I at least believe everything works as intended.

grybmadsci:
99% coverage \o/ the last two lines are tough to test because they rely on different `sys.argv` values at `import` time, and `pytest` only starts up one Python instance, so there's only one real "import time".  I updated `conf.Reset()` to do config file reloading, in case someone's doing something weird with dynamic updates to config files, but i don't re-parse `--config-value` flags in `conf.Reset()` for obvious reasons (they shouldn't have changed).  Rather not change that just for 2 lines of unittest coverage, so... it's 99%, deal with it.

openhtf/conf.py
jettisonjoe: I had to read it twice before I caught the critical `not`. I think it's the `such that` that leads your brain to skip the `not`.

jettisonjoe:
Looks great! Love what you've done with the place. I recommend trying to tighten up some of the wording in that `InjectPositionArgs` docstring to make the guidelines around kwargs more clear, but other than that LGTM.

PR #172, opened 2016-03-15
Print console

amyxchen:
a output module to print test result on the console.

### DIFF
--- a/openhtf/io/output/__init__.py
+++ b/openhtf/io/output/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2014 Google Inc. All Rights Reserved.
+# Copyright 2016 Google Inc. All Rights Reserved.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,15 +12,92 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Openhtf output modules."""
+"""This module contains support for various built-in output mechanisms.
 
-import openhtf.io.output.json_factory
-import openhtf.io.output.mfg_inspector
+Here, a base OutputToFile class is implemented to provide simple output to
+a file via the pickle serialization mechanism. It can be subclassed to implement
+alternative serialization schemes, see json_factory.py and mfg_inspector.py for
+examples.
+"""
 
-# pylint: disable=invalid-name
+import base64
+import contextlib
+import cPickle as pickle
+import os
+import tempfile
+from openhtf.util import data
 
-# Classes used for output
-OutputToJSON = json_factory.OutputToJSON
-UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
-OutputToTestRunProto = mfg_inspector.OutputToTestRunProto
-UploadOrOutput = mfg_inspector.UploadOrOutput
+
+# TODO(wallacbe): Switch to util
+class Atomic(object):
+  """Class that does atomic write in a contextual manner."""
+
+  def __init__(self, filename):
+    self.filename = filename
+    self.temp = tempfile.NamedTemporaryFile(delete=False)
+
+  def write(self, write_data):
+    return self.temp.write(write_data)
+
+  def close(self):
+    self.temp.close()
+    os.rename(self.temp.name, self.filename)
+
+
+class OutputToFile(object):
+  """Output the given TestRecord to a file.
+
+  Instances of this class are intended to be used as an output callback
+  (see Test.AddOutputCallbacks) to output TestRecord results to a file.
+  This base implementation outputs the TestRecord by serializing it via
+  the pickle module.  Subclasses may change this by overriding the
+  serialize_test_record() method.  Additionally, subclasses may implement
+  more complex file naming mechanisms by overriding the open_file() method.
+
+  Args:
+    test_record: The TestRecord to write out to a file.
+  """
+
+  def __init__(self, filename_pattern):
+    self.filename_pattern = filename_pattern
+
+  @staticmethod
+  def serialize_test_record(test_record):
+    """Override method to alter how test records are serialized to file data."""
+    return pickle.dumps(test_record, -1)
+
+  @staticmethod
+  def open_file(filename):
+    """Override method to alter file open behavior or file types."""
+    return Atomic(filename)
+
+  @contextlib.contextmanager
+  def open_output_file(self, test_record):
+    """Open file based on pattern."""
+    output_file = None
+    record_dict = data.ConvertToBaseTypes(test_record)
+    if isinstance(self.filename_pattern, basestring):
+      if '{' in self.filename_pattern:
+        output_file = self.open_file(
+            self.filename_pattern.format(**record_dict))
+      elif '%' in self.filename_pattern:
+        raise ValueError(
+            '%-style filename patterns deprecated, use .format() syntax')
+      else:
+        output_file = self.open_file(self.filename_pattern)
+    elif callable(self.filename_pattern):
+      output_file = self.open_file(self.filename_pattern(record_dict))
+    if output_file:
+      try:
+        yield output_file
+      finally:
+        output_file.close()
+    elif hasattr(self.filename_pattern, 'write'):
+      yield self.filename_pattern
+    else:
+      raise ValueError(
+          'filename_pattern must be string, callable, or File-like object')
+
+  def __call__(self, test_record):
+    with self.open_output_file(test_record) as outfile:
+      outfile.write(self.serialize_test_record(test_record))
--- a/openhtf/io/output/console_summary.py
+++ b/openhtf/io/output/console_summary.py
@@ -1,0 +1,77 @@
+"""Module to display test summary on console."""
+
+import os
+import sys
+import __main__ as tester
+
+from openhtf import util
+from openhtf.io import test_record
+from openhtf.util import measurements as measurement
+
+
+class ConsoleSummary():
+  """Print test results with failure info on console. """
+
+  # pylint: disable=invalid-name
+  def __init__(self, indent=2, output_stream=sys.stdout):
+    self.indent = ' ' * indent
+    if os.name == 'posix':    #Linux and Mac
+      self.RED = '\033[91m'
+      self.GREEN = '\033[92m'
+      self.ORANGE = '\033[93m'
+      self.RESET = '\033[0m'
+      self.BOLD = '\033[1m'
+    else:
+      self.RED = ""
+      self.GREEN = ""
+      self.ORANGE = ""
+      self.RESET = ""
+      self.BOLD = ""
+    
+    self.color_table = {
+        test_record.Outcome.PASS:self.GREEN, 
+        test_record.Outcome.FAIL:self.RED, 
+        test_record.Outcome.ERROR:self.ORANGE, 
+        test_record.Outcome.TIMEOUT:self.ORANGE, 
+    }
+    self.output_stream = output_stream
+  # pylint: enable=invalid-name
+
+  def __call__(self, record):
+    output_lines = [''.join((self.color_table[record.outcome],
+                   self.BOLD, record.code_info.name, ':',
+                   record.outcome.name, self.RESET))]
+    if record.outcome != test_record.Outcome.PASS:
+      for phase in record.phases:
+        new_phase = True
+        phase_time_sec = (float(phase.end_time_millis)
+                          - float(phase.start_time_millis))/1000.0
+        measured = phase.measured_values
+        measurements = phase.measurements
+        for key in measurements:
+          result = measurements[key]
+          if result.outcome == measurement.Outcome.FAIL:
+            if new_phase:
+              output_lines.append('failed phase: %s [ran for %.2f sec]' %
+                                  (phase.name, phase_time_sec))
+              new_phase = False
+
+            output_lines.append('%sfailed_item: %s' % (self.indent, result.name))
+            output_lines.append('%smeasured_value: %s' % 
+                                  (self.indent*2, str(measured[result.name])))
+            output_lines.append('%svalidators:' % (self.indent*2))
+            for validator in result.validators:
+              output_lines.append('%svalidator: %s' % 
+                                  (self.indent*3, str(result.validators)))
+        
+        phase_result = phase.result.phase_result
+        if not phase_result:   #Timeout 
+           output_lines.append('timeout phase: %s [ran for %.2f sec]' %
+                              (phase.name, phase_time_sec))
+        elif 'CONTINUE' not in str(phase_result):  #Exception
+          output_lines.append('%sexception type: %s' % (self.indent, 
+                              record.outcome_details[0].code))
+ 
+    output_lines.append('\n')
+    text = '\n'.join(output_lines)
+    self.output_stream.write(text)
--- a/examples/all_the_things.py
+++ b/examples/all_the_things.py
@@ -27,6 +27,7 @@ import openhtf
 import openhtf.io.output as output
 
 from openhtf.io.output import json_factory
+from openhtf.io.output import mfg_inspector 
 from openhtf.names import *
 # Uncomment for mfg-inspector output, requires setup.py build_proto.
 #from openhtf.io.output import mfg_inspector
@@ -103,8 +104,8 @@ if __name__ == '__main__':
       json_factory.OutputToJSON(
           './{dut_id}.{metadata[test_name]}.{start_time_millis}.json',
           indent=4))
-  test.AddOutputCallbacks(output.OutputToTestRunProto(
-      './%(dut_id)s.%(start_time_millis)s.pb'))
+  test.AddOutputCallbacks(mfg_inspector.OutputToTestRunProto(
+      './{dut_id}.{start_time_millis}.pb'))
   # Example of how to upload to mfg-inspector.  Replace filename with your
   # JSON-formatted private key downloaded from Google Developers Console
   # when you created the Service Account you intend to use, or name it
--- a/examples/hello_world.py
+++ b/examples/hello_world.py
@@ -75,7 +75,7 @@ if __name__ == '__main__':
   # This will always output to the same ./hello_world.json file, formatted
   # slightly for human readability.
   test.AddOutputCallbacks(
-      json_factory.OutputToJSON('./hello_world.json', indent=2))
+      json_factory.OutputToJSON('./{dut_id}.hello_world.json', indent=2))
 
   # PromptForTestStart prompts the operator for a DUT ID, a unique identifier
   # for the DUT (Device Under Test).  OpenHTF requires that a DUT ID is set
--- a/openhtf/io/output/json_factory.py
+++ b/openhtf/io/output/json_factory.py
@@ -2,18 +2,15 @@
 
 import base64
 from json import JSONEncoder
-
 from openhtf.exe import test_state
+from openhtf.io import output
 from openhtf.util import data
 
-
-class OutputToJSON(JSONEncoder):
+class OutputToJSON(output.OutputToFile):
   """Return an output callback that writes JSON Test Records.
-
   Example filename_patterns might be:
     '/data/test_records/{dut_id}.{metadata[test_name]}.json', indent=4)) or
     '/data/test_records/%(dut_id)s.%(start_time_millis)s'
-
   To use this output mechanism:
     test = openhtf.Test(PhaseOne, PhaseTwo)
     test.AddOutputCallback(openhtf.OutputToJson(
@@ -30,33 +27,19 @@ class OutputToJSON(JSONEncoder):
   """
 
   def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
-    super(OutputToJSON, self).__init__(**kwargs)
-    self.filename_pattern = filename_pattern
+    super(OutputToJSON, self).__init__(filename_pattern)
     self.inline_attachments = inline_attachments
-
-  def default(self, obj):
-    if isinstance(obj, BaseException):
-      # Just repr exceptions.
-      return repr(obj)
-    return super(OutputToJSON, self).default(obj)
-
-  # pylint: disable=invalid-name
-  def __call__(self, test_record):
-    assert self.filename_pattern, 'filename_pattern required'
+    self.json_encoder = JSONEncoder(**kwargs)
+ 
+  def serialize_test_record(self, test_record):
+    return self.json_encoder.encode(self.convert_to_dict(test_record))
+  
+  def convert_to_dict(self, test_record):
     if self.inline_attachments:
       as_dict = data.ConvertToBaseTypes(test_record)
       for phase in as_dict['phases']:
         for value in phase['attachments'].itervalues():
           value['data'] = base64.standard_b64encode(value['data'])
     else:
-      as_dict = data.ConvertToBaseTypes(test_record, ignore_keys='attachments')
-    if isinstance(self.filename_pattern, basestring):
-      if '{' in self.filename_pattern: 
-        filename = self.filename_pattern.format(**as_dict)
-      else:
-        filename = self.filename_pattern % as_dict
-      with open(filename, 'w') as f:
-        f.write(self.encode(as_dict))
-    else:
-      self.filename_pattern.write(self.encode(as_dict))
-  # pylint: enable=invalid-name
+      as_dict = data.ConvertToBaseTypes(test_record, ignore_keys=('attachments',))
+    return as_dict
--- a/openhtf/io/output/mfg_inspector.py
+++ b/openhtf/io/output/mfg_inspector.py
@@ -151,6 +151,8 @@ def _ExtractAttachments(phase, testrun, used_parameter_names):
     name = _EnsureUniqueParameterName(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
+    if isinstance(data, unicode):
+      data = data.encode('utf8')
     testrun_param.value_binary = data
     if mimetype in MIMETYPE_MAP:
       testrun_param.type = MIMETYPE_MAP[mimetype]
--- a/openhtf/io/output/mfg_inspector.py
+++ b/openhtf/io/output/mfg_inspector.py
@@ -37,6 +37,7 @@ import sys
 import threading
 import zlib
 
+from openhtf.io import output
 from openhtf.io.output import json_factory
 from openhtf.io.proto import guzzle_pb2
 from openhtf.io.proto import test_runs_pb2
@@ -135,8 +136,8 @@ def _AttachJson(record, testrun):
   copied over and can potentially be quite large.
   """
   record_dict = data.ConvertToBaseTypes(record, ignore_keys=('attachments',))
-  record_json = json_factory.OutputToJSON(
-      sort_keys=True, indent=2).encode(record_dict)
+  record_json = json_factory.OutputToJSON(inline_attachments=False,
+      sort_keys=True, indent=2).serialize_test_record(record)
   testrun_param = testrun.info_parameters.add()
   testrun_param.name = 'OpenHTF_record.json'
   testrun_param.value_binary = record_json
@@ -336,7 +337,7 @@ def _TestRunFromTestRecord(record):
   return testrun
 
 
-class OutputToTestRunProto(object):  # pylint: disable=too-few-public-methods
+class OutputToTestRunProto(output.OutputToFile):  # pylint: disable=too-few-public-methods
   """Return an output callback that writes mfg-inspector TestRun Protos.
 
   Example filename_patterns might be:
@@ -358,22 +359,11 @@ class OutputToTestRunProto(object):  # pylint: disable=too-few-public-methods
   """
 
   def __init__(self, filename_pattern):
-    self.filename_pattern = filename_pattern
+    super(OutputToTestRunProto, self).__init__(filename_pattern)
 
-  def __call__(self, test_record):  # pylint: disable=invalid-name
-    as_dict = data.ConvertToBaseTypes(test_record)
-    serialized = _TestRunFromTestRecord(test_record).SerializeToString()
-    if isinstance(self.filename_pattern, basestring):
-      if '{' in self.filename_pattern:
-        filename = self.filename_pattern.format(**as_dict)
-      else:
-        filename = self.filename_pattern % as_dict
-      with open(filename, 'w') as outfile:
-        outfile.write(serialized)
-    else:
-      filename = self.filename_pattern
-      self.filename_pattern.write(serialized)
-    return filename
+  @staticmethod
+  def serialize_test_record(test_record):
+    return _TestRunFromTestRecord(test_record).SerializeToString()
 
 
 class UploadToMfgInspector(object):  # pylint: disable=too-few-public-methods


openhtf/plugs/cambrionix/__init__.py
```diff
@@ -13,8 +13,11 @@
 # limitations under the License.
 
 import subprocess
+from openhtf import conf
```
fahhem: separate PR?

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+  Args:
+    filename_pattern: A format string specifying the filename to write to,
+      will be formatted with the Test Record as a dictionary.
+    inline_attachments: Whether attachments should be included inline in the
+      output.  Set to False if you expect to have large binary attachments.
+  """
+
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
+    super(PrintConsole, self).__init__(**kwargs)
+    self.FAIL =  '\033[91m'
```
fahhem: I was about to work on this so I have some of the code written. Here's what I've done for coloring, you can decide which you think is better. You've already written it and it has bold output, so yours may be better.

```py
from tornado import log as tornado_log

for handler in logging.getLogger('openhtf').handlers:
  if isinstance(handler, logging.StreamHandler):
    handler.setFormatter(tornado_log.LogFormatter())

# That puts the formatter on 'openhtf', but we may want it just on 'openhtf.output.console'

LOG = logging.getLogger('openhtf.output.console')
LOG.propagate = False
handler = logging.StreamHandler()
handler.setFormatter(tornado_log.LogFormatter())
LOG.addHandler(handler)
```

Then in `__call__` you can do this:

```py
if result == 'PASS':
  func = LOG.info  # green
elif result == 'FAIL':
  func = LOG.error  # red
else:
  func = LOG.warning  # yellow

func('This is what you print')
```

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+    self.PASS =  '\033[92m'
```
fahhem: Either way, please comment as to what color these are

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+    self.ERROR = '\033[93m'
+    self.ENDC =  '\033[0m'
+    self.BOLD =  '\033[1m'
+
+  # pylint: disable=invalid-name
+  def __call__(self, test_record):
+    as_dict = util.convert_to_dict(test_record)
```
fahhem: Why do you convert it to dict? You can use it directly like so:

```py
result = test_record.outcome
```

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+    result = as_dict['outcome']
+    color = None
+    if result == 'PASS':
+      color = self.PASS
+    elif result == 'FAIL':
+      color = self.FAIL
+    else:
+      color = self.ERROR
+
+    print '\n'+color+self.BOLD+os.path.basename(sys.argv[0])+": "+result+self.ENDC
```
fahhem: use %s:

```py
bold_color = color + self.BOLD
print '\n%(color)s%(filename)s: %(result)s%(end_color)s' % {
  'color': color,
  'filename': os.path.basename(sys.argv[0]),
  'result': result,
  'end_color': self.ENDC
}
```

openhtf/io/output/print_console.py
grybmadsci: Agreed, I'd actually prefer to see these named by color instead of function, and see the function->color mapping explicitly elsewhere.

Also, it's better to define all-caps constants like this on the class rather than initializing them in __init__(), so you'll have something like:
```
class PrintConsole(object):
  GREEN = '\033[92m'
  RED = '\033[93m'
  ...
```

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
+"""Module for outputting test record to JSON-formatted files."""
+import os
+import sys
+from json import JSONEncoder
+from openhtf import util
+
+
+class PrintConsole(JSONEncoder):
```
grybmadsci: hm, i don't think you have to inherit from JSONEncoder, it doesn't look like you are using it anywhere?

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
+"""Module for outputting test record to JSON-formatted files."""
```
grybmadsci: Update docstring.

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+      for phase in phases:
+        phase_name = phase['name']
+        phase_time = str((float(phase['end_time_millis']) - \
+                               float(phase['start_time_millis']))/1000)
+        if result == 'FAIL':
+          measured = phase['measured_values']
+          measurements = phase['measurements']
+          for measurement in measurements:
+            res = measurements[measurement]
+            if res['outcome'] == 'FAIL':
```
grybmadsci: better to not convert the test_record to a dict and operate on the records directly, so here you'd have:
```
from openhtf.util import measurements

for measurement in phase.measurements:
  if res.outcome == measurements.Outcome.FAIL:
    ...
```

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,81 @@
...
+    inline_attachments: Whether attachments should be included inline in the
+      output.  Set to False if you expect to have large binary attachments.
+  """
+
+  def __init__(self, filename_pattern=None, inline_attachments=True, **kwargs):
+    super(PrintConsole, self).__init__(**kwargs)
+    self.FAIL =  '\033[91m'
+    self.PASS =  '\033[92m'
+    self.ERROR = '\033[93m'
+    self.ENDC =  '\033[0m'
```
grybmadsci: i would call this 'RESET' or 'NORMAL' instead of 'ENDC', to be more consistent with ANSI SGR nomenclature.  Also maybe link to https://en.wikipedia.org/wiki/ANSI_escape_code in a comment to help out any poor souls trying to figure out how this works.

openhtf/io/output/print_console.py
grybmadsci: once #168 goes in, this will be an Enum that would be better to use directly.  In fact, you could build a map from enum value to color:
```
from openhtf.io import test_record

class PrintConsole(object):
  GREEN = '\033[92m'
  OUTCOME_COLOR_MAP = {
    test_record.Outcome.PASS: GREEN,
    ...
  }
  ...
  def __call__(self, test_record):
    ...
    print '\n' + self._colorize(
        self.OUTCOME_COLOR_MAP[test_record.outcome] + self.BOLD,
        os.path.basename(sys.argv[0]) + ': ' + test_record.outcome.name
    )
    ...

File: openhtf/io/output/print_console.py (hunk 5 above)
Comment (grybmadsci) [inline openhtf/io/output/print_console.py]: I would recommend making a helper function on this class that applies the color and appends the reset automatically:
```
def _colorize(self, color, string):
  return ''.join((color, string, self.NORMAL))
```
As a side note, generally prefer ''.join() over +'ing strings together - it's a Python thing, and it's more efficient for reasons.

File: openhtf/io/output/print_console.py [hunk 10]
```diff
@@ -0,0 +1,81 @@
...
+              else:
+                print ' '*8 + 'validator: None'
+        else:
+          phase_result = phase['result']
+          if phase_result['raised_exception'] == 'True':
+            print 'raised_exception phase: ' + phase_name +' [time taken(s): '\
+                              + phase_time + ']' + self.ENDC
+            exp_type = as_dict['outcome_details'][0]['code']
+            print ' '*4 + 'exception type : ' + exp_type + self.ENDC
+            print ' '*4 + 'exception text: ' + phase_result['phase_result'] + self.ENDC
```
grybmadsci: rather than have the print statements inline, it would be better to start with a list of `output_lines = []`, and then `output_lines.append(...)` each place you would print, and then at the end do `sys.stdout.write(''.join(output_lines))`

jettisonjoe: What's wrong with print?

grybmadsci: The typical rule of thumb is to only use print for interactive scripts (such as a REPL using `input()` and `print`).  `print` behaves slightly differently in some edge cases, and IMO is less clear about the exact output (ie, commas magically turn into space-separation, implicit newline at the end).

Also, print is less forward-compatible with Python3, and doesn't allow for easily outputting to something else (whereas it would be easy to parameterize `sys.stdout` and write to a file instead).

As a side note, I wasn't feeling super clever at 6 am, but now that I look at it, you might also benefit from a small helper class that supports indentation syntax, perhaps like:
```
class IndentingFile(object):
  def __init__(self, output_file, indent=2):
    self._output_file = output_file
    self._current_indent = 0
    self._indent = indent

  def write(self, data):
    self.output_file.write(''.join((self._current_indent * ' ', data, '\n')))

  @contextlib.contextmanager
  def indent(self):
    self._current_indent += self._indent
    yield self
    self._current_indent -= self._indent
```
Then you can:
```
def foo(test_record):
  indenter = IndentingFile(sys.stdout)
  indenter.write('first line')
  with indenter:
    indenter.write('indented one level')
    with indenter:
      indenter.write('indented two levels')
```
That may be more or less readable, just throwing the idea out there.

openhtf/plugs/cambrionix/__init__.py
amyxchen: Yes, separate PR.

openhtf/io/output/print_console.py
amyxchen: Not using JSONEncoder, removed

openhtf/io/output/print_console.py
amyxchen: Done

openhtf/io/output/print_console.py
amyxchen: Changed to RESET

openhtf/io/output/print_console.py
amyxchen: will change after #168 get in.

openhtf/io/output/print_console.py
amyxchen: use output_lines[] and print out once at end using sys.stdout.write()

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,91 @@
+"""Module to display test summary on console."""
+import os
+import sys
+from openhtf import util
+
+
+class PrintConsole():
```
jettisonjoe: Nit-picky, but maybe `PrintToConsole`?

openhtf/io/output/print_console.py
```diff
@@ -0,0 +1,91 @@
...
+  """Print test results with failur info on console. """
```
jettisonjoe: s/failur/failure

Maybe something like:
`Print test outcome (including any failed measurements) to the console.`

openhtf/io/output/print_console.py
jettisonjoe: +1 to using the records directly here.

amyxchen: Changed to use the records directly.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,93 @@
+"""Module to display test summary on console."""
+import os
+import sys
+from openhtf import util
+from openhtf.io import test_record as tr
+from openhtf.util import measurements as mea
+
+class PrintToConsole():
+  """Print test results with failur info on console. """
```
wallacbe: "failur" should be "failure"

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,93 @@
...
+  # pylint: disable=invalid-name
+  def __call__(self, test_record):
+    indent2 = ''.join(('\n', ' '*2))
+    indent4 = ''.join(('\n', ' '*4))
+  
+    #as_dict = util.ConvertToBaseTypes(test_record)
+    result = test_record.outcome
+    status = ''
+    color = None
+    if result == tr.Outcome.PASS:
```
jettisonjoe: I would consider retrieving these colors through a lookup instead of if/else statements. Also, the string values 'PASS', 'FAIL', etc. should be available as part of the `Outcome` enum, I think.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,93 @@
+"""Module to display test summary on console."""
+import os
+import sys
+from openhtf import util
+from openhtf.io import test_record as tr
+from openhtf.util import measurements as mea
```
jettisonjoe: I generally try not to rename imported modules because I feel like doing so hurts readability. Do the other contributors (@grybmadsci, @fahhem)  have thoughts about whether/when we should do this?

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,93 @@
...
+    elif result == tr.Outcome.FAIL:
+      color = self.RED
+      status = 'FAIL'
+    else:
+      color = self.ORANGE
+      status = 'ERROR'
+
+    output_lines = []
+    msg = ''.join((os.path.basename(sys.argv[0]), ': ', status))
+    msg = ''.join((color, self.BOLD, msg, self.RESET))
```
jettisonjoe: I like building up of the output lines for a join at the end, but individual lines can also be formulated using `%` formatting.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,93 @@
...
+                              , phase_time, ']'))
+            output_lines.append(text)
+            exp_type = test_record.outcome_details[0].code
+            text = ''.join((indent2, 'exception type : ', exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2, 'exception text: ', \
+                                     test_record.outcome_details[0].description))
+            output_lines.append(text)
+
+    text = ''.join(output_lines)
```
jettisonjoe: Instead of including the `\n` characters and doing an empty string join at the end, you could just `'\n'.join` all the output lines.

openhtf/io/output/print_ToConsole.py
fahhem: The style guide says to not rename unless it's a long name, and we haven't renamed these yet because they're not that long. I've seen 'numpy as np' often because numpy code uses that namespace often enough, but this code doesn't justify a rename, it only makes it harder to grep for uses and know what 'mea' or 'tr' mean in the code

openhtf/io/output/print_ToConsole.py
fahhem: result.name will have the string name for you

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
+"""Module to display test summary on console."""
+import os
+import sys
+from openhtf import util
+from openhtf.io import test_record
+from openhtf.util import measurements as mea
+import datetime
```
jettisonjoe: Imports should be in three sections: (1) built-ins, (2) third-party deps, (3) local to this codebase. See the style guide [here.](https://www.python.org/dev/peps/pep-0008/#imports) So this should be:

```
import datetime
import os
import sys

from openhtf import util
from openhtf.io import test_record
from openhtf.util import measurements
```

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+class PrintToConsole():
```
jettisonjoe: Need two blank lines between top-level stuff, like between the imports and the start of this class definition. [Style guide](https://www.python.org/dev/peps/pep-0008/#blank-lines).

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+      self.GREEN  = ""
+      self.ORANGE = ""
+      self.RESET  = ""
+      self.BOLD   = ""
+
+    self.color_table = {
+        'PASS':self.GREEN,
+        'FAIL':self.RED,
+        'ERROR':self.ORANGE
+  }
```
jettisonjoe: This brace should be indented to match the line that opens the braces.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+  # pylint: disable=invalid-name
```
jettisonjoe: This pylint disable should probably done on the line where you want it applied, otherwise I think it's disabled for the whole block or until you explicitly re-enable it.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+            text = ''.join((indent2, 'exception type : ', exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2, 'exception text: ', \
+                                     record.outcome_details[0].description))
+            output_lines.append(text)
+
+    text = ''.join(output_lines)
+    text = ''.join((text, '\n'))
+    sys.stdout.write(text)
+  # pylint: enable=invalid-name
```
jettisonjoe: Probably no sense in re-enabling on the last line of the file since there's nothing after this to check.

Also, there be an EOF newline character to comply with the POSIX definition of a "line". I don't think this is officially part of the style guide, but it's a convention our codebase follows everywhere else.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+        'ERROR':self.ORANGE
+  }
+  # pylint: disable=invalid-name
+  def __call__(self, record):
+    indent2 = ''.join(('\n', ' '*2))
+    indent4 = ''.join(('\n', ' '*4))
+  
+    status = str(record.outcome)[-5:]
+    if status[0] == '.':
+      status = status[1:]
```
jettisonjoe: These three lines look like black magic. Can we get a comment explaining what's going on here?

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+
+    color = self.color_table[status]
+
+    output_lines = []
+    msg = ''.join((os.path.basename(sys.argv[0]), ': ', status))
+    msg = ''.join((color, self.BOLD, msg, self.RESET))
+    
+    output_lines.append(msg)
```
jettisonjoe: Here I would consider something like:
```
    msg = ''.join(color,
                  self.BOLD,
                  os.path.basename(sys.argv[0]),
                  ': ',
                  status,
                  self.RESET)
```

Additionally, are you sure `os.path.basename(sys.argv[0])` is what you want here?

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if status == 'FAIL' or status == 'ERROR':
```
jettisonjoe: I thought the plan was to pull these out of the appropriate enums instead of hard-coding which strings to check for.

Also:
```
if status in (test_record.Outcome.FAIL, test_record.Outcome.ERROR):
```

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+    color = self.color_table[status]
+
+    output_lines = []
+    msg = ''.join((os.path.basename(sys.argv[0]), ': ', status))
+    msg = ''.join((color, self.BOLD, msg, self.RESET))
+    
+    output_lines.append(msg)
+
+    if status == 'FAIL' or status == 'ERROR':      
+      phases=record.phases
```
jettisonjoe: `phases = record.phases` (spacing)

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+      for phase in phases:
+        phase_name = phase.name
+        phase_time = str((float(phase.end_time_millis) - \
+                               float(phase.start_time_millis))/1000)
```
jettisonjoe: Check the line break spot and indentation here ([style guide](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)).

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+        if status == 'FAIL':
```
jettisonjoe: See above on pulling this from the enum.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+    output_lines = []
+    msg = ''.join((os.path.basename(sys.argv[0]), ': ', status))
+    msg = ''.join((color, self.BOLD, msg, self.RESET))
+    
+    output_lines.append(msg)
+
+    if status == 'FAIL' or status == 'ERROR':      
+      phases=record.phases
+      for phase in phases:
+        phase_name = phase.name
```
jettisonjoe: Not sure `phase_name` is any more convenient or readable than just using `phase.name` directly.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+        phase_time = str((float(phase.end_time_millis) - \
+                               float(phase.start_time_millis))/1000)
+        if status == 'FAIL':
+          measured = phase.measured_values
+          measurements = phase.measurements
```
jettisonjoe: See comment on `phase_name` versus `phase.name` above.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+          for measurement in measurements:
+            res = measurements[measurement]
```
jettisonjoe: The naming on these two lines is a little confusing to me. Maybe something like `for key in measurements` would be more clear. `res` is also not clear to me.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+            if res.outcome == mea.Outcome.FAIL:
+              if len(phase_name) > 0:
+                text = ''.join(('\nfailed phase: ', phase_name, ' [time taken(s): ', \
+                        phase_time, ']'))
+                output_lines.append(text)
+                phase_name = ''
```
jettisonjoe: I think this can safely go away, especially if you just use `phase.name` directly (see above).

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+        phase_time = str((float(phase.end_time_millis) - \
+                               float(phase.start_time_millis))/1000)
+        if status == 'FAIL':
+          measured = phase.measured_values
+          measurements = phase.measurements
+          for measurement in measurements:
+            res = measurements[measurement]
+            if res.outcome == mea.Outcome.FAIL:
+              if len(phase_name) > 0:
+                text = ''.join(('\nfailed phase: ', phase_name, ' [time taken(s): ', \
```
jettisonjoe: `\` isn't needed to join lines that are already implicitly joined.

If there are a small number of variables being inserted into a proportionally larger amount of hard-coded text, consider using `%s`-style string formatting:
`text = 'failed phase: %s [time taken(s): %s' % (phase.name, phase_time)` etc.

I think the explicit `\n` can go away if the entries of output_lines are crafted right.

openhtf/io/output/print_ToConsole.py
```diff
@@ -0,0 +1,89 @@
...
+                              , phase_time, ']'))
+            output_lines.append(text)
+            exp_type = record.outcome_details[0].code
+            text = ''.join((indent2, 'exception type : ', exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2, 'exception text: ', \
+                                     record.outcome_details[0].description))
+            output_lines.append(text)
+
+    text = ''.join(output_lines)
```
jettisonjoe: If `output_lines` is crafted to contain a line of output per entry (as the name suggests), you can `\n.join()` it here instead of an empty string join.

jettisonjoe:
@amyxchen, sorry for the barrage of nitpicks. I think a lot of them are holdovers from previous comments that somehow got lost in the PR update and aren't visible in the code diff view anymore.

openhtf/io/output/print_ToConsole.py
wallacbe: "Display test summary" may be more descriptive than "print_ToConsole".
Also should use similar case and naming convention as the other files in output directory e.g.  display_test_summary.py, class DisplayTestSummary

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+      self.GREEN = ""
+      self.ORANGE = ""
+      self.RESET = ""
+      self.BOLD = ""
+
+    self.color_table = {
+        test_record.Outcome.PASS:self.GREEN,
+        test_record.Outcome.FAIL:self.RED,
+        test_record.Outcome.ERROR:self.ORANGE,
+        test_record.Outcome.TIMEOUT:self.ORANGE,
```
jettisonjoe: Another style thing: a space after each of these colons.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+    }
+  # pylint: enable=invalid-name
```
jettisonjoe: This pylint disable should probably done only on the line where you want it applied, otherwise I think it's disabled for the whole block or until you explicitly re-enable it.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+                            exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2,
+                            'exception text: ',
+                            record.outcome_details[0].description))
+            output_lines.append(text)
+
+    output_lines.append('\n')
+    text = ''.join(output_lines)
+    sys.stdout.write(text)
```
jettisonjoe: Need an EOF newline character to comply with the POSIX definition of a "line". I don't think this is officially part of the style guide, but it's a convention our codebase follows everywhere else.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+  # pylint: enable=invalid-name
+
+  def __call__(self, record):
+    indent2 = ''.join(('\n', ' '*2))
+    indent4 = ''.join(('\n', ' '*4))
+
+    # get status string from outcome enum
+    status = str(record.outcome)[-5:]
+    if status[0] == '.':
+      status = status[1:]
```
jettisonjoe: Check `record.outcome.name`. You probably don't need to parse this string yourself.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+
+    color = self.color_table[record.outcome]
+
+    output_lines = []
+    msg = ''.join((color,
+                   self.BOLD,
+                   os.path.basename(sys.argv[0]),
```
jettisonjoe: It's not clear what `os.path.basename(sys.argv[0])` represents in this context. Looks like it's probably the path to the python interpreter being used?

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+                   self.RESET,
+                   '\n'))
+
+    output_lines.append(msg)
+
+    if record.outcome in (test_record.Outcome.FAIL, test_record.Outcome.ERROR):
+      phases = record.phases
+      for phase in phases:
+        phase_name = phase.name
+        phase_time = str((float(phase.end_time_millis) -
```
jettisonjoe: Check the line break spot here ([style guide](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)).

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+                   status,
+                   self.RESET,
+                   '\n'))
+
+    output_lines.append(msg)
+
+    if record.outcome in (test_record.Outcome.FAIL, test_record.Outcome.ERROR):
+      phases = record.phases
+      for phase in phases:
+        phase_name = phase.name
```
jettisonjoe: Why not just use `phase.name` directly?

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+        phase_time = str((float(phase.end_time_millis) -
+                          float(phase.start_time_millis))/1000)
+        if record.outcome == test_record.Outcome.FAIL:
+          measured = phase.measured_values
+          measurements = phase.measurements
```
jettisonjoe: See comment on `phase.name` above.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+          for mea in measurements:
+            res = measurements[mea]
```
jettisonjoe: `mea` and `res` are a bit nondescript. It looks like `mea` is the name (or key) of the measurement, and `res` is the actual measurement record. Maybe name these something more descriptive like `key` and `record`?

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+            if res.outcome == measurement.Outcome.FAIL:
+              if len(phase_name) > 0:
+                text = ''.join(('failed phase: ',
+                                phase_name,
+                                ' [time taken(s): ',
+                                phase_time,
+                                ']'))
+                output_lines.append(text)
+                phase_name = ''
```
jettisonjoe: I think this can safely go away, especially if you just use phase.name directly (see above).

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,113 @@
...
+                            'exception type : ',
+                            exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2,
+                            'exception text: ',
+                            record.outcome_details[0].description))
+            output_lines.append(text)
+
+    output_lines.append('\n')
+    text = ''.join(output_lines)
```
jettisonjoe: Consider collapsing 112 and 113 down to just:
`sys.stdout.write('\n'.join(output_lines))`

wallacbe:
Also add the alias to  __init__.py

openhtf/io/output/__init__.py
```diff
@@ -16,10 +16,12 @@
...
 import openhtf.io.output.mfg_inspector
+import openhtf.io.output.console_summary
 
 # pylint: disable=invalid-name
 
 # Classes used for output
 OutputToJSON = json_factory.OutputToJSON
 UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
 OutputToTestRunProto = mfg_inspector.OutputToTestRunProto
+ConsoleSummary = console_summary.ConsoleSummary
```
jettisonjoe: Missing EOF newline on `__init__.py`.

amyxchen: fixed it

jettisonjoe:
LGTM, provided checks pass. Would like to get sign-off from a second approver before merging. @grybmadsci ?

wallacbe:
colorama 0.3.7 may be an option here.
Cross-platform colored terminal text.

jettisonjoe:
@wallacbe, colorama is a cool idea, but let's file that as a separate feature request and not make it block getting this output module merged.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
+"""Module to display test summary on console."""
+
+import os
+import sys
+import __main__ as main
```
grybmadsci: i'd rather use the tester name for this, the main entry point name is commonly meaningful, but not necessarily, and we have an explicit tester naming mechanism, so we might as well use it.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+    self.color_table = {
+        test_record.Outcome.PASS:self.GREEN, 
+        test_record.Outcome.FAIL:self.RED, 
+        test_record.Outcome.ERROR:self.ORANGE, 
+        test_record.Outcome.TIMEOUT:self.ORANGE, 
+    }
+  # pylint: enable=invalid-name
+
+  def __call__(self, record):
+    indent2 = ' '*2
```
grybmadsci: hm, can you take `indent` as an argument to `__init__()`, default it to 2, and do `self.indent = ' ' * indent` in `__init__()`.  Then use `indent` and `indent * 2` below, as it more clearly matches what you mean (indent once, and indent twice).  you could also use `indent, indent` below, since you're using join, whichever you like better.

openhtf/io/output/console_summary.py
jettisonjoe: Good point. This line was my suggestion, but `station_id` is at least as good here and possibly better. I'm cool either way.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+    indent4 = ' '*4
+    color = self.color_table[record.outcome]
+    output_lines = []
```
grybmadsci:
```
output_lines = [''.join((
    color, self.BOLD,
    '%s (%s): %s' % (record.code_info.name, record.station_id, record.outcome.name),
    self.RESET
))]
```

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+from openhtf import util
+from openhtf.io import test_record
+from openhtf.util import measurements as measurement
+
+
+class ConsoleSummary():
+  """Print test results with failure info on console. """
+
+  # pylint: disable=invalid-name
+  def __init__(self):
```
grybmadsci: can you take an `output_stream` argument here, default it to `sys.stdout`, and use it below instead of using `sys.stdout` directly?

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+    msg = ''.join((color,
+                   self.BOLD,
+                   main.__file__,
+                   ': ',
+                   record.outcome.name,
+                   self.RESET))
+
+    output_lines.append(msg)
+
+    if record.outcome in (test_record.Outcome.FAIL, test_record.Outcome.ERROR):
```
grybmadsci: you don't want failure information in the case of `TIMEOUT`?  this seems like it would be better checked as `record.outcome != test_record.Outcome.PASS`

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+      phases = record.phases
+      for phase in phases:
```
grybmadsci: just do `for phase in record.phases:`, there's not really any value in creating a temporary variable that is only ever used immediately on the next line, unless it's to break up complex logic, but this isn't.

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+        new_phase = True
+        phase_time = str((float(phase.end_time_millis)
```
grybmadsci: can you name this `phase_time_sec` so we don't lose track of what the units are?

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+        if record.outcome == test_record.Outcome.FAIL:
+          measured = phase.measured_values
+          measurements = phase.measurements
+          for key in measurements:
+            result = measurements[key]
+            if result.outcome == measurement.Outcome.FAIL:
+              if new_phase:
+                text = ''.join(('failed phase: ',
+                                phase.name,
+                                ' [time taken(s): ',
```
grybmadsci: prefer format strings for this sort of thing instead of `''.join()`, because it's easier to at-a-glance see what you're going for:
```
output_lines.append('failed phase: %s [ran for %.2f sec]' % (phase.name, phase_time_sec))
```

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+                                phase_time,
+                                ']'))
+                output_lines.append(text)
+                new_phase = False
+
+              failed_item = result.name
+              text = ''.join((indent2,
+                              'failed_item: ',
+                              failed_item))
+              output_lines.append(text)
```
grybmadsci:
```
output_lines.append('%sfailed_item: %s' % (self.indent, failed_item))
```

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+              text = ''.join((indent4,
+                              'measured_value: ',
+                              str(measured[failed_item])))
+              output_lines.append(text)
```
grybmadsci:
```
output_lines.append('%s%smeasured_value: %s' % (indent, indent, measured[failed_item]))
```

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+              text = ''.join((indent4,
+                              'validator: ',
+                              str(result.validators)))
+              output_lines.append(text)
```
grybmadsci:
```
output_lines.append('%svalidators:' % (self.indent * 2))
for validator in result.validators:
  output_lines.append('%svalidator: %s' % (self.indent * 3, validator))
```

openhtf/io/output/console_summary.py
```diff
@@ -0,0 +1,107 @@
...
+                            'exception type : ',
+                            exp_type))
+            output_lines.append(text)
+            text = ''.join((indent2,
+                            'exception text: ',
+                            record.outcome_details[0].description))
+            output_lines.append(text)
+
+    output_lines.append('\n')
+    text = '\n'.join(output_lines)
```
grybmadsci:
```
self.output_stream.write('\n'.join(output_lines))
```

grybmadsci:
overall LGTM, just a few nits and needs a rebase.

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: ok

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

openhtf/io/output/console_summary.py
amyxchen: done

grybmadsci:
LGTM, ready to submit as soon as this is rebased.

Followup will be to create a unit test for this code.  I can find a couple minutes to do it if nobody else can.

jettisonjoe:
Thanks for the second LGTM, @grybmadsci!

@amyxchen, I'll merge this as soon as you get a chance to rebase.

PR #183, opened 2016-03-23
Change AutoStart to take a serial argument

fahhem:
Sorry @wallacbe for the shaky API! (If this gets merged)

As discussed elsewhere, we would rather the flags be for framework things, so I'm undoing this non-frameworky flag towards that cause.

### DIFF
--- a/openhtf/exe/triggers.py
+++ b/openhtf/exe/triggers.py
@@ -38,20 +38,13 @@ TestStop:
 import logging
 import time
 
-import gflags
-
 from openhtf.io import user_input
 
-gflags.DEFINE_string('dut_serial', 'UNKNOWN_DUT_ID',
-                     'DUT serial to start the test with. '
-                     'Only use if using the AutoStart trigger.')
-
-FLAGS = gflags.FLAGS
 _LOG = logging.getLogger(__name__)
 
-def AutoStart():  # pylint: disable=invalid-name
-  """Start the test immediately with a dummy DUT ID."""
-  return FLAGS.dut_serial
+def AutoStart(serial='UNKNOWN_DUT_ID'):  # pylint: disable=invalid-name
+  """Start the test immediately with the given serial (or a dummy DUT ID)."""
+  return serial
 
 
 def AutoStop(dummy_dut_id):  # pylint: disable=invalid-name
@@ -59,7 +52,6 @@ def AutoStop(dummy_dut_id):  # pylint: disable=invalid-name
   pass
 
 
-# pylint: disable=invalid-name
 def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
                        text_input=True):
   """Make a test start trigger based on prompting the user for input."""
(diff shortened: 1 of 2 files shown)


wallacbe:
how about passing the dut_serial as a conf value?

openhtf/exe/triggers.py
```diff
@@ -63,9 +57,14 @@ def AutoStop(dummy_dut_id):  # pylint: disable=invalid-name
 def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
                        text_input=True):
   """Make a test start trigger based on prompting the user for input."""
+  # Cache the result in case we're looping.
```
grybmadsci: Hm, 'looping' doesn't necessarily mean for the same DUT.  It's actually not really clear what 'looping' really means, and we've talked about removing it entirely and leaving it up to the caller, doing something like:
```
while True:
  my_test.Execute()
```
Is easy enough.  For now, I think we should make internal 'looping' as closely match the behavior you'd get with a while loop like that (so re-prompt for a new dut_id each time).

openhtf/util/__init__.py
```diff
@@ -117,3 +117,22 @@ def get_version():
     version = 'Unknown - Perhaps openhtf was not installed via setup.py or pip.'
 
   return version
+
+
+class NonLocalResult(mutablerecords.Record('NonLocal', {'result': None})):
+  """Holds a single result as a nonlocal variable.
+
+  Comparable to using Python 3's nonlocal keyword, it allows a inner function to
```
grybmadsci: an* inner

openhtf/util/__init__.py
```diff
@@ -117,3 +117,22 @@ def get_version():
     version = 'Unknown - Perhaps openhtf was not installed via setup.py or pip.'
 
   return version
+
+
+class NonLocalResult(mutablerecords.Record('NonLocal', {'result': None})):
```
grybmadsci: Even if we don't cache dut_id in triggers.py, this is handy for phase_result, if you want to keep it and use it there instead ;)

grybmadsci:
Good call on this, Farz.  I was thinking the same thing when the first change went in, but the conf stuff wasn't quite ready and I didn't want to hold up Brandon.

openhtf/exe/triggers.py
wallacbe: To elaborate some on my loop test case, I'm iterating over a list of values and passing it into the test system via config value so I get a separate test run for each iteration.  I'm also using a fixed dut serial name passed in via flag through os.system() call so I don't have to enter it through the prompt.  My testers are currently prone to raising exceptions  so I like having the separate test run instances that start through my looping script.

fahhem:
From what I can tell, @wallacbe wants to run a test with a known serial from Python, so here's what I'm proposing as the API:

```py
test = openhtf.Test(...)
test.Execute(test_start=triggers.AutoStart(serial))
```

Or for @wallacbe:

```py
conf.Declare('dut_serial', ...)

test = openhtf.Test(...)
for value in some_list:
  conf.Load(conf_key_name=value)
  test.Execute(test_start=triggers.AutoStart(conf.dut_serial))
```

PS. This adds NonLocalResult as well. If this API is controversial, I'll separate out NonLocalResult into another PR so @grybmadsci can use it where he wants

grybmadsci:
LGTM.

jettisonjoe:
Post-facto LGTM.

PR #184, opened 2016-03-23
Massage the protos to more closely match the internal format

fahhem:
This allows updating OpenHTF quicker as internal protos change, as well
as allows sync'ing the other way easier. It also updates certain enums
and messages that were already out-of-sync.


openhtf/io/proto/assembly.proto
```diff
@@ -0,0 +1,103 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
grybmadsci: Assembly information is actually explicitly not in-scope for OpenHTF currently, so I'd rather leave this file out for now.

openhtf/io/proto/test_runs.proto
```diff
@@ -17,9 +17,41 @@ syntax = "proto2";
 
 package openhtf;
 
+import "assembly.proto";
```
grybmadsci: If this line came from a google3 version of the file and you want to keep the diff as minimal as possible, you can comment it out rather than remove it (and other references to assembly stuff).

openhtf/io/proto/test_runs.proto
```diff
@@ -71,10 +188,10 @@ message TestRun {
   // The name of the tester where this run was started
   required string tester_name = 2;
   required TestInfo test_info = 3;
-  required Status test_status = 4;
+  required Status test_status = 4 [default = CREATED];
 
-  // The openhtf cell number where this test took place
-  optional int32 cell_number = 15 [deprecated=true];
+  // The xtf cell number where this test took place
```
grybmadsci: prefer to keep this deprecated in the OpenHTF version, as it's not a thing we really support.  Arguably, it should be deprecated in the google3 version, i'm not sure anyone still uses it in any meaningful way.

openhtf/io/proto/assembly.proto
fahhem: We're going to need it (a lot of why we want to use any of this is for genealogy) and I was going to add the functionality to openhtf. If you don't want it in openhtf, then I'll have to fork the `OutputToTestRun` code unfortunately

openhtf/io/proto/test_runs.proto
```diff
@@ -93,61 +210,96 @@ message TestRun {
   // the test or otherwise responsible for the data.
   optional string operator_name = 22;
 
-  // Not supported by OpenHTF
```
grybmadsci: this is the one delta i'd like to keep, even though i know it creates some divergence between the two copies.  I'd be okay with keeping fields not deprecated if you want to include a comment at the top of the file indicating that this is a mirror of a google internal file and not all fields are used/supported by OpenHTF.

openhtf/io/proto/test_runs.proto
```diff
@@ -93,61 +210,96 @@ message TestRun {
...
-  optional string part_id = 24 [deprecated=true];
-  optional bool synthetic_dut = 25 [deprecated=true];
+  optional string lot_number = 23;
+  optional string part_id = 24;
+  optional bool synthetic_dut = 25;
 
-  //optional Assembly assembly = 26;
-  repeated Timing timings = 27 [deprecated=true];
-  repeated Phase phases = 28 [deprecated=true];
+  optional Assembly assembly = 26;
```
grybmadsci: I do want to exclude the assembly file, though, it's an entire proto file of unused stuff, and the one-line delta of having this line commented out isn't unmanageable.

grybmadsci:
LGTM, with minor changes as per comments, will merge once those are done.

fahhem:
Ah, I didn't look at the diff of test_runs.proto enough (it was a 'new file' to git until the end). I'll undo most of the undeprecation and comment removal, I mostly wanted messages and enums to be in the right place. I can leave assembly out of OpenHTF for now, but I'd like to discuss why it's out of scope

grybmadsci:
SGTM, it's not out of scope as in it's not something we want to do, it's out of scope currently because we didn't want to think about what to make the interface look like, especially with everything else so in-flux.  Happy to have that conversation and see about pulling it in now that we've mostly stabilized the other interfaces.

openhtf/io/proto/assembly.proto
jettisonjoe: @fahhem are you talking about adding assembly info support to some part of OpenHTF _besides_ `OutputToTestRun`?

fahhem: I was, but on second thought, I think it can remain internal. I'll just wrap the `_TestRunFromTestRecord` with assembly info, likely from the 'metadata' dict.

openhtf/io/proto/test_runs.proto
fahhem: The purpose of this is to, when importing to google3, drop these protos entirely and link against the internal ones. Though I may go with the other internal proto that's being made now so I'll just open-source that later rather than do this process multiple times.

openhtf/io/proto/assembly.proto
jettisonjoe: Okay, cool.

fahhem:
Updated this from its stale state. I removed the assembly info but left the comments as markers for the future. Once this is in, I'm going to upstream some code around multidims (the associated attachment), mangled multidim measurement limits (not just numeric_value, but limits too).

openhtf/io/proto/guzzle.proto
```diff
@@ -0,0 +1,33 @@
...
+syntax = "proto2";
+
+package openhtf;
+
+// Indicates the type of data stored in the payload of a
+// TestRunEnvelope object.  deflate is used for compression.
+enum PayloadType {
+  COMPRESSED_TEST_RUN = 1;
+  COMPRESSED_VARZ = 2;
+  COMPRESSED_LOGFILE = 3;
```
grybmadsci: hm, i think 4 is reserved here and next tag is 5?

openhtf/io/proto/guzzle.proto
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// Used for transport of a TestRun from the P-Server to Quantum.
+// Contains an entire TestRun proto, plus shipping information.
+message TestRunEnvelope {
```
grybmadsci: is 1 reserved?

openhtf/io/proto/guzzle.proto
```diff
@@ -0,0 +1,33 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
grybmadsci: for the purposes of OpenHTF, guzzle is the only context in which a 'test run proto' makes sense, so is this worth pulling out into a separate file?  i don't feel too strongly about it, but just a suggestion, as i don't ever see using test runs protos without using the guzzle protos (which are ~10 additional lines, you have more added lines in copyright header than content, heh).

grybmadsci:
LGTM again, pending minor nits.

openhtf/io/proto/guzzle.proto
fahhem: This file is going to change again in the next few days, but I'd like to get it in to be able to continue my own work. I'll import that file once it's ready

openhtf/io/proto/guzzle.proto
fahhem: Yeah, but as with the other comment, I'll be replacing this file soon anyway

openhtf/io/proto/guzzle.proto
fahhem: It's to match the internal format, which has it as a separate file. True, it's a tiny file, but it's along a logical line (format vs upload path), merging them is a huge pain, and combining them is a small gain.

fahhem:
Once this is in, I can remove a lot of diff-noise in my version and work on upstreaming everything else

grybmadsci:
Very important question, @fahhem ... Do you sleep, ever?

fahhem:
Define `sleep()`

fahhem:
Have I addressed your comments?

openhtf/io/proto/guzzle.proto
fahhem: Also, modifying this file would result in unexpected behavior anyway (all unknown fields are dropped), so I don't expect any churn here until it gets replaced.

fahhem:
I tried not to spam, but it's been almost a week and nothing. You said they were minor nits and I addressed all 3.

grybmadsci:
Yeah sorry I lost track of this one, had a busy week on my end, nags after a week are totally legit, my bad :)

openhtf/io/proto/guzzle.proto
grybmadsci: SGTM, keeping it separate to match internal format is fine.

PR #187, opened 2016-03-23
Convert gflags usage to argparse

fahhem:
I got stuck at how best to pass the arguments to the relevant code. `user_input.PromptManager`'s use was already ambiguous, but then `test_record.CodeInfo` wanted something in a `classmethod` and I hit a wall. There are many ways to do this, but it's either too late to think of a good one or there just isn't a good option.

Another option for `CodeInfo` may be to forgo the optimization of not capturing the source at all and create a source-wipe step that only happens later, but that brings its own `user_input.PromptManager`-like problems.

What do you think?

There is significant win in the `conf` module that even if we want to keep `gflags` in general, I'd switch `conf` to `argparse` at least because that's 45 lines of command-line parsing that we don't want to maintain when it's done just as well elsewhere.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -107,7 +107,7 @@ class Test(object):
 
   def Execute(self, loop=None,
               test_start=triggers.AutoStart, test_stop=triggers.AutoStop,
-              http_port=8888):
+              http_port=http_api.DEFAULT_HTTP_PORT):
     """Starts the framework and executes the given test.
     Args:
       test_start: Trigger for starting the test, defaults to AutoStart with a
@@ -124,7 +124,7 @@ class Test(object):
 
     _LOG.info('Executing test: %s', self.code_info.name)
     executor = exe.TestExecutor(self, test_start, test_stop)
-    http_server = http_api.Server(executor, http_port=http_port)
+    http_server = http_api.Server(executor, http_port)
     StopOnSigInt([http_server.Stop, executor.Stop])
     http_server.Start()
     executor.Start()
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -15,6 +15,7 @@
 
 """The main OpenHTF entry point."""
 
+import argparse
 import collections
 import copy
 import inspect
@@ -25,7 +26,6 @@ import signal
 import socket
 import sys
 
-import gflags
 import mutablerecords
 
 from enum import Enum
@@ -34,9 +34,11 @@ from openhtf import conf
 from openhtf import exe
 from openhtf import plugs
 from openhtf import util
+from openhtf.exe import phase_executor
 from openhtf.exe import triggers
 from openhtf.io import http_api
 from openhtf.io import test_record
+from openhtf.io import user_input
 from openhtf.util import functions
 from openhtf.util import logs
 from openhtf.util import measurements
@@ -49,7 +51,6 @@ from openhtf.util import measurements
 conf.Declare('station_id', 'The name of this test station',
              default_value=socket.gethostname())
 
-FLAGS = gflags.FLAGS
 __version__ = util.get_version()
 _LOG = logging.getLogger(__name__)
 
@@ -105,7 +106,8 @@ class Test(object):
     return self._test_info.metadata
 
   def Execute(self, loop=None,
-              test_start=triggers.AutoStart, test_stop=triggers.AutoStop):
+              test_start=triggers.AutoStart, test_stop=triggers.AutoStop,
+              http_port=8888):
     """Starts the framework and executes the given test.
     Args:
       test_start: Trigger for starting the test, defaults to AutoStart with a
@@ -122,12 +124,12 @@ class Test(object):
 
     _LOG.info('Executing test: %s', self.code_info.name)
     executor = exe.TestExecutor(self, test_start, test_stop)
-    server = http_api.Server(executor)
-    StopOnSigInt([server.Stop, executor.Stop])
-    server.Start()
+    http_server = http_api.Server(executor, http_port=http_port)
+    StopOnSigInt([http_server.Stop, executor.Stop])
+    http_server.Start()
     executor.Start()
     executor.Wait()
-    server.Stop()
+    http_server.Stop()
 
 
 class TestData(collections.namedtuple(
@@ -162,15 +164,18 @@ class TestData(collections.namedtuple(
     return plug_type_map
 
 
+def CreateArgParser():
+  """Creates an argparse.ArgumentParser for parsing command line flags."""
+  parser = argparse.ArgumentParser('OpenHTF-based testing', parents=[
+      conf.ARG_PARSER, user_input.ARG_PARSER, phase_executor.ARG_PARSER,
+      logs.ARG_PARSER])
+  return parser
+
+
 @functions.RunOnce
 def SetupFramework():
   """Sets up various bits of the framework. Only needs to be called once."""
-  try:
-    sys.argv = FLAGS(sys.argv)  # parse flags
-  except gflags.FlagsError, e:  # pylint: disable=invalid-name
-    print '%s\nUsage: %s ARGS\n%s' % (e, sys.argv[0], FLAGS)
-    sys.exit(1)
-
+  CreateArgParser().parse_args()
   logs.setup_logger()
 
 
--- a/openhtf/conf.py
+++ b/openhtf/conf.py
@@ -121,6 +121,7 @@ Loaded configuration values may be purged via the Reset() method, but this
 should only be used for testing purposes.
 """
 
+import argparse
 import functools
 import inspect
 import logging
@@ -128,22 +129,23 @@ import sys
 import threading
 import yaml
 
-import gflags
 import mutablerecords
 
+from openhtf.util import argv
 from openhtf.util import threads
 
 # If provided, --config-file will cause the given file to be Load()ed when the
 # conf module is initially imported.
-gflags.DEFINE_string('config-file', None,
-                     'File from which to load configuration values.')
+ARG_PARSER = argv.ModuleParser()
+ARG_PARSER.add_argument(
+    '--config-file', type=argparse.FileType('r'),
+    help='File from which to load configuration values.')
 
-gflags.DEFINE_multistring(
-    'config-value', [], 'Allows specifying a configuration key=value '
-    'on the command line.  The format should be --config-value=key=value. '
-    'This value will override any loaded value, and will be a string.')
-
-FLAGS = gflags.FLAGS
+ARG_PARSER.add_argument(
+    '--config-value', action='append', default=[],
+    help='Allows specifying a configuration key=value on the command line. '
+    'The format should be --config-value=key=value. This value will override '
+    'any loaded value, and will be a string.')
 
 
 class Configuration(object):  # pylint: disable=too-many-instance-attributes
@@ -181,10 +183,10 @@ class Configuration(object):  # pylint: disable=too-many-instance-attributes
       self.has_default = 'default_value' in kwargs
   # pylint: enable=invalid-name,bad-super-call,too-few-public-methods
 
-  __slots__ = ('_flags', '_logger', '_lock', '_functools', '_modules',
-               '_declarations', '_flag_values', '_loaded_values')
+  __slots__ = ('_logger', '_lock', '_modules', '_declarations',
+               '_flag_values', '_flags', '_loaded_values', 'ARG_PARSER')
 
-  def __init__(self, flags, logger, lock, _functools, **kwargs):
+  def __init__(self, logger, lock, parser, **kwargs):
     """Initializes the configuration state.
 
     We have to pull everything we need from global scope into here because we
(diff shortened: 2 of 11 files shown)


fahhem:
Oh, and the tests will fail miserably because this is severely in-progress. I'm not looking to get this merged without significant forward progress.

openhtf/__init__.py
```diff
@@ -161,14 +162,27 @@ def plug_type_map(self):
...
+                      help='User prompt timeout in seconds.')
+  # Option 2:
+  parser = argparse.ArgumentParser('OpenHTF-based testing', parents=[
+      user_input.DEFAULT_TIMEOUT_S_PARSER])
+
+
+  return parser
 
+
+@functions.RunOnce
```
jettisonjoe: Nice.

openhtf/__init__.py
```diff
@@ -161,14 +162,27 @@ def plug_type_map(self):
...
+def SetupFramework(arguments=None):
+  """Sets up various bits of the framework. Only needs to be called once."""
+  parser = arguments or CreateArgParser().parse_args()
+  # Option 3:
```
jettisonjoe: I'm leaning towards option 3.

Option 1 is the worst because it requires the framework entry logic and the `user_input` submodule to know about _each other_.

Option 2 is slightly better because the entry logic is the relationship only goes one way, but the API for doing it is less straightforward.

Option 3 is also a one-way relationship, and is slightly more straightforward, at least when you're looking at this end of it.

The fourth (and possibly best but also most effort) option would be to redesign prompt management so it doesn't need to use a singleton that's created at import time, which is where the complexity seems to stem from, and I'll take the blame for doing it this way to begin with. I was following a pattern established by Python's own `logging` module.

openhtf/conf.py
```diff
@@ -206,62 +206,16 @@ def __init__(self, flags, logger, lock, _functools, **kwargs):
...
-          indices_to_delete.add(idx)
-          if '=' not in arg:
-            indices_to_delete.add(idx + 1)
-
-    # If we found at least one value, mark the flag as not using default.
-    flag.using_default_value = False
-
-    # Return a copy of argv, minus any args we parsed.
-    return [v for i, v in enumerate(argv) if i not in indices_to_delete]
-
```
jettisonjoe: Nice.

openhtf/__init__.py
fahhem: This isn't the only place this comes up, so the fourth option is pretty untenable, unfortunately. CodeInfo does it too, and worse

jettisonjoe: Well shucks. :disappointed:

openhtf/util/functions.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
```
grybmadsci: 2016

openhtf/util/functions.py
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Utilities for functions."""
+
+def RunOnce(func):
+  """Decorate a function to only allow it to be called once."""
+  def _Wrapper(*args, **kwargs):
```
grybmadsci: can you `@functools.wraps` this?

openhtf/__init__.py
```diff
@@ -161,14 +162,27 @@ def plug_type_map(self):
...
-    sys.exit(1)
+def CreateArgParser():
+  """Creates an argparse.ArgumentParser for parsing command line flags."""
+  parser = argparse.ArgumentParser('OpenHTF-based testing')
+  # openhtf.io.user_input
+  # Option 1:
+  parser.add_argument('--prompt_timeout_s', type=int,
+                      help='User prompt timeout in seconds.')
+  # Option 2:
+  parser = argparse.ArgumentParser('OpenHTF-based testing', parents=[
```
grybmadsci: hm, should you also include the conf parser in the parents?  while we don't expect this parser to actually parse those args, without it the config flags won't show up in the help string... and those parent parsers should specify `add_help=False`, according to https://docs.python.org/2.7/library/argparse.html#parents

openhtf/conf.py
grybmadsci: +1, really like that this goes away, it made me sad to write it in the first place, but gflags doesn't have a mechanism for doing this.

test/util/functions_test.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
```
grybmadsci: it was 2014 in the last file, 2015 in this one lol... 2016 :P

fahhem: I didn't realize they were different, I just copy-pasted it from whatever file I happened to have open somewhere else :) I'll clean this up once I have a decision on the 3 options :/

openhtf/__init__.py
fahhem: Good idea! Thanks!

wallacbe:
There is a request from my team to use argparse so two parsing infrastructures dont have to be included in the virtualenv

openhtf/__init__.py
jettisonjoe: After some discussion this morning, @grybmadsci and I recommend option 3, with the added caveat of taking the flag's default from a pseudo-constant defined at the top of `user_input.py`. 

The concern being that it's weird to have a flag default here, and a default-default in the `user_input` module. It's hard in that case to understand which "default" is being respected, and you might change it one place without knowing about its use in the other. Taking the default from the `user_input` module and using it at flag declaration here alleviates the confusion of multiple defaults, and also preserves the one-way relationship of `user_input` not needing to make any assumptions about how it's being used.

wallacbe:
Should we also remove 'python-gflags==2.0' from setup.py?

jettisonjoe:
@wallacbe Yeah, once we're not using gflags anywhere, there's no reason to leave it in `setup.py`.

jettisonjoe:
@fahhem, do you feel like you have enough information to proceed with this?

fahhem:
Yeah, just been swamped with other stuff. I think I'll have time this week though

fahhem:
I haven't done the adb flags, but the rest are there. Thoughts?

wallacbe:
Anyway we can move  'python-gflags==2.0', into extras in setup.py if adb dependency is still present?

openhtf/__init__.py
```diff
@@ -105,7 +106,8 @@ def metadata(self):
     return self._test_info.metadata
 
   def Execute(self, loop=None,
-              test_start=triggers.AutoStart, test_stop=triggers.AutoStop):
+              test_start=triggers.AutoStart, test_stop=triggers.AutoStop,
+              http_port=8888):
```
grybmadsci: can you pull the 8888 into a CONSTANT at the top of the module?

openhtf/__init__.py
```diff
@@ -162,15 +164,17 @@ def plug_type_map(self):
...
+  """Creates an argparse.ArgumentParser for parsing command line flags."""
+  parser = argparse.ArgumentParser('OpenHTF-based testing', parents=[
+      conf.ARG_PARSER, user_input.ARG_PARSER, phase_executor.ARG_PARSER,
+      logs.ARG_PARSER])
+  return parser
+
 @functions.RunOnce
-def SetupFramework():
+def SetupFramework(arguments=None):
   """Sets up various bits of the framework. Only needs to be called once."""
```
grybmadsci: Can you update this docstring to describe the semantics of `arguments`?

openhtf/__init__.py

fahhem:
Man, coveralls is annoying...

openhtf/__init__.py
fahhem: Done.

grybmadsci:
Don't have time to finish looking now, but I like the overall direction.  Nice job with the Action for setting the module attribute, that's a clever way around having to plumb the args all the way through everywhere.

I'll look more and have more comments tomorrow (i expect mostly minor stuff), but thanks for doing this, it's a pretty important change for a bunch of people.

openhtf/io/user_input.py
```diff
@@ -115,7 +118,7 @@ def Respond(self, prompt_id, response):
         self._cond.notifyAll()
 
 
-# Module-level instance to achieve shared prompt state.
+# Module-level instance to achieve shared prompt state. Set by SetupFramework.
 PROMPT_MANAGER = PromptManager()
```
jettisonjoe: Did you want to actually change this so it's set by SetupFramework?

openhtf/io/user_input.py
```diff
@@ -115,7 +118,7 @@ def Respond(self, prompt_id, response):
         self._cond.notifyAll()
 
 
-# Module-level instance to achieve shared prompt state.
+# Module-level instance to achieve shared prompt state. Set by SetupFramework.
```
jettisonjoe: I'm open to doing something better than a module-level instance here if there are better ideas.

jettisonjoe:
LGTM. Nice work.

openhtf/io/user_input.py
fahhem: Long-term, I think it needs to come off the TestState and/or PhaseData objects, not 'global'.

fahhem:
And to @wallacbe's question, yeah, I moved python-gflags into the usb_plugs extra so it's not required by default anymore.

openhtf/__init__.py
```diff
@@ -162,16 +164,26 @@ def plug_type_map(self):
...
 
+def CreateArgParser():
+  """Creates an argparse.ArgumentParser for parsing command line flags."""
+  parser = argparse.ArgumentParser('OpenHTF-based testing', parents=[
+      conf.ARG_PARSER, user_input.ARG_PARSER, phase_executor.ARG_PARSER,
+      logs.ARG_PARSER])
+  return parser
+
+
 @functions.RunOnce
```
grybmadsci: the use of RunOnce with functions that take arguments scares me a little, because the arguments are silently ignored.  not sure what the best fix is for this...

openhtf/__init__.py
```diff
@@ -162,16 +164,26 @@ def plug_type_map(self):
...
-
-  logs.setup_logger()
+def SetupFramework(arguments=None):
+  """Sets up various bits of the framework. Only needs to be called once.
+
+  Args:
+    arguments: An argparse.Namespace-like object with at least the attributes
+        from logs.ARG_PARSER.
+  """
+  arguments = arguments or CreateArgParser().parse_args()
```
grybmadsci: it's a little weird that some modules use the action to save the flag values locally, but `logs.py` has to have the argument values pulled out of the namespace here and passed in.

openhtf/conf.py
```diff
@@ -194,74 +196,25 @@ def __init__(self, flags, logger, lock, _functools, **kwargs):
...
-          for decorating methods.
       **kwargs: Modules we need to access within this class.
     """
-    self._flags = flags
     self._logger = logger
     self._lock = lock
-    self._functools = _functools
     self._modules = kwargs
     self._declarations = {}
+    self.ARG_PARSER = parser
```
grybmadsci: hm, looks like you didn't mean to keep this?

openhtf/exe/phase_executor.py
```diff
@@ -32,21 +32,23 @@
 framework.
 """
 
+import argparse
```
grybmadsci: hm, i think you don't need this anymore, yeah?

openhtf/exe/phase_executor.py
```diff
@@ -32,21 +32,23 @@
...
+from openhtf.util import argv
 from openhtf.util import threads
 
-
-FLAGS = gflags.FLAGS
-gflags.DEFINE_integer('phase_default_timeout_ms', 3 * 60 * 1000,
-                      'Test phase timeout in ms', lower_bound=0)
+ARG_PARSER = argv.ModuleParser()
+ARG_PARSER.add_argument(
+    '--phase_default_timeout_s', default=3 * 60, action=argv.StoreInModule,
```
grybmadsci: let's not have the default in two places.  i think you can just leave out the default=3 * 60 here?  if that default shows up in the help, it might be worth just putting the `DEFAULT_PHASE_TIMEOUT_S` first and doing `default=DEFAULT_PHASE_TIMEOUT_S` in `add_argument`.

openhtf/io/user_input.py
```diff
@@ -21,7 +21,7 @@
 prompt state should use the openhtf.prompts pseudomodule.
 """
 
-
+import argparse
```
grybmadsci: don't think you need this anymore

openhtf/io/user_input.py
```diff
@@ -31,17 +31,20 @@
...
 _LOG = logging.getLogger(__name__)
 
-FLAGS = gflags.FLAGS
-gflags.DEFINE_integer('prompt_timeout_s',
-                       None,
-                      'User prompt timeout in seconds.')
+ARG_PARSER = argv.ModuleParser()
+ARG_PARSER.add_argument(
+    '--prompt_timeout_s', type=int, action=argv.StoreInModule,
+    target='openhtf.io.user_input.DEFAULT_TIMEOUT_S',
```
grybmadsci: hm, i think it might be better to do `'.'.join((__module__, 'DEFAULT_TIMEOUT_S'))` that way you don't have a magic string in the file that depends on the file location.

test/conf_test.py
```diff
@@ -17,16 +17,11 @@
 sys.argv.extend([
     '--config-value=flag_key=flag_value',
     '--config-value', 'other_flag=other_value',
-# You can specify arbitrary keys, but they'll get ignored if they aren't
```
grybmadsci: hm, why is this going away?

openhtf/util/logs.py
```diff
@@ -55,22 +55,24 @@ def MyRandomMethod():
...
-                   'warning', ['debug', 'info', 'warning', 'error', 'critical'],
-                   'Console log verbosity level (stderr).')
-gflags.DEFINE_boolean('quiet', False, "If True, don't output logs to stderr.")
-gflags.DEFINE_string('log-file', '', 'Filename to output logs to, if any.')
-gflags.DEFINE_enum('log-file-level', 'warning', ['debug', 'info', 'warning',
-                                                 'error', 'critical'],
-                   'Logging verbosity level for log file output.')
+LEVEL_CHOICES = ['debug', 'info', 'warning', 'error', 'critical']
+DEFAULT_LEVEL = 'warning'
+ARG_PARSER = argv.ModuleParser()
```
grybmadsci: See comment in `__init__.py`, I think we should use the `StoreInModule()` mechanism here too for consistency.

grybmadsci:
All right, LGTM pending this round of comments!  Will keep an eye out, but feel free to ping me if you get this ready to merge :)

openhtf/__init__.py
fahhem: Switched to StoreInModule across the board

openhtf/conf.py
fahhem: no, because the config module gets replaced at the end of the module, this has to be accessible on the Configuration object. I opted to keep the name as-is

openhtf/exe/phase_executor.py
fahhem: Done.

openhtf/exe/phase_executor.py
fahhem: I didn't even notice that `default=` in the code, the github coloring is better than my vim setup :/

openhtf/io/user_input.py
fahhem: Done

openhtf/io/user_input.py
fahhem: I went with `'%s.DEFAULT_TIMEOUT_S' % __name__`. `join` seems overkill for this, I'd be okay with switching to `__name__ + '.DEFAULT_TIMEOUT_S'`

openhtf/util/logs.py
fahhem: Done

test/conf_test.py
fahhem: Because this was messing with the actual sys.argv used in other tests. I added a tearDownModule to fix it

fahhem:
Okay, should be ready to merge now, thanks!

test/conf_test.py
grybmadsci: ah good call, i always forget that all the tests get run in the same Python instance.

PR #191, opened 2016-03-25
Remove capture_test_source flag

fahhem:
After using it internally, I realized this flag can't work the way it's used right now. By definition, the phases are defined before the test is created or executed, but we parse flags right before we `Execute` a test. Which means that the `capture_test_source` flag was misleading: it was always used before its value was known.

I'm trying to come up with a way to use `argparse` instead of `gflags` in another PR, and while thinking about that, the only tenable solution seems to have a flag cause the source to get removed from the `Test` when you create the test. It could be something like one of these:
```py
test = openhtf.Test(PhaseOne, PhaseTwo, strip_sources=True)
```
```py
test = openhtf.Test(PhaseOne, PhaseTwo)
test.StripSources()
```
```py
test = openhtf.Test(PhaseOne, PhaseTwo).StripSources()
```

### DIFF
--- a/openhtf/io/test_record.py
+++ b/openhtf/io/test_record.py
@@ -20,18 +20,11 @@ import collections
 import inspect
 import os
 
-import gflags
 import mutablerecords
 
 from enum import Enum
 
 from openhtf import util
-
-gflags.DEFINE_bool(
-    'capture_test_source', True,
-    'If True, inspect test source and save it in the output.')
-
-FLAGS = gflags.FLAGS
 
 
 class InvalidMeasurementDimensions(Exception):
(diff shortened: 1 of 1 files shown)


jettisonjoe:
If `StripSources()` has to be done after `Test()` anyway, why not move this into `Initialize()` with the other framework knobs we turn, and go back to having it be flag-based?

grybmadsci:
+1 to Joe's idea to twiddle this in Initialize.  It might also taken as a kwarg to Initialize to allow programmatic setting as well (depending on your deployment strategy, flags may be a nuisance)?

Either way, I'll merge this PR because it's bad to have a flag that doesn't actually do what it says, I'd rather have the feature missing for the time being, but let's roll the fix for this into the gflags -> argparse PR, as it is a pretty critical feature for current users.

PR #195, opened 2016-03-25
Refactor multicast discovery mechanism.

jettisonjoe:
This change moves the port binding side of multicast discovery to the frontend server, and the sender side to the framework instances. This results in slightly more network traffic, but enables multiple framework instances on a single host to all be discovered by a single frontend.

As a side note, this also makes frontend SIGINT response instantaneous, because the multicast listener is now a daemon thread, and the store itself is just an object.

Also addresses an (unfiled) bug where the entire contents of HTTP GET responses are sent as one long header.

### DIFF
--- a/openhtf/exe/triggers.py
+++ b/openhtf/exe/triggers.py
@@ -53,7 +53,7 @@ def AutoStop(dummy_dut_id):  # pylint: disable=invalid-name
 
 
 def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
-                       text_input=False, timeout_s=60*60*24):
+                       text_input=True, timeout_s=60*60*24):
   """Make a test start trigger based on prompting the user for input."""
   def trigger():  # pylint: disable=missing-docstring
     prompt_manager = user_input.get_prompt_manager()
@@ -63,7 +63,7 @@ def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
 
 
 def PromptForTestStop(message='Hit ENTER to complete the test.',
-                      text_input=False, timeout_s=60*60*24):
+                      text_input=True, timeout_s=60*60*24):
   """Make a test stop trigger based on prompting the user for a response."""
   def trigger(dummy_dut_id):  # pylint: disable=missing-docstring
     prompt_manager = user_input.get_prompt_manager()
--- a/openhtf/exe/triggers.py
+++ b/openhtf/exe/triggers.py
@@ -63,7 +63,7 @@ def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
 
 
 def PromptForTestStop(message='Hit ENTER to complete the test.',
-                      text_input=True, timeout_s=60*60*24):
+                      text_input=False, timeout_s=60*60*24):
   """Make a test stop trigger based on prompting the user for a response."""
   def trigger(dummy_dut_id):  # pylint: disable=missing-docstring
     prompt_manager = user_input.get_prompt_manager()
--- a/openhtf/io/frontend/__main__.py
+++ b/openhtf/io/frontend/__main__.py
@@ -93,7 +93,12 @@ class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,n
       response = requests.get('http://%s:%s' % self.hostport)
       if response.status_code == 200:
         state = json.loads(response.text)
-        self.station_id = state['framework']['station_id']
+        station_id = state['framework']['station_id']
+        if (self.station_id is not UNKNOWN_STATION_ID) and (
+            station_id != self.station_id):
+          _LOG.warning('Station (%s) underwent an identity change.',
+                       self.hostport)
+        self.station_id = station_id
         self.state = state
     except requests.RequestException as e:
       _LOG.debug('Station (%s) unreachable: %s', self.hostport, e)
@@ -128,9 +133,7 @@ class StationStore(threading.Thread):
 
   def __getitem__(self, hostport):  # pylint:disable=invalid-name
     """Provide dictionary-like access to the station store."""
-    if hostport not in self.stations:
-      self.stations[hostport] = Station(hostport)
-    station = self.stations[hostport]
+    station = self.stations.setdefault(hostport, Station(hostport))
     station.Refresh()
     return station
 
@@ -144,7 +147,7 @@ class StationStore(threading.Thread):
       port = None
       try:
         port = json.loads(response)[PING_RESPONSE_KEY]
-        self._Track(host, port)
+        self._Track(host, int(port))
       except (KeyError, ValueError):
         _LOG.debug('Ignoring unrecognized discovery response from %s: %s' % (
             host, response))
@@ -158,7 +161,7 @@ class StationStore(threading.Thread):
   def _Track(self, *hostport):
     """Start tracking the given station."""
     station = self[hostport]
-    if station.station_id == UNKNOWN_STATION_ID:
+    if station.station_id is UNKNOWN_STATION_ID:
       station.Refresh()
 
   def Stop(self):
@@ -183,7 +186,7 @@ class StationStateHandler(tornado.web.RequestHandler):
 
   def get(self, host, port):
     self.render('station.html',
-                state=self.store[(host, port)].state,
+                state=self.store[host, int(port)].state,
                 host=host,
                 port=port,
                 interval=FLAGS.poll_interval_ms)
@@ -196,7 +199,7 @@ class StationPieceHandler(tornado.web.RequestHandler):
 
   def get(self, host, port, template):
     self.render('%s.html' % template,
-                state=self.store[(host, port)].state)
+                state=self.store[host, int(port)].state)
 
 
 class PromptHandler(tornado.web.RequestHandler):
@@ -205,7 +208,7 @@ class PromptHandler(tornado.web.RequestHandler):
     self.store = store
 
   def get(self, host, port):
-    state = self.store[(host, port)].state
+    state = self.store[host, int(port)].state
     if state is not None:
       prompt = state['framework']['prompt']
       if prompt is not None:
@@ -221,7 +224,7 @@ class PromptResponseHandler(tornado.web.RequestHandler):
   def post(self, host, port, prompt_id):
     msg = json.JSONEncoder().encode(
         {'id': prompt_id, 'response': self.request.body})
-    self.store[(host, int(port))].Notify(msg)
+    self.store[host, port].Notify(msg)
     self.write('SENT')
 
 
--- a/openhtf/io/frontend/__main__.py
+++ b/openhtf/io/frontend/__main__.py
@@ -38,7 +38,6 @@ import os
 import signal
 import socket
 import sys
-import threading
 
 import gflags
 import mutablerecords
@@ -47,7 +46,6 @@ import tornado.ioloop
 import tornado.web
 
 from openhtf.io.http_api import PING_STRING
-from openhtf.io.http_api import PING_RESPONSE_KEY
 from openhtf.util import logs
 from openhtf.util import multicast
 
@@ -88,7 +86,7 @@ class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,n
     """
 
 
-class StationStore(threading.Thread):
+class StationStore(object):
   """Self-updating store of stations visible on the local network(s).
 
   Station data is stored in the 'stations' attribute, a dictionary mapping
@@ -96,7 +94,6 @@ class StationStore(threading.Thread):
   """
   def __init__(self):
     super(StationStore, self).__init__()
-    self._death_event = threading.Event()
     self.hostname = socket.gethostname()
     self.stations = collections.defaultdict(Station)
 
@@ -138,25 +135,23 @@ class StationStore(threading.Thread):
     conn.close()
     return result
 
-  def _Discover(self):
-    """Use multicast to discover stations on the local network."""
-    responses = multicast.send(PING_STRING,
-                               FLAGS.discovery_address,
-                               FLAGS.discovery_port,
-                               FLAGS.discovery_ttl)
-    for host, response in responses:
+
+  def StartDiscovery(self):
+    """Continuously listen for station advertisements."""
+    def discover(data, sender):
+      """Try to discover stations by parsing multicast messages."""
+      _LOG.debug('Received %s from %s over multicast.', data, sender)
       try:
-        port = json.loads(response)[PING_RESPONSE_KEY]
+        port = json.loads(data)[PING_STRING]
+        host = sender[0]
         self._Track(host, port)
       except (KeyError, ValueError):
-        _LOG.warning('Ignoring unrecognized discovery response from %s: %s' % (
-            host, response))
+        _LOG.warning('Ignoring unrecognized multicast message from %s: %s' % (
+            host, data))
 
-  def run(self):
-    """Continuously scan for new stations and add them to the store."""
-    while not self._death_event.is_set():
-      self._Discover()
-      self._death_event.wait(FLAGS.discovery_interval_s)
+    listener = multicast.MulticastListener(discover)
+    listener.daemon = True
+    listener.start()
 
   def _Track(self, host, port):
     """Start tracking the given station."""
@@ -184,11 +179,6 @@ class StationStore(threading.Thread):
       except KeyError:
         _LOG.warn('Malformed station state response from (%s:%s): %s',
                   host, port, response)
-
-  def Stop(self):
-    """Stop the store."""
-    self._death_event.set()
-    self.join()
 
 
 class MainHandler(tornado.web.RequestHandler):
@@ -285,11 +275,10 @@ def main(argv):
     """Handle SIGINT by stopping running executor and handler."""
     _LOG.error('Received SIGINT. Stopping frontend server.')
     tornado.ioloop.IOLoop.instance().stop()
-    store.Stop()
   signal.signal(signal.SIGINT, sigint_handler)
 
   print('Starting openhtf frontend server on http://localhost:%s.' % FLAGS.port)
-  store.start()
+  store.StartDiscovery()
   tornado.ioloop.IOLoop.instance().start()
 
 
--- a/openhtf/io/frontend/__main__.py
+++ b/openhtf/io/frontend/__main__.py
@@ -37,6 +37,7 @@ import os
 import signal
 import socket
 import sys
+import threading
 
 import gflags
 import mutablerecords
@@ -46,6 +47,7 @@ import tornado.ioloop
 import tornado.web
 
 from openhtf.io.http_api import PING_STRING
+from openhtf.io.http_api import PING_RESPONSE_KEY
 from openhtf.util import logs
 from openhtf.util import multicast
 
@@ -86,7 +88,7 @@ class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,n
     """
 
 
-class StationStore(object):
+class StationStore(threading.Thread):
   """Self-updating store of stations visible on the local network(s).
 
   Station data is stored in the 'stations' attribute, a dictionary mapping
@@ -94,6 +96,7 @@ class StationStore(object):
   """
   def __init__(self):
     super(StationStore, self).__init__()
+    self._stop_event = threading.Event()
     self.hostname = socket.gethostname()
     self.stations = collections.defaultdict(Station)
 
@@ -129,23 +132,25 @@ class StationStore(object):
     except requests.RequestException:
       _LOG.warning('Unable to post to %s:%s.' % (host, port))
 
-
-  def StartDiscovery(self):
-    """Continuously listen for station advertisements."""
-    def discover(data, sender):
-      """Try to discover stations by parsing multicast messages."""
-      _LOG.debug('Received %s from %s over multicast.', data, sender)
+  def _Discover(self):
+    """Use multicast to discover stations on the local network."""
+    responses = multicast.send(PING_STRING,
+                               FLAGS.discovery_address,
+                               FLAGS.discovery_port,
+                               FLAGS.discovery_ttl)
+    for host, response in responses:
       try:
-        port = json.loads(data)[PING_STRING]
-        host = sender[0]
+        port = json.loads(response)[PING_RESPONSE_KEY]
         self._Track(host, port)
       except (KeyError, ValueError):
-        _LOG.warning('Ignoring unrecognized multicast message from %s: %s' % (
-            host, data))
+        _LOG.warning('Ignoring unrecognized discovery response from %s: %s' % (
+            host, response))
 
-    listener = multicast.MulticastListener(discover)
-    listener.daemon = True
-    listener.start()
+  def run(self):
+    """Continuously scan for new stations and add them to the store."""
+    while not self._stop_event.is_set():
+      self._Discover()
+      self._stop_event.wait(FLAGS.discovery_interval_s)
 
   def _Track(self, host, port):
     """Start tracking the given station."""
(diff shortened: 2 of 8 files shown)


grybmadsci:
Ah, sorry I never responded to your initial RFC on this issue, been kind of a crazy week on my end.  I'm not terribly sold on this fix, for one it's pretty spammy and not a very clean solution, and secondly it doesn't fix the problem so much as relocate it to a less common use case, where one might have multiple frontends running.

If we can't come up with anything better without blowing up the scope of this, then I'd be okay with it, but I would like to spend some time f2f brainstorming this to see if we can't do better?

grybmadsci:
Hm, a  snip from https://msdn.microsoft.com/en-us/library/windows/desktop/ms740621(v=vs.85).aspx

> Using SO_REUSEADDR
> 
> The SO_REUSEADDR socket option allows a socket to forcibly bind to a port in use by
> another socket. The second socket calls setsockopt with the optname parameter set to
> SO_REUSEADDR and the optval parameter set to a boolean value of TRUE before calling
> bind on the same port as the original socket. Once the second socket has successfully
> bound, the behavior for all sockets bound to that port is indeterminate. For example, if all of
> the sockets on the same port provide TCP service, any incoming TCP connection requests
> over the port cannot be guaranteed to be handled by the correct socket — the behavior is
> non-deterministic. A malicious program can use SO_REUSEADDR to forcibly bind sockets
> already in use for standard network protocol services in order to deny access to those
> service. No special privileges are required to use this option.
> 
> If a client application binds to a port before a server application is able to bind to the same
> port, then problems may result. If the server application forcibly binds using the
> SO_REUSEADDR socket option to the same port, then the behavior for all sockets bound to
> that port is indeterminate.
>
> The exception to this non-deterministic behavior is multicast sockets. If two sockets are
> bound to the same interface and port and are members of the same multicast group, data
> will be delivered to both sockets, rather than an arbitrarily chosen one.

Notably, `If two sockets are bound to the same interface and port and are members of the same multicast group, data will be delivered to both sockets, rather than an arbitrarily chosen one.`, which sounds a lot like what we want.  Looking into Linux behavior with SO_REUSEADDR on multicast sockets.

grybmadsci:
Looks like UNIX does the same thing, from http://www.kohala.com/start/mcast.api.txt

> More than one process may bind to the same SOCK_DGRAM UDP port if the bind()
> is preceded by:
> 
> 	int one = 1;
> 	setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one))
> 
> In this case, every incoming multicast or broadcast UDP datagram destined to
> the shared port is delivered to all sockets bound to the port.

grybmadsci:
It will take a little experimentation, but I *think* the original way of doing multicast works, but all testers have to set SO_REUSEADDR on the socket, then *all* locally running testers receive ping queries from *any* frontends.  One nuance of this, I believe all processes binding to the shared socket must have the same EUID.  I'd imagine this isn't a problem in most production environments, and I think I prefer requiring this over the more spammy network configuration.

jettisonjoe:
Nice find. I wish I had found this before making the PR! :tired_face: I'll change it.

jettisonjoe:
@grybmadsci This is fixed and ready for another round of review.

openhtf/io/frontend/__main__.py
```diff
@@ -96,47 +96,41 @@ class StationStore(threading.Thread):
   """
   def __init__(self):
     super(StationStore, self).__init__()
-    self._death_event = threading.Event()
+    self._stop_event = threading.Event()
```
grybmadsci: lol, fwiw, 'kill' is fairly standard nomenclature around thread/process termination, macabre as it may be.  'stop' is typically used for system-level services and implies start/stop/restart semantics.

openhtf/io/frontend/__main__.py
```diff
@@ -96,47 +96,41 @@ class StationStore(threading.Thread):
...
     self.hostname = socket.gethostname()
     self.stations = collections.defaultdict(Station)
```
grybmadsci: this is probably a complaint of mine about the Python 'use tuples as structs' paradigm, but i don't like to see things like `(host, port)` tuples that aren't namedtuples, because it relies on an implied mapping of index to semantic meaning of respective value, that's only found in comments/docstrings.

the python socket library should provide that namedtuple, since its methods use it as inputs, but it doesn't and that makes me sad.  if you feel like it, you could make a namedtuple to use instead, but i won't give you too hard a time about it since it's really a python socket interface thing.

openhtf/io/frontend/__main__.py
jettisonjoe: I hear what you're saying, but the method that sets this event is literally called `Stop()`...

openhtf/io/frontend/__main__.py
```diff
@@ -96,47 +96,41 @@ class StationStore(threading.Thread):
...
+      result = None
+    return result
 
-      # Get the state of the openhtf instance at 192.168.2.25:10500.
-      station_state = MessageFramework(192.168.2.25, 10500)
 
-      # Respond to a prompt for the DUT's serial number:
-      MessageFramework(192.168.2.25, 10500, method='POST', 'SN-NCC1701D')
+  @staticmethod
+  def PostToFramework(host, port, message):
```
grybmadsci: `NotifyStation`?

openhtf/io/frontend/__main__.py
```diff
@@ -96,47 +96,41 @@ class StationStore(threading.Thread):
...
     super(StationStore, self).__init__()
-    self._death_event = threading.Event()
+    self._stop_event = threading.Event()
     self.hostname = socket.gethostname()
     self.stations = collections.defaultdict(Station)
 
   @staticmethod
-  def MessageFramework(host, port, method='GET', message=None):
-    """Interact with a running instance of openhtf via its HTTP API.
+  def QueryFramework(host, port):
```
grybmadsci: with a `stations` map, it probably makes more sense to call this `QueryStation`.

openhtf/io/frontend/__main__.py
grybmadsci: now that i look at it more, i actually think it might be preferable to include `station_id` in the initial ping response (instead of requiring a second RTT to get it), and indexing this based on `station_id` instead of `(host, port)`.  host and port can then be stored as attributes on the Station tuple.  And you could do something like:
```
class Station(namedtuple(...)):
  @property
  def hostport(self):
    return self.host, self.port
```
And then when you need to do socket things you can pass in `my_station.hostport`.

openhtf/io/http_api.py
```diff
@@ -119,14 +121,25 @@ def do_GET(self):  # pylint: disable=invalid-name
       """
       result = {'test': util.ConvertToBaseTypes(self.executor.GetState()),
                 'framework': util.ConvertToBaseTypes(self.executor)}
-      self.wfile.write(json.dumps(result))
+      with open(os.devnull, 'w') as f:
```
grybmadsci:
```
with open(os.devnull, 'w') as f, contextlib.redirect_stderr(f):
  ...
```

openhtf/io/http_api.py
```diff
@@ -119,14 +121,25 @@ def do_GET(self):  # pylint: disable=invalid-name
...
 
     def do_POST(self):  # pylint: disable=invalid-name
       """Parse a prompt response and send it to the PromptManager."""
-      raw = self.rfile.read()
+      length = int(self.headers.getheader('content-length'))
+      raw = self.rfile.read(length)
       data = json.loads(raw)
       user_input.get_prompt_manager().Respond(
           uuid.UUID((data['id'])), data['response'])
+      with open(os.devnull, 'w') as f:
```
grybmadsci: same as above

openhtf/io/http_api.py
```diff
@@ -119,14 +121,25 @@ def do_GET(self):  # pylint: disable=invalid-name
...
+        with contextlib.redirect_stderr(f):  # Don't spam the terminal.
+          self.send_response(200)
+          self.end_headers()
+          self.wfile.write('OK')
```
grybmadsci: might want to update `PromptManager.Respond()` to indicate if the response was actually used and propagate that info here - it would be pretty confusing if someone is at the frontend and entered a value and then that value was ignored in favor of a seemingly random one.

openhtf/util/multicast.py
```diff
@@ -101,12 +109,12 @@ def send(message,
          port=DEFAULT_PORT,
          ttl=DEFAULT_TTL,
          timeout_s=2):
-  """Sends a message on the given multicast socket and returns responses.
+  """Sends a message to the given multicast socket and returns responses.
```
grybmadsci: this might be better called `query`, since you don't just send, you also wait for responses?

openhtf/io/frontend/__main__.py
grybmadsci: right, my implication was that both are oddly named, but it's definitely a nit at best, so i wouldn't worry about fixing it :)

openhtf/io/http_api.py
```diff
@@ -119,14 +121,25 @@ def do_GET(self):  # pylint: disable=invalid-name
       """
       result = {'test': util.ConvertToBaseTypes(self.executor.GetState()),
                 'framework': util.ConvertToBaseTypes(self.executor)}
-      self.wfile.write(json.dumps(result))
+      with open(os.devnull, 'w') as f:
+        with contextlib.redirect_stderr(f):  # Don't spam the terminal.
```
fahhem: What is being spammed to the terminal? Using `redirect_stderr` isn't recommended for libraries since it affects the global sys.stderr object.

fahhem: Ah, it looks like `send_response()` calls `log_request()` which calls `log_message()`. It's recommended in the docs to override `log_message(format, *args)` with something to handle 'custom formatting'. I think we can override `log_message()` or `log_request()` to properly control the terminal output

jettisonjoe: Good call. I traced this down at the time, but for some reason didn't think to override it and opted for the brute-force redirect instead. I'll fix it.

openhtf/io/frontend/__main__.py
jettisonjoe: My thinking was that it's preferable to _allow_ multiple stations with the same ID by default rather than to ignore (a seemingly random) one, as long as we can provide further identifying info to disambiguate. That's why I chose to keep the unambiguous host/port pair used throughout the code here. Note also that in the new frontend UI mocks although station_id is the most prominent visual identifier, host/port are displayed as well just in case you end up with duplicates.

openhtf/io/frontend/__main__.py
jettisonjoe: Good point.

openhtf/io/frontend/__main__.py
jettisonjoe: Sounds good.

openhtf/io/frontend/__main__.py
grybmadsci: ah, yea i wasn't sure how much you cared about having multiple stations with the same id show up.  in that case, how about adding host and port to the `Station` class and moving `QueryStation` and `NotifyStation` there, rather than having them be `@staticmethod`s on `StationStore`?  You can still keep the `StationStore.stations` map the same, but it removes a little bit of clutter from this class.

openhtf/util/multicast.py
jettisonjoe: Cool.

openhtf/io/http_api.py
jettisonjoe: Good idea. I won't actually use this until the new frontend, but I'll plumb it through here now.

openhtf/io/frontend/__main__.py
jettisonjoe: Moving these to the Station class is a cool idea. I'll do it up and see how it looks.

jettisonjoe:
Addressed reviewer feedback. Ready for another look.

openhtf/io/frontend/__main__.py
```diff
@@ -86,6 +88,31 @@ class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,n
...
+    return result
+
+  def Notify(self, message):
+    """Send a message to an OpenHTF instance in response to a prompt.
+
+    Args:
+      message: Prompt response.
+    """
+    try:
+      print('HACKABEES %s:%s' % self.hostport)
```
grybmadsci: lol assuming this was a debug message

openhtf/io/frontend/__main__.py
```diff
@@ -244,8 +236,8 @@ def initialize(self, store):
 
   def post(self, host, port, prompt_id):
     msg = json.JSONEncoder().encode(
-      {'id': prompt_id, 'response': self.request.body})
-    self.store.MessageFramework(host, port, method='POST', message=msg)
+        {'id': prompt_id, 'response': self.request.body})
+    self.store.stations[(host, int(port))].Notify(msg)
```
grybmadsci: i don't think you need the `()` inside the `[]`, do you?  i think you can just do `.stations[host, int(port)].Notify`

openhtf/io/http_api.py
```diff
@@ -113,20 +116,39 @@ class _HTTPHandler(BaseHTTPServer.BaseHTTPRequestHandler):
...
     def do_POST(self):  # pylint: disable=invalid-name
       """Parse a prompt response and send it to the PromptManager."""
-      raw = self.rfile.read()
+      length = int(self.headers.getheader('content-length'))
+      raw = self.rfile.read(length)
       data = json.loads(raw)
-      user_input.get_prompt_manager().Respond(
+      result = user_input.get_prompt_manager().Respond(
           uuid.UUID((data['id'])), data['response'])
+      with open(os.devnull, 'w') as f:
```
grybmadsci: hm, what's with the `open(os.devnull)`s?  is this a `BaseHTTPServer` thing?

openhtf/io/frontend/__main__.py
```diff
@@ -174,7 +165,8 @@ def GetStationState(self, host, port):
       _LOG.warning(
           'Store was queried for an unknown station: %s:%s' % (host, port))
       return None
-    response = self.MessageFramework(host, port)
+    self.stations[host, port].Query()
```
grybmadsci: this is going to break in all sorts of exciting ways if you give it a host/port of an unknown station (it'll try to use the default values of None, None for host/port to build the request address of `http://None:None` i think).

openhtf/io/frontend/__main__.py
```diff
@@ -77,7 +77,9 @@
 
 
 class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,no-init
-    'Station', [], {'station_id': UNKNOWN_STATION_ID, 'state': None})):
+    'Station', [], {'hostport': (None, None),
```
grybmadsci: i think host and port, either as a tuple or as two values, should be not default-constructable; it doesn't make sense to have a Station without a host/port because you are indexing them by host/port.

openhtf/io/frontend/__main__.py
grybmadsci: sadly, the semantics of `defaultdict` don't allow for passing args to the constructor.  I would make this a normal dict and add a method to the `StationStore` class to allow for 'lookup or create' semantics.  If you want to be fancy, you could even make it `__getitem__`, and then you get dict-like access to the `StationStore` directly:
```
def __getitem__(self, hostport):
  if hostport in self.stations:
    return self.stations[hostport]
  self.stations[hostport] = Station(hostport)
  return self.stations[hostport]
```
then down in _Track:
```
def _Track(self, *hostport):
  station = self[hostport]
  ...
```
And you can remove the `station.hostport = (host, port)` line.

The value here is defaultdict has potential to do some weird stuff with create-but-not-initialized Station instances.

openhtf/io/frontend/__main__.py
```diff
@@ -96,72 +123,36 @@ class StationStore(threading.Thread):
...
                                FLAGS.discovery_ttl)
     for host, response in responses:
+      port = None
       try:
         port = json.loads(response)[PING_RESPONSE_KEY]
-        self._Track(host, port)
       except (KeyError, ValueError):
         _LOG.warning('Ignoring unrecognized discovery response from %s: %s' % (
             host, response))
+      self._Track(host, port)
```
grybmadsci: hm, moving this outside the try block means you might try to `_Track(host, None)`, but that's going to fail because there's no port set.  is that intended?

openhtf/io/frontend/__main__.py
```diff
@@ -96,72 +123,36 @@ class StationStore(threading.Thread):
...
+      self._stop_event.wait(FLAGS.discovery_interval_s)
 
   def _Track(self, host, port):
     """Start tracking the given station."""
     station = self.stations[host, port]
     if station.station_id == UNKNOWN_STATION_ID:
+      station.hostport = (host, port)
       self.GetStationState(host, port)
 
   def GetStationState(self, host, port):
```
grybmadsci: you might even move this to `Station` as well and call it `Refresh` or something.  In fact, it looks like there are no other uses of `Query()` currently, so you might just roll the `Query()` functionality directly into said `Refresh()` method.  then in `_Track()` you can:
```
if station.station_id is UNKNOWN_STATION_ID:
  station.Refresh()
```
Note the use of `is` over `==` for UNKNOWN_STATION_ID, better to use `is` for sentinel checks.

grybmadsci:
Some kind of annoying comments from me, sorry!  I think the important part is that `Station` tuples can't hang around in the weird `None, None` for host/port state.  If you find some way around that without changing as much as I commented on, that's fine too.

openhtf/io/frontend/__main__.py
jettisonjoe: This is avoided by getting rid of the `(None, None)` default hostport. Changing it back now.

openhtf/exe/triggers.py
```diff
@@ -53,7 +53,7 @@ def AutoStop(dummy_dut_id):  # pylint: disable=invalid-name
 
 
 def PromptForTestStart(message='Provide a DUT ID in order to start the test.',
-                       text_input=False, timeout_s=60*60*24):
+                       text_input=True, timeout_s=60*60*24):
```
grybmadsci: lol i made the same change in my init branch :P

openhtf/io/frontend/__main__.py
```diff
@@ -86,6 +87,31 @@ class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,n
       framework: A dictionary representation of a TestExecutor object.
       test: A dictionary representation of a TestState object.
     """
+  def Refresh(self):
+    """Update state with a GET request to OpenHTF's HTTP API."""
+    try:
+      response = requests.get('http://%s:%s' % self.hostport)
+      if response.status_code == 200:
+        state = json.loads(response.text)
+        self.station_id = state['framework']['station_id']
```
grybmadsci: probably worth logging something if `self.station_id is not UNKNOWN_STATION_ID and self.station_id != state['framework']['station_id']`.

openhtf/io/frontend/__main__.py
```diff
@@ -77,7 +77,8 @@
 
 
 class Station(mutablerecords.Record(  # pylint: disable=too-few-public-methods,no-init
```
grybmadsci: kind of a stretch to call this a Record now with all these extra methods, you might promote this to a real class, up to you, i don't feel too strongly about it, it's a small utility class either way.

openhtf/io/frontend/__main__.py
```diff
@@ -96,47 +122,17 @@ class StationStore(threading.Thread):
...
-          result = json.loads(response.read())
-    except socket.error as e:
-      result = None
-    conn.close()
-    return result
+  def __getitem__(self, hostport):  # pylint:disable=invalid-name
+    """Provide dictionary-like access to the station store."""
+    if hostport not in self.stations:
+      self.stations[hostport] = Station(hostport)
+    station = self.stations[hostport]
```
grybmadsci: `station = self.stations.setdefault(hostport, Station(hostport))`

openhtf/io/frontend/__main__.py
```diff
@@ -145,49 +141,29 @@ def _Discover(self):
...
       self._Discover()
-      self._death_event.wait(FLAGS.discovery_interval_s)
+      self._stop_event.wait(FLAGS.discovery_interval_s)
 
-  def _Track(self, host, port):
+  def _Track(self, *hostport):
     """Start tracking the given station."""
-    station = self.stations[host, port]
+    station = self[hostport]
     if station.station_id == UNKNOWN_STATION_ID:
```
grybmadsci: use `is` isntead of `==`

openhtf/io/frontend/__main__.py
```diff
@@ -145,49 +141,29 @@ def _Discover(self):
                                FLAGS.discovery_port,
                                FLAGS.discovery_ttl)
     for host, response in responses:
+      port = None
       try:
         port = json.loads(response)[PING_RESPONSE_KEY]
         self._Track(host, port)
```
grybmadsci: you might consider `int(port)` here to fail as fast as possible if you get a weird value in the response.

openhtf/io/frontend/__main__.py
```diff
@@ -207,7 +183,7 @@ def initialize(self, store):
 
   def get(self, host, port):
     self.render('station.html',
-                state=self.store.GetStationState(host, port),
+                state=self.store[(host, port)].state,
```
grybmadsci: no inner parens required

openhtf/io/frontend/__main__.py
```diff
@@ -244,8 +220,8 @@ def initialize(self, store):
 
   def post(self, host, port, prompt_id):
     msg = json.JSONEncoder().encode(
-      {'id': prompt_id, 'response': self.request.body})
-    self.store.MessageFramework(host, port, method='POST', message=msg)
+        {'id': prompt_id, 'response': self.request.body})
+    self.store[(host, int(port))].Notify(msg)
```
grybmadsci: no inner parens required: `self.store[host, int(port)].Notify(msg)` should work.

openhtf/io/frontend/__main__.py
```diff
@@ -229,7 +205,7 @@ def initialize(self, store):
     self.store = store
 
   def get(self, host, port):
-    state = self.store.GetStationState(host, port)
+    state = self.store[(host, port)].state
```
grybmadsci: you probably want to `int(port)` here too like you do in `post()`.

grybmadsci:
LGTM structurally, feel free to merge after these last few cosmetic nits.

openhtf/io/frontend/__main__.py
jettisonjoe: Noted, will consider doing this in frontend PR instead of here, though.

openhtf/io/frontend/__main__.py
jettisonjoe: kk

openhtf/io/frontend/__main__.py
jettisonjoe: Keeping the 'stop' nomenclature for consistency with the method that sets it.

openhtf/io/frontend/__main__.py
jettisonjoe: Good call. I always forget to do this. Every freaking time.

openhtf/io/frontend/__main__.py

openhtf/io/frontend/__main__.py
jettisonjoe: kk

openhtf/io/frontend/__main__.py
jettisonjoe: gp

openhtf/io/frontend/__main__.py
jettisonjoe: Another thing I always forget. Fixed.

PR #196, opened 2016-03-26
Allow a phase to be reused with different kwargs

fahhem:
This means you can parameterize a phase on something like a voltage to test at or which channel to check.

If this is considered beneficial, I would propose to extend this to the measurement names and docstrings like so:

```py
@measures(Measurement('current_at_%(volts)dV'))
def MeasureCurrent(test_data, volts=None):
  yeah, well, you know that's just like
  test_data.measurements['current_at_%d' % volts] = your opinion man

...

openhtf.Test(
    MeasureCurrent.WithArgs(volts=9),
    MeasureCurrent.WithArgs(volts=12),
    MeasureCurrent.WithArgs(volts=15), ...)
```

### DIFF
--- a/openhtf/util/measurements.py
+++ b/openhtf/util/measurements.py
@@ -138,6 +138,19 @@ class Measurement(  # pylint: disable=no-init
     self.validators.append(validator)
     return self
 
+  def WithArgs(self, **kwargs):
+    """Creates a new Measurement, see openhtf.PhaseInfo.WithArgs."""
+    new_meas = mutablerecords.CopyRecord(self)
+    if '{' in new_meas.name:
+      formatter = lambda x: x.format(**kwargs)
+    else:
+      # str % {'a': 1} is harmless if str doesn't use any interpolation.
+      # .format is as well, but % is more likely to be used in other contexts.
+      formatter = lambda x: x % kwargs
+    new_meas.name = formatter(self.name)
+    new_meas.docstring = formatter(self.docstring)
+    return new_meas
+
   def __getattr__(self, attr):
     """Support our default set of validators as direct attributes."""
     # Don't provide a back door to validators.py private stuff accidentally.
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -259,8 +259,9 @@ class PhaseInfo(mutablerecords.Record(
     """Send these keyword-arguments to the phase when called."""
     # Make a copy so we can have multiple of the same phase with different args
     # in the same test.
-    new_obj = copy.copy(self)
-    new_obj.extra_kwargs.update(kwargs)
+    new_info = mutablerecords.CopyRecord(self)
+    new_info.extra_kwargs.update(kwargs)
+    new_info.measurements = [m.WithArgs(**kwargs) for m in self.measurements]
     return new_obj
 
   def __call__(self, phase_data):
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -16,6 +16,7 @@
 """The main OpenHTF entry point."""
 
 import collections
+import copy
 import inspect
 import itertools
 import json
@@ -216,7 +217,8 @@ class PhasePlug(mutablerecords.Record(
 
 class PhaseInfo(mutablerecords.Record(
     'PhaseInfo', ['func', 'code_info'],
-    {'options': PhaseOptions, 'plugs': list, 'measurements': list})):
+    {'options': PhaseOptions, 'plugs': list, 'measurements': list,
+     'extra_kwargs': dict})):
   """Phase function and related information.
 
   Attributes:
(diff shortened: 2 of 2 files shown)


grybmadsci:
This is awesome, nice job @fahhem .. I was actually trying to come up with a nice way to do this for the past few days, because it's something I encounter in just about every test I write (we do a lot of "test X at each of a handful of wavelengths"), but nothing I came up with was quite as nice as this, in both the user-facing interface, and the simplicity of the implementation.

I'll wait til Joe gets a chance to take a look, but LGTM++ from me.

One thing you might do in the meantime, could you add an example to the `examples` directory demonstrating this?  You might roll this, `run_if`, and `timeout_s` into one example, up to you, but I wouldn't include any more than those core concepts (no plugs, measurements, for example).

grybmadsci:
Oh, one other thing that might be useful in conjunction with this (entirely up to you if you want to roll it in here or not, just throwing ideas out there) is the auto-flattening of iterables passed to the `Test()` constructor.  This would allow for easier constructs like:
```
  mytest = Test(
    InitPhase,
    (WavelengthPhase.WithArgs(wavelength=w) for w in conf.wavelengths),
    DataCollectionPhase)
```
Right now, I'm manually building a `*args` using `itertools.chain` and the like, which isn't bad, but not ideal.  I think in either case, a test engineer without a strong python background is just going to copypasta it, but I think the syntax above is a whole lot easier to have an idea what's going on than calls into `itertools` and friends.

fahhem:
I proposed that list-flattening at my first openhtf meeting, actually. It was
shot down, so I'm doing my own stuff for this (+, .extend, etc). It's not that
hard to do yourself, so I'd just write a `Flatten()` function and call that
inside of `Test()`.
In fact, I'll eventually be consolidating my shared stuff into a '<co-name>htf'
directory and there will be a `Test()` in there that will call `Flatten()`
automatically.

On Sat, Mar 26, 2016 at 8:58 AM, grybmadsci < notifications@github.com > wrote:
Oh, one other thing that might be useful in conjunction with this (entirely up
to you if you want to roll it in here or not, just throwing ideas out there) is
the auto-flattening of iterables passed to the Test() constructor. This would allow for easier constructs like:

  mytest = Test(
    InitPhase,
    (WavelengthPhase.WithArgs(wavelength=w) for w in conf.wavelengths),
    DataCollectionPhase)


Right now, I'm manually building a *args using itertools.chain and the like, which isn't bad, but not ideal. I think in either case, a test
engineer without a strong python background is just going to copypasta it, but I
think the syntax above is a whole lot easier to have an idea what's going on
than calls into itertools and friends.

grybmadsci:
:( This isn't reality TV, things don't get "shot down", they just might not make sense to the group given the context at the time.  If I recall, the reasoning at the time was that we didn't want people doing silly things that could otherwise be done by passing phases in directly.

The addition of parameterized phases as a concept is particularly appealing to me because it's a thing that I do in almost every test I've written, and I believe @wallacbe does that sort of thing a lot, too.  However, as per the project philosophy, we do tend to be very conservative about the addition of new concepts like this, hence my wanting to get at least @jettisonjoe to weigh in before merging.

I think the addition of this concept alters the context for flattening iterables of phases, because such flattening is an extension of this concept in the sense that it's likely the primary use for parameterized phases will be looping over some input value.  This puts us in a position to establish the convention of "when using `.WithArgs()` mapped over a list of values, do this, otherwise generally don't," whereas before we had no clear reason to tell users when to use `Test(PhaseOne, PhaseTwo)` vs `Test([PhaseOne, PhaseTwo])`, and that ambiguity was the primary reason for the pushback the first time.

Anyway, don't feel pressured to implement anything else, this PR is good as-is, I'm just opening the floor for re-consideration of iterable flattening; in fact, I've created a separate issue to track that discussion.

fahhem:
TV or not, it was 'shot down' because it wasn't hard for someone who does this to do themselves (aka Pythoneer) but would confuse anybody who wasn't doing it on purpose. I added the measurement name/docstring mangling, so you can now say:

```py
@measures('channel_%(channel)s_volts')
def MeasureChannelVolts(test, channel):
  ...

MeasureChannelVolts.WithArgs(channel=1).measurements[0].name == 'channel_1_volts'
```

I added the ability to use {} or %()s because:

* % is more likely to be used in other contexts and might confuse someone if they put "measures %s" to mean "measures percentages" only to get a weird "measures {'argname': value}" in the output
* {} is more familiar to folks coming from JS, ruby, etc. %s is in C/C++, but %()s is pretty uncommon.

openhtf/util/measurements.py
```diff
@@ -138,6 +138,19 @@ def WithValidator(self, validator):
     self.validators.append(validator)
     return self
 
+  def WithArgs(self, **kwargs):
+    """Creates a new Measurement,s ee openhtf.PhaseInfo.WithArgs."""
```
jettisonjoe: typo in docstring: s/s ee/ see/

jettisonjoe:
LGTM. Will weigh in on iterable flattening on the other thread.

openhtf/util/measurements.py
fahhem: Whoops! Fixed.

openhtf/util/measurements.py
```diff
@@ -138,6 +138,19 @@ def WithValidator(self, validator):
     self.validators.append(validator)
     return self
 
+  def WithArgs(self, **kwargs):
+    """Creates a new Measurement, see openhtf.PhaseInfo.WithArgs."""
+    new_meas = mutablerecords.CopyRecord(self)
+    if '{' in new_meas.name:
+      formatter = lambda x: x.format(**kwargs)
+    else:
+      # str % {'a': 1} is harmless if str doesn't use any interpolation.
```
grybmadsci: let's actually enforce that measurement names only contain alphanumeric and underscore, and don't start with a number.  this way you're guaranteed to be able to set measurements via normal attribute access instead of having to use `setattr` because you used some stupid character.

fahhem: I can check if it starts with a number here, but I can only check alphanumeric later since `WithArgs` may be chained. Could I get this in without that for now?

jettisonjoe: Fine by me, especially with a TODO.

PR #200, opened 2016-03-27
TestExecutor cleanup

grybmadsci:
Refactor TestExecutor slightly, add comments for clarity, two functional changes are:
  - Output callbacks get run in a separate thread that we `.join()` after all other framework teardown.  This prevents output callbacks from raising exceptions that show up in the TestExecutor, and also allows plugs (and other framework things) to get torn down without waiting on potentially long-running or mis-behaved output callbacks.
  - Access to `self._exit_stack` is locked.  I'm guessing it was very unlikely for race conditions to occur, and if they did we didn't care because we were shutting everything down anyway, but it really should be locked so that calls to Stop() are actually thread-safe.

### DIFF
--- a/openhtf/exe/__init__.py
+++ b/openhtf/exe/__init__.py
@@ -119,6 +119,12 @@ class TestExecutor(threads.KillableThread):
     else:
       self._RunTestLoop()
 
+  def _ResetAttributes(self):
+    """Reset local stateful attributes to None."""
+    self._exit_stack = None
+    self._test_state = None
+    self._output_thread = None
+
   @threads.Loop
   def _RunTestLoop(self):
     """Handles one whole test from start to finish.
@@ -133,17 +139,9 @@ class TestExecutor(threads.KillableThread):
 
       # Any access to self._exit_stack must be done while holding this lock.
       with self._lock:
-        # Save exit_stack on self so we can access it if Stop() is called.
+        # Initial setup of exit stack and final cleanup of attributes.
         self._exit_stack = exit_stack
-
-        # Reset some stateful attributes to None when we're done with all other
-        # teardown.
-        exit_stack.callback(lambda: setattr(self, '_exit_stack', None))
-        exit_stack.callback(lambda: setattr(self, '_test_state', None))
-        exit_stack.callback(lambda: setattr(self, '_output_thread', None))
-
-        # After teardown of everything else, wait on the output thread before
-        # attempting to start another test.
+        exit_stack.callback(self._ResetAttributes)
         exit_stack.callback(self._MaybeJoinOutputThread)
 
       # Wait here until the test start trigger returns a DUT ID.  Don't hold
@@ -162,19 +160,9 @@ class TestExecutor(threads.KillableThread):
           # best-effort basis.
           plug_manager.TearDownPlugs()
           raise TestStopError('Test Stopped.')
-
-        # Now that we have the lock, we can register this.
         exit_stack.callback(plug_manager.TearDownPlugs)
 
-        # Perform initialization of some top-level stuff we need.  The order of
-        # initialization is important here, don't mess with it unless you know
-        # what you're doing.
-        #
-        # Note that we pass exit_stack explicitly for clarity; that stack is
-        # only valid inside the current context, and we only stash a reference
-        # on self so that Stop() can access it, but otherwise it should be
-        # passed in to minimize the implied state of being in-context when
-        # these are called.
+        # Perform initialization of some top-level stuff we need.
         self._test_state = self._MakeTestState(dut_id, plug_manager, suppressor)
         executor = self._MakePhaseExecutor(exit_stack, suppressor)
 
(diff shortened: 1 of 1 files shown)


openhtf/exe/__init__.py
```diff
@@ -122,40 +131,113 @@ def _RunTestLoop(self):
...
-      executor = self._MakePhaseExecutor()
-
+      # Any access to self._exit_stack must be done while holding this lock.
+      with self._lock:
+        # Save exit_stack on self so we can access it if Stop() is called.
+        self._exit_stack = exit_stack
+
+        # Reset some stateful attributes to None when we're done with all other
+        # teardown.
+        exit_stack.callback(lambda: setattr(self, '_exit_stack', None))
```
fahhem: make a single function that does all this and add that as a callback instead, preferably as an instance method

openhtf/exe/__init__.py
```diff
@@ -87,17 +92,21 @@ def Start(self):
...
-      # Tell the stack to exit.
-      with self._current_exit_stack.pop_all() as stack:
-        # Supress the error we're about to raise.
-        stack.push(lambda *exc_details: True)
-        raise TestStopError('Stopping.')
+    with self._lock:
+      if self._exit_stack:
+        # If self._exit_stack is not None, it will reset itself back to None
+        # when close()'d.
+        self._exit_stack.close()
```
fahhem: Hm, I wonder why I went with raising a suppressed error instead of this? I think it tells the callbacks something exceptional is happening (test stopping not finishing)?

openhtf/exe/__init__.py
```diff
@@ -122,40 +131,113 @@ def _RunTestLoop(self):
...
+      with self._lock:
+        if not self._exit_stack:
+          # We shouldn't get here, but just in case something went weird with a
+          # call to Stop() and we ended up resuming execution here but the
+          # exit stack was already cleared, bail.
+          raise TestStopError('Test Stopped.')
+
+        # After teardown of everything else, wait on the output thread before
+        # attempting to start another test.
+        exit_stack.callback(self._MaybeJoinOutputThread)
```
fahhem: I would only add this to the exit stack after the output thread has been created

openhtf/exe/__init__.py
```diff
@@ -122,40 +131,113 @@ def _RunTestLoop(self):
...
+      # changes (like the transition to FINISHING).
       self._status = self.FrameworkStatus.EXECUTING
       suppressor.failure_reason = 'Failed to execute test.'
       self._ExecuteTestPhases(executor)
       self._status = self.FrameworkStatus.FINISHING
-      self.test.OutputTestRecord(self._test_state.GetFinishedRecord())
 
-  def _MakePlugManager(self, suppressor):
-    """Perform some initialization and create a PlugManager."""
+      # Output the test record now that we're done executing.  We do this in a
```
fahhem: this comment makes sense inside the `MakeAndStartOutputThread` function

openhtf/exe/__init__.py
grybmadsci: I was wondering this myself, by pushing the lambda, the TestStopError is immediately suppressed and none of the other callbacks even see it (tested it in an interactive interpreter to convince myself).  pop_all() does *copy* the stack, so maybe it was to avoid the race condition that i'm now protecting against with the lock?

openhtf/exe/__init__.py
grybmadsci: good call, done.  this kind of blew up on me, largely from overly wordy comments (that were as much for internal justification as aiding the next person in understanding the code), reduced that verbosity a bit too.

openhtf/exe/__init__.py
grybmadsci: good call, moved there and put in the docstring instead of a comment.

openhtf/exe/__init__.py
grybmadsci: the problem with that is it has to be added *before* the plug teardown is added (or else plug teardown blocks on output callbacks), hence the annoying stashing of the thread on self and 'maybe' join business :( I don't really like it, open to suggestions, but it's important that plug teardown happen first.

grybmadsci:
Thanks for the Easter code review, @fahhem :)  I'm going to be afk for a couple hours, will wait for a resolution to the last couple comments before merging.

openhtf/exe/__init__.py
```diff
@@ -122,40 +137,99 @@ def _RunTestLoop(self):
...
       self._status = self.FrameworkStatus.EXECUTING
       suppressor.failure_reason = 'Failed to execute test.'
       self._ExecuteTestPhases(executor)
       self._status = self.FrameworkStatus.FINISHING
-      self.test.OutputTestRecord(self._test_state.GetFinishedRecord())
 
-  def _MakePlugManager(self, suppressor):
-    """Perform some initialization and create a PlugManager."""
+      # Output the test record now that we're done executing.
+      self._output_thread = self._MakeAndStartOutputThread()
```
fahhem: So the order of execution is:

* Execute phases
* start outputting
* ...
* teardown plugs
* finish outputting

I was actually hoping to teardown the plugs before outputting (so the TearDown of a plug could output to the test run still), which would also serve to alleviate this weirdness where you're putting a pre-emptive strike on the output thread before it's created :)

grybmadsci: Hm, yeah that's a good point, I was considering the same thing with providing access to the record logger in BasePlug, it's kind of awkward if the output callbacks get called before plug tear down (and it's just racey if they happen in parallel).

I had done it this way because in my case i'm using the output callback to display result via GUI and block until user closes the dialog, but that's actually me conflating test stop trigger and output, so i'll fix that on my end, and switch this to be less stupid and just tear down plugs *then* output the record.

when i do that i think i'll remove the separate thread business for now.  it'll likely come back once the output mechanism gets a facelift, but i think it's overkill for the current state, what do you think?

fahhem: Sounds good! What facelift for the output mechanism? I sort of thought of
making the TestExecutor return the TestRecord at the end and have the Test
class do the outputting (after the while loop got pulled out too) since it
has the callbacks anyway.

grybmadsci: We'll have to toss some idea around.  I haven't spent too much time thinking about the best way to implement it, but I'd like to add some more safety around output callbacks (right now they can raise and it's awkward) and provide some kind of fallback mechanism.

The use for this is mainly "upload the result somewhere, and if it fails save it locally to disk".  The problem with "try to upload and always also save to disk" is that depending on the upload API, it may be hard to check "has this one already been uploaded".  Anyway we can chat about that later, going to make those changes and get this in :)

openhtf/exe/__init__.py
```diff
@@ -87,17 +92,21 @@ def Start(self):
...
+        #with self._exit_stack.pop_all() as stack:
+        #  # Supress the error we're about to raise.
+        #  stack.push(lambda *exc_details: True)
+        #  raise TestStopError('Stopping.')
     self.Kill()
 
   def Wait(self):
     """Waits until death."""
-    self.join(365*24*60*60)  # Timeout needed for SIGINT handling, so 1 year.
+    self.join(sys.float_info.max)  # Timeout needed for SIGINT handling.
```
jettisonjoe: Ah, nice, good call.

jettisonjoe:
LGTM. Love the new inline documentation.

PR #217, opened 2016-04-08
Measurements updates

grybmadsci:


openhtf/util/measurements.py
```diff
@@ -119,20 +117,22 @@ class Measurement(  # pylint: disable=no-init
...
 
   def WithUnits(self, units):
-    """Declare the units for this Measurement."""
+    """Declare the units for this Measurement, returns self for chaining."""
     self.units = units
     return self
 
   def WithDimensions(self, *dimensions):
-    """Declare dimensions for this Measurement."""
+    """Declare dimensions for this Measurement, returns self for chaining."""
```
fahhem: Do you want to go with "chainable" or "returns self for chaining"?

openhtf/util/measurements.py
```diff
@@ -151,15 +151,15 @@ def WithArgs(self, **kwargs):
     new_meas.docstring = formatter(self.docstring)
     return new_meas
 
-  def __getattr__(self, attr):
+  def __getattr__(self, attr):  # pylint: disable=invalid-name
```
fahhem: I'd rather fix pylint to understand the builtin magic methods than to add these everywhere

openhtf/util/measurements.py
```diff
@@ -151,15 +151,15 @@ def WithArgs(self, **kwargs):
...
     """Support our default set of validators as direct attributes."""
     # Don't provide a back door to validators.py private stuff accidentally.
     if attr.startswith('_') or not hasattr(validators, attr):
       raise AttributeError("'%s' object has no attribute '%s'" % (
           type(self).__name__, attr))
 
     # Create a wrapper to invoke the attribute from within validators.
-    def _WithValidator(*args, **kwargs):
+    def _WithValidator(*args, **kwargs):  # pylint: disable=invalid-name
```
fahhem: same here, I think this should be a pylintrc change instead. That can be a separate PR entirely since we have so many lint issues at the moment

fahhem:
LGTM, thanks!

openhtf/util/measurements.py
grybmadsci: agreed, but that's a separate issue that's already been filed and should be done in a separate PR :)

openhtf/util/measurements.py
grybmadsci: yep, filed a github issue to do just that a little while back.

openhtf/util/measurements.py
grybmadsci: meh don't think it really matters.  i only made the one just say 'chainable' because it wouldn't fit on one line otherwise, and it was already clear what 'chainable' meant because of these other comments.

openhtf/util/measurements.py
```diff
@@ -372,21 +362,30 @@ def _maybe_make(meas):
...
-  # If we were passed in an iterable, make sure each element is an
-  # instance of Measurement.
-  if isinstance(measurements, collections.Iterable):
-    measurements = [_maybe_make(meas) for meas in measurements]
-  else:
-    measurements = [_maybe_make(measurements)]
+  measurements = [_maybe_make(meas) for meas in measurements]
 
-  # 'descriptors' is guaranteed to be a list of Measurements here.
+  # 'measurements' is guaranteed to be a list of Measurement objets here.
```
fahhem: objects

examples/measurements.py
```diff
@@ -0,0 +1,136 @@
...
+  test.measurements.hello_again_measurement = 'Again!'
+
+
+# Multiple measurements can be specified in a single decorator, using either of
+# the above syntaxes.  Technically, these syntaxes can be mixed and matched, but
+# as a matter of convention you should always use one or the other within a
+# single decorator call.  You'll also note that you can stack multiple
+# decorations on a single phase.  This is useful if you have a handful of simple
+# measurements, and then one or two with more complex declarations (see below).
+@measures('first_measurement', 'second_measurement')
```
fahhem: btw, measurements can now have spaces and still be easily set (`setattr()` works, but for constants, ugh), so you could put a space in here to show that off (or another phase since these comments are kinda long already)

fahhem:
LGTM to me!

jettisonjoe:
LGTM; Awesome! We needed this!

grybmadsci:
All right one more pass, I overloaded this a little bit because I wanted to clean some things up around how some of the unit tests work when I started implementing the measurements.py unittest... sorry for the somewhat large diff.

openhtf/util/measurements.py
grybmadsci: done.

examples/measurements.py
grybmadsci: measurements *should* only be valid python identifiers, and i think there's an issue filed to enforce that.

test/io/output/output_test.py
```diff
@@ -13,108 +13,48 @@
...
-    with open(os.path.join(self.tempdir, 'record.testrun'), 'rb') as trfile:
-      actual = testrun_pb2.TestRun.FromString(trfile.read())
-      self._CompareOutput(
-          text_format.MessageToString(expected),
-          text_format.MessageToString(actual))
+    testrun_output = StringIO()
+    mfg_inspector.OutputToTestRunProto(testrun_output)(self.record)
+    actual = testrun_pb2.TestRun.FromString(testrun_output.getvalue())
+    data.AssertEqualsAndDiff(
+        text_format.MessageToString(self.testrun),
```
fahhem: could you store the proto in text_format instead? That would make updating this test much easier. Same with the pickle'd version, even though you don't seem to be testing against it anywhere?

openhtf/util/data.py
```diff
@@ -0,0 +1,139 @@
...
+      expected.splitlines(), actual.splitlines(),
+      fromfile='expected', tofile='actual', lineterm=''):
+    logging.error(line)
+  logging.error('^^^^^  Data diff  ^^^^^')
+
+  # Then raise the AssertionError as expected.
+  assert expected == actual
+
+
+def AssertRecordsEqualNonvolatile(first, second, indent=0):
```
fahhem: If this is meant to be a general-use method, I would pass in the `_VOLATILE_FIELDS` set here (and into the recursive calls)

test/io/output/output_test.py
fahhem: Oh, I see, it's the test record being serialized, awesome! Having a way to update it and the outputs would still be useful as we make changes to the data in the framework/outputs. If it's text it's at least diffable

openhtf/util/data.py
grybmadsci: hm, it *wasn't* intended to be, but then it kinda grew into it.  I'll refactor that.

test/io/output/output_test.py
grybmadsci: Yeah the pickled version is pre-generated input, and isn't tested against anything.  I agree that makes it super opaque to work with outside the test.  It was generated by running the `all_the_things.py` example with a pickling output callback.  I'll update `all_the_things.py` to output that pickled record by default and add a comment here as to how to re-generate it.  I may also add a switch to apply here to run the test in "generate correct output" mode.

That still doesn't really help with making it more human readable.  The `AssertEqualsAndDiff`was intended to help a bit with that :( i think to some extent it's just the nature of testing encodings like this.  I'll do what I can to make it a little less painful.

PR #228, opened 2016-04-18
Cleanup initialization and Test config.

grybmadsci:
Moved some things around a little so that we have a place to add test-side configuration options.  In the process, cleaned up our call to signal.signal so we can Execute() tests off the main thread.  Also removed support for test looping and a couple other features.

I've started adding deprecation messages where things are being removed to ease the pain of transitioning users.  Not sure about removing threads.Loop(), but I'd like to try to remove it and see if anyone complains (it's just clutter to keep it around if it isn't used internally anywhere).

Oh and if you haven't had to re-build test expected output yet, the `UPDATE_OUTPUT = True` mechanism is *awesome*.


openhtf/__init__.py
```diff
@@ -71,21 +72,33 @@ def PhaseTwo(test):
...
 
   def __init__(self, *phases, **metadata):
     code_info = test_record.CodeInfo.ForModuleFromStack(levels_up=2)
+    self._test_options = TestOptions()
     self._test_info = TestData(phases, metadata=metadata, code_info=code_info)
-    self._output_callbacks = []
-    self.loop = False
+    self._lock = threading.Lock()
+    self._stopped = False
+    signal.signal(signal.SIGINT, self.Stop)
```
fahhem: This sets the signal handler and ignores the previous one. If this was in `SetupFramework()`, then you could call that from the main thread then do your test creation/execution wherever you wanted

openhtf/__init__.py
```diff
@@ -105,31 +118,79 @@ def code_info(self):
...
-    executor.Wait()
-    http_server.Stop()
+      raise ValueError(
+          'DEPRECATED. Looping is no longer natively supported by OpenHTF, '
+          'use a while True: loop around Test.Execute() instead.')
+
+    # These internally ensure they are safe to call multiple times with no weird
+    # side effects.
+    logs.SetupLogger()
+    CreateArgParser().parse_args()
```
fahhem: there are side-effects here though, if you tried to override a value from a flag, the flag will be reparsed and set again. I had put this in `SetupFramework()` so the user could call that early, then do what they wanted with how things were setup, then call `Execute()` as necessary

openhtf/__init__.py
```diff
@@ -105,31 +118,79 @@ def code_info(self):
...
+    _LOG.error('Received SIGINT.')
+    with self._lock:
+      if self._executor:
+        # TestState str()'s nicely to a descriptive string, so let's log that
+        # just for good measure.
+        _LOG.error('Stopping Test due to SIGINT: %s', self._executor.GetState())
+        self._executor.Stop()
+      else:
+        # Once we're stopped for SIGINT, don't try to start anything else.
+        _LOG.error('No Test running, ignoring SIGINT.')
```
fahhem: This only happens (because of the else) if SIGINT comes before `Execute()` gets a few lines in. After that, Ctrl-C won't stop future executions, is that what you want?

fahhem:
I'd really prefer a separate `SetupFramework()` that lets the user insert themselves between stuff like log setup and args parsing and actual test execution. I use that for various logging and argument parsing stuff already. Otherwise, thanks for the cleanup of the loop/test_stop cruft!

openhtf/__init__.py
grybmadsci: I'm trying to steer away from global "framework" state and toward more cleanly encapsulated state.  Moving this to a top-level would force Executor to be a singleton.  There's still a ton of places where we force singleton use anyway, but I'd like to start migrating away from that model where it makes sense and is easy.

That said, to truly support multiple executors, this needs to be cleaned up a bit, I'll see what I can do.

openhtf/__init__.py
grybmadsci: Oh, it probably doesn't matter because SIGINT probably triggers the program to quit, but good catch, this should be de-dented one.

openhtf/__init__.py
```diff
@@ -164,21 +225,14 @@ def plug_type_map(self):
     return plug_type_map
 
 
-def CreateArgParser():
+def CreateArgParser(add_help=True):
```
grybmadsci: Actually I meant to reverse this and make `add_help` default to False and override it to True in Execute, this is intended to provide the user with a means to include OpenHTF flags in their own flags' help docs.  So they could do:
```
my_parser = argparse.ArgumentParser('My Test arg parser', parents=[openhtf.CreateArgParser()])
```

grybmadsci:
Definitely open to discussion around the SetupFramework() thing, with some constraints:
 - I want to avoid forcing singleton behavior wherever possible.  If there's a legitimate need, we can break this, but it can't be a hypothetical use case.
 - I want to steer away from the 'Framework' as an abstraction (as a corollary to the first point) and use 'Test' as the atomic unit.  There are inevitably going to be some global things, mostly around flags and logging, but I would prefer to avoid global framework knobs otherwise.

The idea is that `Test.Configure()` should be able to provide a place to do the sorts of things you used to think of as going into `SetupFramework()`.  What specifically did you have in mind, @fahhem, and is there a reason why it can't go in `Test.Configure()`?

fahhem:
Well, if you moved the flag parsing and logs stuff into `Configure`, then it would provide what I needed. The only problem left is that the signal handling needs to change significantly to handle non-singleton use (store and chain to the previous signal handler?). Using the arg parser as a parent is one of the use-cases, but if you re-parse flags with a different parser than before, you can alter things (such as flag reuse/replacement)

grybmadsci:
SGTM, I'll move stuff into `Configure()`.  I may punt on the complexity of the signal handler, might take a stab at it, we'll see how crazy it gets.

fahhem:
Oh, I don't care about the signal handler myself, sorry I didn't make that clear. I only mess with the logger and args at the moment

grybmadsci:
Do you know of any way to check if the arguments have already been parsed?  If there's a not too painful way to inspect that and not re-call `parse_args()` I'd be fine with that, but I don't know that there is such a way.

fahhem:
If you put it in `Configure()` it shouldn't matter, you're essentially saying "do ..., and parse the args" by calling that

openhtf/__init__.py
```diff
@@ -71,21 +72,39 @@ def PhaseTwo(test):
...
+    """DEPRECATED: Use AddOutputCallbacks() instead."""
+    # TODO(madsci): Remove this before we push to PyPI, here for transitionary
+    # purposes. 
+    raise AttributeError(
+        'DEPRECATED, use AddOutputCallbacks() instead of AddOutputCallback()')
+
+  def AddOutputCallbacks(self, *callbacks):
     """Add the given function as an output module to this test."""
-    self._output_callbacks.append(callback)
+    self._test_options.output_callbacks.extend(callbacks)
```
fahhem: which is preferred?

```py
test.Configure(output_callbacks=[...])
```
```py
test.AddOutputCallbacks(...)
```

openhtf/exe/test_state.py
```diff
@@ -87,6 +87,19 @@ def _asdict(self):
         'pending_phases': [(phase.func.__name__, phase.func.__doc__)
                            for phase in self.pending_phases]}
 
+  def GetRunningPhaseName(self):
```
fahhem: `GetLastRunPhaseName` or `GetLastRunningPhase`?

fahhem:
LGTM other than that awkward name, but I don't mind much

openhtf/exe/test_state.py
grybmadsci: ah, yeah it was modeled after `self.current_phase_record`, but a more accurate name is probably a good idea, done!

openhtf/__init__.py
grybmadsci: The latter is preferred, and that's documented in the docstring for `Configure()`.  I do agree that generally we want to avoid having multiple ways to do things, but any setter method in Python could really just poke at `foo._thing` directly rather than call `foo.SetThing` - that's similar to this case, `Configure()` only exposes output_callbacks by virtue of those callbacks being stored in the same manner as other options, but you're really not supposed to access them that way.

short answer, don't do the former, there is only one accepted way to do it, it's the latter :)

jettisonjoe: I went back and forth in my head on this one for a while, and I'm okay with keeping both ways in the mix. The value in the ease of having `AddOutputCallbacks` it worth it in my opinion because almost everyone needs output callbacks on almost every test.

openhtf/__init__.py
```diff
@@ -105,31 +124,80 @@ def code_info(self):
...
-    executor = exe.TestExecutor(self, test_start, test_stop)
-    http_server = http_api.Server(executor, http_port)
-    StopOnSigInt([http_server.Stop, executor.Stop])
-    http_server.Start()
-    executor.Start()
-    executor.Wait()
-    http_server.Stop()
+      raise ValueError(
+          'DEPRECATED. Looping is no longer natively supported by OpenHTF, '
+          'use a while True: loop around Test.Execute() instead.')
```
jettisonjoe: @wallacbe, take note.

openhtf/exe/triggers.py
```diff
@@ -42,31 +34,21 @@
...
-
-
-def PromptForTestStop(message='Hit ENTER to complete the test.',
-                      text_input=False, timeout_s=60*60*24):
-  """Make a test stop trigger based on prompting the user for a response."""
-  def trigger(dummy_dut_id):  # pylint: disable=missing-docstring
-    prompt_manager = user_input.get_prompt_manager()
-    return prompt_manager.DisplayPrompt(
-        message, text_input=text_input, timeout_s=timeout_s)
-  return trigger
```
jettisonjoe: This going away might bite users of the old frontend.

openhtf/util/threads.py
```diff
@@ -26,16 +26,24 @@ class ThreadTerminationError(SystemExit):
...
-  _proc.once = fn  # way for tests to invoke the function once
-                   # you may need to pass in "self" since this may be unbound.
-  return _proc
+  if not force:
+    raise AttributeError(
+        'threads.Loop() is DEPRECATED.  If you really like this and want to '
+        'keep it, file an issue at https://github.com/google/openhtf/issues '
+        'and use it as @Loop(force=True) for now.')
+
+  def RealLoop(fn):
```
jettisonjoe: This threw _me_ for a real loop for a second there.

openhtf/exe/triggers.py
grybmadsci: ah, think we should leave it for now until we have a better solution for 'recent tests', then?

openhtf/util/threads.py
grybmadsci: lol yeah i thought i'd be all sorts of clever on this one :)  it's short-term cleverness hehe

jettisonjoe:
LGTM. On the signal handler point, I don't think it's urgent; we can punt on it for now and discuss solutions later.

Issue #230, opened 2016-04-20
Exceptions raised getting code info should not crash the framework.


grybmadsci:
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "build/bdist.linux-x86_64/egg/tbtech/tests/metrology.py", line 78, in <module>
  File "/usr/local/lib/python2.7/dist-packages/openhtf/util/measurements.py", line 383, in decorate
    phase = openhtf.PhaseInfo.WrapOrCopy(wrapped_phase)
  File "/usr/local/lib/python2.7/dist-packages/openhtf/__init__.py", line 312, in WrapOrCopy
    func = cls(func, test_record.CodeInfo.ForFunction(func))
  File "/usr/local/lib/python2.7/dist-packages/openhtf/io/test_record.py", line 98, in ForFunction
    source = inspect.getsource(func)
  File "/usr/lib/python2.7/inspect.py", line 701, in getsource
    lines, lnum = getsourcelines(object)
  File "/usr/lib/python2.7/inspect.py", line 690, in getsourcelines
    lines, lnum = findsource(object)
  File "/usr/lib/python2.7/inspect.py", line 538, in findsource
    raise IOError('could not get source code')
IOError: could not get source code

grybmadsci:
I only see this outside a virtualenv, haven't investigated the root cause yet, but at the very least this should be a warning a a fatal crash and burn.

fahhem:
Even the python stacktrace couldn't get the source code, and it looks like you might be doing something fancy with compiled python or a ZipImporter? I agree, we should wrap the getsource functions with a try-except. I can take this one if you haven't started it

grybmadsci:
I don't think we were doing anything fancy, just running on an ubuntu install, but outside a virtualenv.

jettisonjoe:
Outside a virtualenv, meaning using the default system python?

grybmadsci:
Yep, default system python on Ubuntu.

PR #231, opened 2016-04-21
Get source code more safely

fahhem:
For #230


openhtf/io/test_record.py
```diff
@@ -78,6 +78,13 @@ class PhaseRecord(  # pylint: disable=too-few-public-methods,no-init
   """
 
 
+def _GetSourceSafely(obj):
+  try:
+    return inspect.getsource(obj)
+  except IOError:
+    return ''
```
grybmadsci: probably worth logging a warning when this happens, otherwise, LGTM, thanks!

fahhem:
Added a non-spammy log message for the first source code check to fail. I had to add back arguments to functions.CallOnce, however.

openhtf/util/functions.py
```diff
@@ -18,28 +18,37 @@
 import inspect
 
 
-def CallOnce(func):
+def CallOnce(func=None, allow_args=False):
```
grybmadsci: I really don't like this fake memoization, even if it's explicitly invoked via the `allow_args` keyword.  At one point I thought we had `LogOnce` functionality, but I can't seem to find it.  How about implementing that instead of a generic utility function like this?

fahhem:
Now there's a LogOnce in logs.py

openhtf/util/logs.py
```diff
@@ -88,6 +89,20 @@ def MyRandomMethod():
...
+
+LogRecord = collections.namedtuple(
+    'LogRecord', 'level logger_name source lineno timestamp_millis message')
+
+
+def LogOnce(log_func, msg, *args, **kwargs):
+  """"Logs a message only once."""
+  if msg not in _LOGONCE_SEEN:
+    log_func(msg, *args, **kwargs)
+    # Only check the message since it's likely from the source code so lifespan
```
grybmadsci: i'm not a huge fan of keying on the message, would rather key on the stack frame that it was called from or something, but i have a feeling that's going to be annoying to implement, so let's just punt for now, this looks good enough, thanks!

PR #232, opened 2016-04-22
Allow overriding the argument parser

fahhem:
Added some docs for overriding and added a keyword arg to `Test.Configure()`

Also put arg parsing before logging setup to allow args to modify logging.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -94,7 +94,7 @@ class Test(object):
   def AddOutputCallback(self, callback):
     """DEPRECATED: Use AddOutputCallbacks() instead."""
     # TODO(madsci): Remove this before we push to PyPI, here for transitionary
-    # purposes. 
+    # purposes.
     raise AttributeError(
         'DEPRECATED, use AddOutputCallbacks() instead of AddOutputCallback()')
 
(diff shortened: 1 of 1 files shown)


fahhem:
Added `parse_all_args` because there is a `Configure()` call in `__init__()` that can cause problems.

If the library user wants to allow new arguments, they need to provide their own `ArgumentParser()` instance to `Configure()`. If they want to disallow unknown args, they pass `parse_all_args=True` in as well. By default, openhtf won't disallow unknown args because the current API doesn't allow that to happen without punishing users with external arguments.

grybmadsci:
This seems like a bit much complexity, what are the reasons for not simply changing `parse_args` to `parse_known_args`?

fahhem:
An attempt to see if you cared for the strictness of it. Since you don't, I'll drop any of that related code in a sec

fahhem:
Alright, done now

grybmadsci:
Ah, I mostly just don't know the specifics of any side effects of calling `parse_args()` multiple times.  This is a little easier to follow I think, though, so until there's a need to make it more complicated, might as well keep it simple.  Thanks for the fix!

PR #233, opened 2016-04-22
Add unittests for PhaseInfo

### LINKED ISSUE #207: unit test phase info stuff

fahhem:
Fixes #207

### DIFF
--- a/test/phase_info_test.py
+++ b/test/phase_info_test.py
@@ -34,31 +34,31 @@ class TestPhaseInfo(unittest.TestCase):
 
   def testBasics(self):
       phase = openhtf.PhaseInfo.WrapOrCopy(PlainFunc)
-      assert phase.func is PlainFunc
-      assert len(phase.plugs) == 0
-      assert phase.name == 'PlainFunc'
-      assert phase.doc == 'Plain Docstring'
+      self.assertIs(phase.func, PlainFunc)
+      self.assertEqual(0, len(phase.plugs))
+      self.assertEqual('PlainFunc', phase.name)
+      self.assertEqual('Plain Docstring', phase.doc)
       phase(None)
 
       test_phase = openhtf.PhaseInfo.WrapOrCopy(NormalTestPhase)
-      assert test_phase.name == 'NormalTestPhase'
-      assert test_phase(None) == 'return value'
+      self.assertEqual('NormalTestPhase', test_phase.name)
+      self.assertEqual('return value', test_phase(None))
 
   def testMultiplePhases(self):
       phase = openhtf.PhaseInfo.WrapOrCopy(PlainFunc)
       second_phase = openhtf.PhaseInfo.WrapOrCopy(phase)
       for attr in type(phase).all_attribute_names:
         if attr == 'func': continue
-        assert getattr(phase, attr) is not getattr(second_phase, attr)
+        self.assertIsNot(getattr(phase, attr), getattr(second_phase, attr))
 
   def testWithArgs(self):
       phase = openhtf.PhaseInfo.WrapOrCopy(ExtraArgFunc)
       phase = phase.WithArgs(input='input arg')
       result = phase(None)
-      assert result == 'input arg'
+      self.assertEqual('input arg', result)
 
       second_phase = phase.WithArgs(input='second input')
       first_result = phase(None)
       second_result = second_phase(None)
-      assert first_result == 'input arg'
-      assert second_result == 'second input'
+      self.assertEqual('input arg', first_result)
+      self.assertEqual('second input', second_result)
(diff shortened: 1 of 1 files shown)


test/phase_info_test.py
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import openhtf
+
+
+def PlainFunc():
```
grybmadsci: phases will always take at least one positional arg, so this isn't representative of how PhaseInfo is actually used.

test/phase_info_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+def ExtraArgFunc(input=None):
+  return input
+
+
+class TestPhaseInfo(unittest.TestCase):
+
+  def testBasics(self):
+      phase = openhtf.PhaseInfo.WrapOrCopy(PlainFunc)
+      assert phase.func is PlainFunc
```
grybmadsci: prefer `self.assertIs` and `self.assertEqual` over just assert

test/phase_info_test.py
```diff
@@ -0,0 +1,61 @@
...
+      assert len(phase.plugs) == 0
+      assert phase.name == 'PlainFunc'
+      assert phase.doc == 'Plain Docstring'
+      phase(None)
```
grybmadsci: can you make `phase` return something and assert you get that value out here? to make sure return values are plumbed through correctly.

test/phase_info_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+  def testMultiplePhases(self):
+      # Tests if you can call WrapOrCopy on a PhaseInfo object and modify the
+      # result without modifying the original.
+      phase = openhtf.PhaseInfo.WrapOrCopy(PlainFunc)
+      phase.plugs.append(1)
```
grybmadsci: you don't really have to modify some random attribute, it's probably safer to use `assertIsNot` to compare phase and second_phase and 
```
for attr in phase.all_attribute_names:
  self.assertIsNot(getattr(phase, attr), getattr(second_phase, attr))
```
This makes sure all attributes are copied and not just plugs, and is more future-proofed against changes in `PhaseInfo` contents.

test/phase_info_test.py
fahhem: There is a branch in the `__call__` for this exact situation, I'll add another case for the extra arg

test/phase_info_test.py
fahhem: pytest has a smart assert system in place that can convert this to a well-documented output, negating the need for `self.assert*`. I can switch to those for consistency with other files, but pytest provides a lot of features that shouldn't be lost just because they're not available internally

test/phase_info_test.py
```diff
@@ -0,0 +1,61 @@
...
+      assert len(phase.plugs) == 1
+
+      second_phase = openhtf.PhaseInfo.WrapOrCopy(phase)
+      second_phase.plugs.append(2)
+      assert len(phase.plugs) == 1
+      assert len(second_phase.plugs) == 2
+
+  def testWithArgs(self):
+      phase = openhtf.PhaseInfo.WrapOrCopy(ExtraArgFunc)
+      phase = phase.WithArgs(input='input arg')
```
grybmadsci: instead of reassigning to phase, you might save the original and do `assertIsNot` checks between the three.  it's a little redundant with the above test, so i wouldn't bother checking attributes as well, but a good sanity check that `.WithArgs()` is making new things instead of modifying the original.

test/phase_info_test.py
fahhem: included in the other test

test/phase_info_test.py
fahhem: I'm checking that with the second_phase down below

test/phase_info_test.py
grybmadsci: ah yeah i noticed and meant to go back and delete this comment, haha, sorry!

test/phase_info_test.py
grybmadsci: Hm, that's neat to know, but pytest is really a layer up from here.  It's not likely anyone will run these tests outside of `setup.py test`, but there's no particular reason to not use base `unittest` features, I think, and it would be better to be consistent with all the other tests for now.

test/phase_info_test.py
grybmadsci: Ah, right I'd forgotten that we even supported that... that would be kind of a silly test phase lol, but good to test that code path.

test/phase_info_test.py
fahhem: one day I'm going to get a decent PR past you with no nitpicking...

PR #234, opened 2016-04-22
Move TestExecutor to Test creation time

fahhem:
Another step in #131

### DIFF
--- a/openhtf/exe/__init__.py
+++ b/openhtf/exe/__init__.py
@@ -67,11 +67,11 @@ class TestExecutor(threads.KillableThread):
                          ['CREATED', 'START_WAIT', 'INITIALIZING', 'EXECUTING',
                           'STOP_WAIT', 'FINISHING'])
 
-  def __init__(self, test, test_start):
+  def __init__(self, test):
     super(TestExecutor, self).__init__(name='TestExecutorThread')
 
     self.test = test
-    self._test_start = test_start
+    self._test_start = None
     self._exit_stack = None
     self._test_state = None
     self._output_thread = None
@@ -83,6 +83,9 @@ class TestExecutor(threads.KillableThread):
     return {'station_id': conf.station_id,
             'prompt': user_input.get_prompt_manager().prompt,
             'status': self._status.name}
+
+  def SetTestStart(self, test_start):
+    self._test_start = test_start
 
   def Start(self):
     """Style-compliant start method."""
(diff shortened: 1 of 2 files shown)


openhtf/exe/__init__.py
```diff
@@ -36,6 +36,11 @@ class TestStopError(Exception):
   """Test is being stopped."""
 
 
+def _InvalidTestStart():
```
grybmadsci: Hm, I'm not sure we need to enforce this.  test_start is more of a 'wait for this before starting', so None implies "don't wait for anything before starting", which seems reasonable to me.  AutoStart has been deprecated for a similar reason - starting the test automatically is the default not the exceptional case.  Can we just check for None in `TestExecutor._WaitForTestStart()` and return immediately in that case?

fahhem: What do we do for the serial if it's None?

On Fri, Apr 22, 2016 4:05 PM, grybmadsci notifications@github.com wrote:
In openhtf/exe/__init__.py :

> @@ -36,6 +36,11 @@ class TestStopError(Exception):
>    """Test is being stopped."""
>  
>  
> +def _InvalidTestStart():


Hm, I'm not sure we need to enforce this. test_start is more of a 'wait for this
before starting', so None implies "don't wait for anything before starting",
which seems reasonable to me. AutoStart has been deprecated for a similar reason
- starting the test automatically is the default not the exceptional case. Can
we just check for None in TestExecutor._WaitForTestStart() and return immediately in that case?

grybmadsci:
Aside from the one comment on the `test_start` mechanism, LGTM.

openhtf/exe/__init__.py
grybmadsci: We assume it'll be set by the end of the test.  This was originally hacked into XTF because of board test - you don't know the USB serial until you power up the device, which required starting the test without knowing the serial yet.  This turned out to be a common enough case that it's been built into OpenHTF for a while now.  In fact, the ability to set the `dut_id` by returning it from the `test_start` callback is really a convenience at this point.

fahhem: Alright, now no test_start just leaves the dut_id as None. If they don't set the dut_id, they'll get some weird behavior later.

openhtf/exe/__init__.py
```diff
@@ -166,6 +169,8 @@ def _OutputTestRecord(self):
 
   def _WaitForTestStart(self, suppressor):
     """Wait for the test start trigger to return a DUT ID."""
+    if self._test_start is None:
+     return
```
grybmadsci: i think there's only 1 space indent here, but should be 2?

fahhem: whoops!

PR #235, opened 2016-04-22
Fixup some debt from TestData

fahhem:
`TestState` now only takes `TestData` instead of the whole `Test` object, and `test.data` is used when accessing test data from within `TestExecutor`.

### DIFF
--- a/openhtf/exe/test_state.py
+++ b/openhtf/exe/test_state.py
@@ -51,24 +51,24 @@ class TestState(object):
   """This class handles tracking the state of the test.
 
   Args:
-    test: openhtf.Test instance describing the test to run.
+    test_data: openhtf.TestData instance describing the test to run.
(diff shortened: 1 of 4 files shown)


openhtf/exe/test_state.py
```diff
@@ -58,17 +58,17 @@ class TestState(object):
   """
   State = Enum('State', ['CREATED', 'RUNNING', 'COMPLETED'])
 
-  def __init__(self, test, plug_map, dut_id, station_id):
+  def __init__(self, test_data, plug_map, dut_id, station_id):
```
grybmadsci: don't forget to update the docstring too

fahhem: Done

grybmadsci:
LGTM, but this is rebased on #234 so I'll wait til we get that in to merge this one.

Issue #167, opened 2016-03-12
Validate measurements at set-time rather than end of test.


grybmadsci:
This will better allow for more up-to-date visualization on the frontend, and we can roll in a fix for #4

jettisonjoe:
Since the new frontend is going to want this, I hope to roll it into the work I'm currently doing.

PR #236, opened 2016-04-24
Validate measurements when set rather than at end of test.

grybmadsci:
Fix for #167


openhtf/util/measurements.py
```diff
@@ -31,6 +31,18 @@
 themselves are similarly output in the 'measured_values' field on the
 PhaseRecord.  See test_record.py for more information.
 
+Validation of undimensioned measurements happens when they are set, so that
+users of the HTTP API can see PASS/FAIL outcome on those measurements
+immediately after they are set.  Multidimensional measurements, however,
+don't usually make sense to validate until all data is available, so they
+instead enter a PARTIALLY_SET outcome state until the end of the test phase,
+at which point they are validated and become with PASS or FAIL.  Note that
+validators are only called at the end of the phase if at least one value was
```
fahhem: validators on multidim measurements or all measurements?

openhtf/util/measurements.py
```diff
@@ -31,6 +31,18 @@
...
+set in the multidimensional measurement, otherwise it remains UNSET.
+
+See examples/validators.py for some examples on how to define and use custom
```
fahhem: I think you forgot to 'git add' this?

openhtf/util/measurements.py
grybmadsci: only multidim measurements by virtue of undimensioned measurements getting validated when set, and therefore never getting called at the end of the phase regardless.

openhtf/util/measurements.py
grybmadsci: nope it just doesn't exist yet :)

openhtf/util/measurements.py
```diff
@@ -189,100 +204,115 @@ class MeasuredValue(object):
...
-      # We have dimensions, create the tuples to output
-      return [tuple(dimensions) + (value,) for dimensions, value in
-              self.values.iteritems()]
-    elif self.num_dimensions == 1:
-      # With only a single dimension, we don't have to tuple() them.
-      return self.values.items()
-    else:
-      # We have no dimensions, just output our value.
-      return self.value
+    return [tuple(dimensions) + (value,) for dimensions, value in
```
fahhem: dimensions should already be a tuple, as set by `__setitem__` above

openhtf/util/measurements.py
```diff
@@ -189,100 +204,115 @@ class MeasuredValue(object):
...
-          self.name, coordinates, self.values[coordinates], value)
-    self.values[coordinates] = value
+          self.name, coordinates, self.value_dict[coordinates], value)
+    self.value_dict[coordinates] = value
+    self.measurement.outcome = Outcome.PARTIALLY_SET
 
   def __getitem__(self, coordinates):  # pylint: disable=invalid-name
-    return self.values[coordinates]
+    # Let the AttributeError propagate as that's fairly intuitive to handle.
+    return self.value_dict[coordinates]
```
fahhem: you don't wrap single dimensions in tuples here, so you can set [1] and then fail on getting [1] later

openhtf/util/measurements.py
```diff
@@ -189,100 +204,115 @@ class MeasuredValue(object):
...
+  setters that provide attribute and dict-like interfaces.
 
   A Collection is created with a list of Measurement objects (defined above).
   Measurements can't be added after initialization, only accessed and set.
 
   MeasuredValue values can be set as attributes (see below).  They can also be
-  read as attributes, but you get a MeasuredValue object back if the measurement
-  accessed is dimensioned (this is how setting of dimensioned measurements
-  works, and so is unavoidable).
+  read as attributes, but you get a DimensioendMeasuredValue object back if the
```
fahhem: mispelled Dimensioned

openhtf/util/measurements.py
```diff
@@ -326,18 +357,25 @@ def __getattr__(self, name):  # pylint: disable=invalid-name
...
 
   def __getitem__(self, name):  # pylint: disable=invalid-name
     self._AssertValidKey(name)
+
+    # __getitem__ is used to set dimensioned values via __setitem__ on the
+    # DimensionedMeasuredValue object, so we can't do any checking here.
     if self._measurements[name].dimensions:
-      return self._values.setdefault(name, MeasuredValue.ForMeasurement(
-          self._measurements[name]))
+      return self._values.setdefault(
```
fahhem: nit: using `setdefault` means making an unset value for every call, whether the value was set before or not. This can be costly for repeated accesses, especially with multidimensional measurements we expect repeated accesses for.

openhtf/util/measurements.py
fahhem: I meant the sentence was confusing

grybmadsci: clarified.

openhtf/util/measurements.py
grybmadsci: right, cleaned up.

openhtf/util/measurements.py
grybmadsci: ah right, good catch.

openhtf/util/measurements.py
grybmadsci: fixed

openhtf/util/measurements.py
grybmadsci: hm, i think that's a bit of a micro-optimization, but wouldn't hurt, done.

PR #241, opened 2016-05-03
Separate plugs and execution

fahhem:
Moves the plug manager creation into early test creation and makes its API more concrete.

I also separated the output code into `Test.Execute()` which cleaned up that separation a little, too. I'm not attached to that if either of you don't want that.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -82,7 +82,7 @@ class Test(object):
     self._test_options = TestOptions()
     self._test_data = TestData(phases, metadata=metadata, code_info=code_info)
     self._lock = threading.Lock()
-    self._executor = exe.TestExecutor(self._test_data)
+    self._executor = exe.TestExecutor(self._test_data, plugs.PlugManager())
     self._stopped = False
     # Make sure Configure() gets called at least once before Execute().  The
     # user might call Configure() again to override options, but we don't want
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -321,8 +321,8 @@ class PhaseInfo(mutablerecords.Record(
 
   def __call__(self, phase_data):
     kwargs = dict(self.extra_kwargs)
-    kwargs.update({plug.name: phase_data.plug_map[plug.cls]
-                   for plug in self.plugs if plug.update_kwargs})
+    kwargs.update(phase_data.plug_manager.ProvidePlugs(
+        (plug.name, plug.cls) for plug in self.plugs if plug.update_kwargs))
     arg_info = inspect.getargspec(self.func)
     if len(arg_info.args) == len(kwargs) and not arg_info.varargs:
       # Underlying function has no room for phase_data as an arg. If it expects
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -82,7 +82,7 @@ class Test(object):
     self._test_options = TestOptions()
     self._test_data = TestData(phases, metadata=metadata, code_info=code_info)
     self._lock = threading.Lock()
-    self._executor = exe.TestExecutor(self)
+    self._executor = exe.TestExecutor(self._test_data)
     self._stopped = False
     # Make sure Configure() gets called at least once before Execute().  The
     # user might call Configure() again to override options, but we don't want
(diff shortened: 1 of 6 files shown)


openhtf/__init__.py
```diff
@@ -103,8 +103,13 @@ def AddOutputCallbacks(self, *callbacks):
     """Add the given function as an output module to this test."""
     self._test_options.output_callbacks.extend(callbacks)
 
-  def OutputTestRecord(self, record):
+  def OutputTestRecord(self):
     """Feed the record of this test to all output modules."""
+    test_state = self._executor.GetState()
+    if not test_state:
+      return
```
grybmadsci: this code path probably warrants a log message

openhtf/__init__.py
```diff
@@ -103,8 +103,13 @@ def AddOutputCallbacks(self, *callbacks):
...
+
+    record = test_state.GetFinishedRecord()
     for output_cb in self._test_options.output_callbacks:
       output_cb(record)
```
grybmadsci: while you're here, can you add a `try:` block around each callback so exceptions in output callbacks don't cause this thread to die?

openhtf/exe/phase_data.py
```diff
@@ -72,9 +72,7 @@ def _asdict(self):
     """Return a dict of this PhaseData's public data."""
     return {'measurements': self.measurements,
             'attachments': self.attachments.keys(),
-            'plugs': {
-                k.__module__ + '.' + k.__name__: str(v)
-                for k, v in self.plug_map.iteritems()}}
+            'plug_manager': self.plug_manager}
```
grybmadsci: hm, i forget why this even needed an `_asdict` in the first place.  if possible, we should get rid of this.  if it's needed, then I'm guessing it's somewhere that expects all base types, so i'd rather keep the `plugs` dict and access some helper method on `self.plug_manager` to pull the strings out that we used to generate directly on `self.plug_map`

openhtf/plugs/__init__.py
```diff
@@ -184,19 +185,29 @@ class PlugManager(object):
...
       for plug_type in plug_types:
-        self.plug_map[plug_type] = plug_type()
+        self._plug_map[plug_type] = plug_type()
     except Exception:  # pylint: disable=broad-except
       _LOG.exception('Exception insantiating plug type %s', plug_type)
       self.TearDownPlugs()
       raise
 
+  @property
+  def _asdict(self):
```
grybmadsci: ah, okay, i see what you did, you're relying on the recursive nature of `ConvertToBaseTypes`.. that's fine actually, but can you look into where that's actually called, and potentially just remove both `_asdict()` methods if they're not used?  they might be used for the http_api, I think...

openhtf/exe/__init__.py
```diff
@@ -67,14 +66,12 @@ class TestExecutor(threads.KillableThread):
                          ['CREATED', 'START_WAIT', 'INITIALIZING', 'EXECUTING',
                           'STOP_WAIT', 'FINISHING'])
 
-  def __init__(self, test):
+  def __init__(self, test_data, plug_manager):
     super(TestExecutor, self).__init__(name='TestExecutorThread')
 
-    self._test = test
+    self._test_data = test_data
+    self._plug_mgr = plug_manager
```
grybmadsci: can we call it `self._plug_manager` here to be consistent?

grybmadsci:
LGTM, pending minor nits

openhtf/__init__.py
fahhem: I changed to take the finished record again, I don't know why I took that out.

openhtf/__init__.py
fahhem: Right, good idea. Wasn't there a bug for this? I can't find it.

openhtf/exe/__init__.py
fahhem: Done

openhtf/plugs/__init__.py
fahhem: it's included in TestState->PhaseData. The templates refer to measurements inside of PhaseData, but not the plugs, so I could remove them for now. Though I'm pretty sure Joe's frontend would use them

fahhem:
nits addressed

PR #242, opened 2016-05-03
add SaveAndRestore to conf

grybmadsci:
This intended to act similar to a FlagSaver but for conf values, and is primarily for unit tests.


fahhem:
sweet, lgtm!

Issue #226, opened 2016-04-12
Display print statements or logs in front-end


agdeepak:
Current front end shows Display Prompts for the user inputs and Phase statuses. Requesting for a feature to redirect some print or log statements from the console to the front end for user to see the test execution progress, status and/or exceptions thrown during test run.

jettisonjoe:
The new frontend will display any log entries that are captured in the test record.

As a side note, test scripts generally shouldn't use print statements except for debug purposes during development.

jettisonjoe:
Logs are now displayed in the web frontend via #246. Closing.

PR #246, opened 2016-05-12
Overhaul the included web frontend.

jettisonjoe:
This change is a total re-write of the built-in web frontend using Angular2,
Typescript, npm, and webpack.


fahhem:
Holy frontend, batman! I'll try to do at least a partial review tonight, but 77 files might take a while :)

CONTRIBUTING.md
```diff
@@ -36,7 +34,52 @@ To set up an OpenHTF dev environment, follow the steps below.
...
 3. `sudo pip install virtualenv`
-4. `(cd to openhtf directory)`
+4. (cd to openhtf directory)
 5. `virtualenv venv`
 6. `. venv/bin/activate`
 7. `python setup.py develop`
+
+
+## Web Frontend Development
+OpenHTF ships with a built-in web gui found in the `openhtf.io.frontend` module.
```
fahhem: Mention that if they don't want to develop, the pip install action above will include a working frontend.

fahhem: Ah-ha, it's mentioned, but down below. I'd put it up front since that's the expected normal/majority case

openhtf/io/frontend/__main__.py
```diff
@@ -100,12 +90,14 @@ def Refresh(self):
                        self.hostport)
         self.station_id = station_id
         self.state = state
+        result = True
```
fahhem: just return True here, since it's the only path that sets it to true, and put a return False at the bottom

openhtf/io/frontend/__main__.py
```diff
@@ -225,40 +214,65 @@ def post(self, host, port, prompt_id):
     msg = json.JSONEncoder().encode(
         {'id': prompt_id, 'response': self.request.body})
     self.store[host, port].Notify(msg)
-    self.write('SENT')
+    self.write('')
```
fahhem: does a write even have to happen? wouldn't this not do anything?

openhtf/io/frontend/__main__.py
```diff
@@ -225,40 +214,65 @@ def post(self, host, port, prompt_id):
...
 
 
 def main(argv):
   """Start the frontend."""
-  try:
-    argv = FLAGS(argv)  # parse flags
-  except gflags.FlagsError, e:
-    print('%s\nUsage: %s ARGS\n%s' % (e, sys.argv[0], FLAGS))
-    sys.exit(1)
+  parser = argparse.ArgumentParser(description='OpenHTF web frontend server.',
```
fahhem: Yeah argparse!

openhtf/io/frontend/__main__.py
```diff
@@ -225,40 +214,65 @@ def post(self, host, port, prompt_id):
...
+                      default=multicast.DEFAULT_TTL,
+                      help='TTL for station discovery pings.')
+  parser.add_argument('--dev', action='store_true',
+                      help='Start in development mode.')
+  args = parser.parse_args()
 
   logs.setup_logger()
 
-  store = StationStore()
+  path = BUILD_PATH if os.path.exists(BUILD_PATH) else PREBUILT_PATH
```
fahhem: Should this be based on --dev mode?

openhtf/io/frontend/prebuilt/index.html
```diff
@@ -0,0 +1,33 @@
...
+    <link rel="stylesheet" href="styles.css">
+    <script src="scripts/vendor.bundle.js"></script>
+    <script defer src="scripts/app.bundle.js"></script>
+    <script>
+      var FRONTEND_SERVER = '{{host}}:{{port}}';
+    </script>
+  </head>
+  <base href="/">
+
+  <!-- 3. Display the application -->
```
fahhem: 3?

jettisonjoe:
Dang, @fahhem, thanks for the super fast start on reviewing this. A lot of the 77 files are gonna be in the prebuilt and won't be very review-able so hopefully it's not as bad as it looks. ...Crap I forgot to update the prebuilt before I sent this PR. I'll do that when I address the first round of feedback.

openhtf/io/frontend/src/app/index.html
```diff
@@ -0,0 +1,33 @@
...
+    <link rel="stylesheet" href="/styles.css">
+    <script src="/scripts/vendor.bundle.js"></script>
+    <script defer src="/scripts/app.bundle.js"></script>
+    <script>
+      var FRONTEND_SERVER = '{{host}}:{{port}}';
+    </script>
+  </head>
+  <base href="/">
+
+  <!-- 3. Display the application -->
```
fahhem: Ah, this is the non-prebuilt version. Remove this comment. It would be nice if these pages commented on what their purpose is, since I can't know without either reading/knowing the entire frontend or running it.

openhtf/io/frontend/src/app/station.header.ts
```diff
@@ -0,0 +1,34 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+import {Component, Input} from 'angular2/core';
+import {NumberPipe}       from 'angular2/common';
+
+import {Countdown} from './utils';
+
+import 'file?name=/templates/station.header.html!./station.header.html';
```
fahhem: Why are these imports so funky looking?

openhtf/io/frontend/src/app/station.prompt.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /**
+   * Send the response to the given prompt through the StationService.
+   */
+  sendResponse(response, id) {
+    let ip = this.routeParams.get('ip');
+    let port = this.routeParams.get('port');
+    this.stationService.respondToPrompt(ip, port, id, response);
+    $('#response_field').val('');
```
fahhem: Why suddenly leave angular2 space here? This is the controller for this template after all

openhtf/io/frontend/src/app/station.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.framework = state.framework;
+      if (state.test) { 
+        this.test.update(state.test.record);
+
+        // Update the running phase.
+        let n = state.test.record.phases.length;
+        this.test.updatePhase(n, state.test.running_phase_record);
+
+        // Update pending phases using dummy phase records.
+        for (var i = 1; i < state.test.pending_phases.length; i++) {
```
fahhem: Why updated all future phases every time?

openhtf/io/frontend/src/app/test.ts
```diff
@@ -0,0 +1,191 @@
...
+   * @param phaseList - An option list of phases to override the record.
+   */
+  update(record: any) {
+    // Test state is tricky to track and update particularly because we want
+    // log entries and other data displayed in the gui to be selectable. If
+    // new components are being created on every update, we will lost active
+    // selections even when the gui appears to remain static. Instead, we need
+    // to keep the gui _actually_ static when no changes are happening, which
+    // means the inputs to view componts need to stay pointing at the same
+    // objects in memory. If someone has a better way to do this, please let
```
fahhem: Here's one option that could even lower the bandwidth of this UI:

The server currently sends the whole test state each time. If it stored the previous N states with a monotonically increasing number, the frontend would ask for an update and pass it it's current state number. The server would return a json patch object (http://jsonpatch.com/) that would be able to convert state #X to #Y, assuming #X was in the server's history. If not, it would return the state as-is and the UI would replace everything, but that would be a rare case of slow-bandwidth and the UI has probably changes significantly in the meantime.

This would 'solve' the problem by taking advantage of json patches not replacing objects, only adding/replacing/removing/etc, which is what all of this code does. It also reduces the number of times the test state organization is duplicated (by the framework, the typescript, and the HTML at the moment) since this doesn't care about the organization anymore.

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -0,0 +1,206 @@
...
+   * @param {string} color - materialize-css color classes for a status.
+   * @return {string} Darker versions of the given color classes.
+   */
+  transform(color: string) : string {
+    return this.darkStatusMap[color];
+  }
+}
+
+
+/** A Pipe for darkening status colors even more. **/
```
fahhem: Feels like you might be able to DRY this out by creating a MapPipeTransform base-class and having all of these just set the map class variable:

```ts

export class MappingPipeTransform implements PipeTransform {
  transformMap = {};

  /**
   * @param {string} input - Input data to be mapped.
   * @return {string} Output
   */
  transform(input: string) : string {
    return this.transformMap[input];
  }
}

/** Pipe to darken status colors **/
@Pipe({name: 'colorToDark'})
export class ColorToDark implements MappingPipeTransform {
  transformMap = {
    ...
  };
}
```

openhtf/io/frontend/src/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "stylus": "~0.54.2",
+    "path": "~0.12.7",
+    "script-loader": "~0.7.0"
+  },
+  "dependencies": {
+    "angular2": "^2.0.0-beta.13",
+    "es6-promise": "^3.0.2",
+    "es6-shim": "^0.35.0",
+    "reflect-metadata": "0.1.2",
+    "rxjs": "5.0.0-beta.6",
```
fahhem: react?

openhtf/io/frontend/__main__.py
```diff
@@ -52,33 +50,14 @@
...
-gflags.DEFINE_integer('discovery_ttl',
-                      multicast.DEFAULT_TTL,
-                      'TTL for station discovery pings.')
-
-
 _LOG = logging.getLogger(__name__)
 
 UNKNOWN_STATION_ID = 'UNKNOWN_STATION'
+BUILD_PATH = os.path.join(os.path.dirname(__file__), 'src', 'dist')
+PREBUILT_PATH = os.path.join(os.path.dirname(__file__), 'prebuilt')
```
fahhem: I think you have to use pkg_resources instead of this if you want to support wheel-based deployment, right?

fahhem:
I prioritize code reviews as much as I can. You got lucky that I was at my desk at the time!

CONTRIBUTING.md
```diff
@@ -36,7 +34,52 @@ To set up an OpenHTF dev environment, follow the steps below.
 1. `sudo apt-get install python-pip swig libssl-dev python-dev libffi-dev`
 2. `sudo pip install --upgrade pip`
 3. `sudo pip install virtualenv`
-4. `(cd to openhtf directory)`
+4. (cd to openhtf directory)
```
grybmadsci: While you're poking around here, I would actually make this a code block instead of each line separately, and do `# cd to openhtf directory`, making it much more copy-paste friendly.  as is it's kinda tedious.

openhtf/io/frontend/__main__.py
```diff
@@ -225,40 +214,65 @@ def post(self, host, port, prompt_id):
...
-  except gflags.FlagsError, e:
-    print('%s\nUsage: %s ARGS\n%s' % (e, sys.argv[0], FLAGS))
-    sys.exit(1)
+  parser = argparse.ArgumentParser(description='OpenHTF web frontend server.',
+                                   prog='python -m openhtf.io.frontend')
+  parser.add_argument('--port', type=int, default=12000,
+                      help='Port on which to serve the frontend.')
+  parser.add_argument('--poll_interval_ms', type=int, default=3000,
+                      help='Time between frontend polls in milliseconds.')
+  parser.add_argument('--discovery_interval_s', type=int, default=5,
```
grybmadsci: Hm, it's probably worth providing a means to disable multicast discovery, and also a way to explicitly register stations at invokation time, perhaps using `conf`?

openhtf/io/frontend/prebuilt/styles/station.header.css
```diff
@@ -0,0 +1,61 @@
+.station-header {
+  margin: 0px 24px 0px 24px;
+  overflow: hidden;
+}
+
```
grybmadsci: you have two blank lines here but only 1 in station.css.. i don't know what style is commonly accepted, but they should probably be consistent at least.

CONTRIBUTING.md
jettisonjoe: Good call.

openhtf/io/frontend/__main__.py
jettisonjoe: Fair.

openhtf/io/frontend/__main__.py
jettisonjoe: I thought I did that for a reason, but it _does_ seem to work without it, so I'm taking it out.

openhtf/io/frontend/__main__.py
jettisonjoe: I went back and forth a little on this, actually. If you use the flag, you might end up with nothing to serve, because there's nothing forcing you to actually build. On the other hand, if you _do_ build, but you want to serve the prebuilt for comparison or something, there's no simple way to do it. You either have to blow away what you built, or go in and change the path in code. I could go either way on this.

openhtf/io/frontend/prebuilt/index.html
jettisonjoe: lol, oops. Leftovers from Angular2 demo code.

openhtf/io/frontend/src/app/index.html
jettisonjoe: Fair enough. I'll add some comments.

openhtf/io/frontend/src/app/station.header.ts
jettisonjoe: The demo code had many style inconsistencies. I just picked one of the conventions arbitrarily and randomly extended it as the situation required. I'll see if I can dig up a ts style guide.

openhtf/io/frontend/src/app/station.prompt.ts
jettisonjoe: Gp.

openhtf/io/frontend/src/app/station.ts
jettisonjoe: Easier to do it every time than to have a separate, _different_ update at the beginning. I could put it behind a flag to `updateState()`. But since the state comes as one big package you need that whole package anyway just to update the pending phases. Could implement run-once style logic or something. Definitely open to suggestions here.

openhtf/io/frontend/src/app/test.ts
jettisonjoe: This sounds like exactly what I need. I spent a little time searching for something like this. Looks like I should have spent a little longer! I wasn't happy with the way the update logic ended up looking and would be _very_ happy to switch it out for something better like json patch.

openhtf/io/frontend/src/package.json
jettisonjoe: Oops. Was experimenting a bit. Meant to clean that up.

openhtf/io/frontend/__main__.py
jettisonjoe: Poooooosssssibly?

CONTRIBUTING.md
jettisonjoe: Gp. Will do.

openhtf/io/frontend/__main__.py
jettisonjoe: Oh yeah! I was thinking about that a while back, just forgot to implement it here. It would be really good to have. I'll add it.

openhtf/io/frontend/prebuilt/styles/station.header.css
jettisonjoe: Yeah, true. I can't even stick to my own conventions consistently. :-\

openhtf/io/frontend/src/app/station.header.ts
jettisonjoe: @fahhem I see now, you're asking about line 21 in particular, the template import. These 'file' loader imports are so webpack can put the files into directories that will be served by the static file handler, at the same time as allowing their sources to live side-by-side with the components that use them.

I have no idea what I'm doing when it comes to webpack; I'm learning as I go. So if there's a more proper way to do this I'm all ears!

openhtf/io/frontend/src/app/dashboard.service.ts
```diff
@@ -0,0 +1,126 @@
...
+  dashMap: {};
+  server: any;
+  polling_job: number;
+
+  /**
+   * Create a DashboardService.
+   */
+  constructor(private http: Http) {
+    this.listings = []; // Because ngFor only takes lists.
+    this.dashMap = {}; // Because we want persistant records.
```
grybmadsci: s/persistant/persistent

I'm so helpful, I know...

openhtf/io/frontend/src/app/utils.ts
jettisonjoe: On one hand I was sort of hoping to come up with something more robust that would also allow me to define the whole color palette in one spot (right now some colors are pulled from here, some are in html, and some are hard-coded in css). On the other hand, the MapPipe base class does clean this up quite a bit, so I think I'm just gonna go with the base class concept for now. Thanks!

openhtf/io/frontend/__main__.py
jettisonjoe: I added a glob to `package_data` in `setup.py` that should grab the whole prebuilt, so this logic _should_ work both when developing/running from source, and when installed via a wheel.

openhtf/io/frontend/src/app/test.ts
jettisonjoe: Instead of storing N states I'm thinking of storing a starting state and N patches (or N patches and a state at t-N if you prefer to think of it that way), and returning the series of patches from #X to #Y on a request. Thus trading CPU for memory on the server. Any holes you guys see with that approach?

jettisonjoe:
I addressed most of the smaller points in this latest commit. Now moving on to json patch, configurable station list, and multicast disable.

openhtf/io/frontend/src/app/test.ts
fahhem: At first, I was hesitant, but if you note that we can have tests with large attachments in the record and that we store them all in memory, then it makes sense. Every time you get a new state, would you generate the patch, throw out the new state, update your stored t-N state to t-N+1 and throw out the oldest patch, too? Seems like a lot of complexity?

The old system, while a tad over-engineered, essentially did this:

Frontend Server: Hold 'now' state with N patches for converting any state after now-N to now. When the state changes, generate a patch and send that to any connected websockets. Store the new state and patch, drop the old state and oldest patch.
UI: Hold 'now' state with timestamp (version number). Listen on websocket for patches. On reconnect, send request for 'next' state with timestamp of 'now', and receive either a list of patches or a brand-new state if the timestamp was from > now-N.

If we ignore the operational transform stuff (unnecessary when only 1 system can change the state at a time), then the system was pretty nice. Some tweaks had to be implemented to save bandwidth/latency, such as not sending attachments over the wire until requested, but I think those are relevant here as well, unfortunately.

jettisonjoe: > Frontend Server: Hold 'now' state with N patches for converting any state after now-N to now.

Since the N patches were now-relative, does that mean they all had to be updated (regenerated?) on a state update?

> Some tweaks had to be implemented to save bandwidth/latency, such as not sending attachments over the wire until requested, but I think those are relevant here as well, unfortunately.

Agreed, very likely we will need to do something smarter with attachments in the near future.

openhtf/io/frontend/src/app/dashboard.service.ts
jettisonjoe: Good eye. Eagerly awaiting the rest of your feedback. :stuck_out_tongue_winking_eye:

jettisonjoe: Assuming there _is going to be_ a rest of your feedback...

openhtf/io/frontend/src/app/test.ts
fahhem: > Since the N patches were now-relative, does that mean they all had to be updated (regenerated?) on a state update?

No, the N patches are relative to now-N. You can't generate the actual now-N state anymore, that's lost to free() calls. You can only update someone who does have now-N to now-N+i for every i in 1 to N.

jettisonjoe: Yeah, ok, that's what I thought. The way you worded the previous description confused me, though.

wallacbe:
add requests to setup.py

grybmadsci:
All right starting poking around some more, here's a screenshot:
https://gyazo.com/9956238b4ef042b1bc8327b36e7557a3

Some comments on this one:
  - It's unclear what "STOP_WAIT" is, that's an internal enum value that's really only confusing to the user.
  - The phase timings don't line up vertically.
  - While running the test, the current phase runtime was jumping left and right a lot when its precision changed (ie 1.234 seconds -> 2 seconds), if you want to provide millisecond precision there, i'd always print the trailing 0's.

jettisonjoe:
> It's unclear what "STOP_WAIT" is, that's an internal enum value that's really only confusing to the user.
The phase timings don't line up vertically.

I can see two paths toward making this more clear. (1) Transform these enum values to something more straightforward in the frontend. (2) Change the enum values and usage (and add more documentation) to more clearly reflect what's going on internally.

> While running the test, the current phase runtime was jumping left and right a lot when its precision changed (ie 1.234 seconds -> 2 seconds), if you want to provide millisecond precision there, i'd always print the trailing 0's.

I fixed that for the test header, but I guess I missed it on the phases. Will fix.

> We still don't have a solution for the 'test is done but we missed the end event' problem :(

Yeah, that's supposed to come with the new API changes. We could do hackz to get around it (like make the framework sleep for 1/2 second to make it _very_ likely the frontend will get a 1/4 second state update with a finished record, then detect that the record is finished and pause for several seconds in the frontend), but that seems sloppy when we should just crank out the API changes anyway. I'm open to better hackz if you can think of any, though.

> We need an official OpenHTF favicon ;)

Was `;)` your submission to the OpenHTF favicon design contest? You're right, though I hate seeing that favicon missing warning in the dev console.

wallacbe:
Can we show test_name in the station list and test_name, test_description and test version on the tester page?

Display overall pass/fail at the top

Move Submit button up to the same line as the text field to make more compact

jettisonjoe:
Yeah, I can look at putting those things into the station header.

PASS/FAIL is written to appear at the top when the test completes. The problem is that there's a timing issue causing us to miss test completion. @grybmadsci and I have some API refactoring planned that will help deal with that.

On the 'Submit' button position, I didn't want to cramp the input field, especially in smaller windows. But I could tweak the css to put these on the same line if the window is wide enough.

openhtf/io/frontend/src/app/station.ts
jettisonjoe: Update: This sloppiness is going away with the shopping out of client state updates to json-patch.

openhtf/io/frontend/src/app/test.ts
jettisonjoe: So in _this_ PR I'm going to implement json-patch usage in the client to make state updates a lot cleaner and cut out the whole record class paradigm. I'll make a new discussion-PR for potential paths to implementation of _server_ patch logic, because I realized that there are really two problems we're trying to solve with json-patch here. Client state, and network traffic.

jettisonjoe:
Fixes #277

jettisonjoe:
Fixes #226

jettisonjoe:
@wallacbe, on the issue of displaying metadata; that stuff is only available once we have a test record, meaning once test execution has started. If the station is between DUT's waiting for a start trigger, for instance, we don't have metadata. Realizing that led me to the thought that test metadata is testrun-specific rather than station-specific information, so might not belong in the dashboard view (list of stations).

What do you think of the idea of adding a third button (PHASES, LOGS, METADATA) in the test header for this instead of trying to squeeze the data itself into the header?

wallacbe:
What is currently displayed in the frontend, is it stationName or testType?  Is station name enough for the operator to identify the station, especially if it's only showing the host name?

jettisonjoe:
@wallacbe, currently there are three pieces of identifying information:

  * `station_id`
  * IP address
  * port

With `station_id` being the most prominent. If no `station_id` is specified in the config yaml, OpenHTF defaults to using the hostname as the `station_id`. So provided you're setting unambiguous `station_id`s through the config for each station you deploy, that _should_ (hopefully?) be sufficient for operators to identify which station they're looking at in the frontend.

At the same time, it's probably a good practice to provide operators with a link, shortcut, or bookmark directly to the station page for their station rather than making them search for it themselves in the dashboard.

wallacbe:
OK SGTM proceed.

grybmadsci:
right, OpenHTF has no concept of test type, the closest thing is `test_record.code_info.name`, which typically is the filename of the main entry point.  those metadata fields are specific to the mfg-inspector output module, but they are nothing more than arbitrary key/value pairs as far as OpenHTF is concerned.

Some thoughts:
  - It's probably worth being able to see the test metadata in the frontend once the test is started, so a metadata tab isn't a bad idea.
  - For the summary view, I think it's best to keep it as simple as possible, worst case we force test authors to put a little more thought into `station_id`.
  - I don't recommend rolling it into this PR, but at some point the ability to view the codeinfo would be pretty slick, particularly if we wanted to get super fancy and plumb information through about what line execution is at.  For now, as a catch-all, a debug pane that views the raw test record would be handy (probably useful for your debugging too @jettisonjoe )

jettisonjoe:
> I don't recommend rolling it into this PR, but at some point the ability to view the codeinfo would be pretty slick, particularly if we wanted to get super fancy and plumb information through about what line execution is at. For now, as a catch-all, a debug pane that views the raw test record would be handy (probably useful for your debugging too @jettisonjoe )

_Definitely_ want to do something like this eventually.

jettisonjoe:
Addressed reviewer feedback. Ready for another round of review.

openhtf/exe/phase_executor.py
```diff
@@ -179,16 +184,12 @@ def _ExecuteOnePhase(self, phase):
...
-        self.test_state.pending_phases.pop(0)
       return
 
     _LOG.info('Executing phase %s', phase.name)
 
-    phase_record = test_record.PhaseRecord(phase.name, phase.code_info)
-    self.test_state.running_phase_record = phase_record
-
     with phase_data.RecordPhaseTiming(phase, phase_record):
+        phase, self.test_state) as outcome_wrapper:
```
fahhem: Am I looking at a broken diff?

openhtf/io/frontend/__main__.py
```diff
@@ -118,19 +132,34 @@ def Notify(self, message):
...
-  def __init__(self):
+  def __init__(self, discovery_address, discovery_port, discovery_ttl,
+               discovery_interval_s, disable_discovery):
     super(StationStore, self).__init__()
+    self.discovery_address = discovery_address
+    self.discovery_port = discovery_port
+    self.discovery_ttl = discovery_ttl
+    self.discovery_interval_s = discovery_interval_s
+    self.disable_discovery = disable_discovery
     self._stop_event = threading.Event()
```
fahhem: I'd either make them all _private or all public

openhtf/exe/phase_executor.py
jettisonjoe: Yeah, I screwed that up during my rebase. Fixing now.

openhtf/io/frontend/__main__.py
jettisonjoe: @grybmadsci, this is now included. PTAL at my solution.

openhtf/io/frontend/prebuilt/styles/station.header.css
jettisonjoe: Fixed.

openhtf/io/frontend/__main__.py
jettisonjoe: Confirmed the frontend runs after a wheel-based install.

openhtf/io/frontend/__main__.py

jettisonjoe:
> for now, as a catch-all, a debug pane that views the raw test record would be handy

I'd like to save this for a later iteration and put the effort in to make something nice. Accessing the .../raw/... URL's has been more than sufficient for my debugging, despite the no-indent formatting.

grybmadsci:
- I really like the prompt UI, except one detail, it's not clear that you're supposed to click on the prompt to go into "enter text" mode.  maybe a slightly greyed out "(click here)" or "(click here to enter text)" after the prompt text might be a simple enough fix.
- the 'config' key in meta doesn't display correctly (you get 'object Object').  'config' at this point is just a vanilla dict, and we should be able to display such dicts in the metadata.
- At first glance, I was going to comment that a 'Phases:' label would give some context, but i think the problem might be that the underline on the phases/logs/meta selector might not be quite salient enough.  perhaps a slightly lighter box around the currently selected one might help?
- As a corollary, the inserting of the prompt box seems a bit crowbarred in where it is.  not sure i have an action item/suggestion here, maybe smooth the animation a little more and visually attach it to the header rather than having it be a floating bubble between the header and phases/logs/meta pane?  By having it floating in between you lose some of the visual association between that underline selector and the content it reflects.
- Logs almost certainly need a log level selector, at the very least enable/disable debug level logs.
- Times don't need leading 0's, i think that's just clutter. the trailing 0's are different because they actually provide information about precision ("1.00" is not the same as "1") and also allow you to right-align and not have it jump all over the place.
- It's a bit of a nit, but the fully collapsed phases feel a bit cramped together, maybe just space them out a tiny bit more?
- the expanded phase view needs a little more context as to what it is you're looking at, it looks like it's currently got str()'d validator, set value (if any), and pass/fail/unset status.  notably absent is the actual measurement name (looks like it works correctly for the currently running phase, so i'm guessing this is a bug).  and some headers to indicate what the columns are.
- as a followup to that, the expanded phase view ultimately will need some improvements, but these are more feature requests to keep in mind for the future.  specifically, a measurement can have multiple validators (which i don't think renders nicely currently), multidimensional values might need some extra work put into how we visualize them, and it would be nice if you could jump to the logs and/or codeinfo relevant to a particular phase.

CONTRIBUTING.md
```diff
@@ -16,10 +16,8 @@ The OpenHTF codebase is set up to use pip/virtualenv to manage dependencies.
 Pip is the python package manager, while virtualenv is a tool to isolate python
```
jettisonjoe: This file had some draft changes, but PR #313 contains the latest and greatest text, and will override the stuff in this PR on rebase.

openhtf/exe/phase_executor.py
```diff
@@ -167,26 +166,32 @@ def ExecutePhases(self):
       PhaseOutcome instance that wraps the phase return value (or exception).
     """
     while self.test_state.pending_phases:
-      result = self._ExecuteOnePhase(self.test_state.pending_phases[0])
+      phase = self.test_state.pending_phases.pop(0)
+      result = self._ExecuteOnePhase(phase)
+      
+      repeats = 0
+      while result == openhtf.PhaseResult.REPEAT:
+        _LOG.debug('Repeat #%s of phase %s.' % (repeats, phase))
```
grybmadsci: prefer passing format args directly to debug over using % operator.

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+    'PENDING': 'blue-grey lighten-3 white-text',
+    'WAITING': 'blue-grey lighten-3 white-text',
+    'OFFLINE': 'blue-grey lighten-3 white-text',
+  };
+}
+
+
+/** A Pipe for darkening status colors. **/
+@Pipe({name: 'colorToDark'})
+export class ColorToDark extends MapPipe {
```
fahhem: this is only used after a statusToColor, same with ColorToSuperDark. I'd just make them statusToColor, statusToDark, statusToSuperDark. That'll eliminate an extra pipe and having the non-dark color repeated 3 times.

It may instead make sense to create CSS classes with the name of the status, such as status-running, status-online, etc with the color and white-text, then status-light, status-dark and status-super-dark having the lightens and darkens. Then there would only be one pipe 'statusToCss'

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+   */
+  constructor(seconds: number) {
+    this.lengthSeconds = seconds;
+    this.onDone = function() { };
+  }
+
+  /**
+   * Reset the countdown.
+   */
+  reset() {
```
fahhem: It seems reset is never used except right after `constructor` and before `start`. Maybe combine it into one of those, preferably `constructor` so that `millisRemaining` is always defined.

jettisonjoe:
>  - I really like the prompt UI, except one detail, it's not clear that you're supposed to click on the prompt to go into "enter text" mode. maybe a slightly greyed out "(click here)" or "(click here to enter text)" after the prompt text might be a simple enough fix.

Fixed. Had to get rid of part of the animation, but this is more clear.

>  - the 'config' key in meta doesn't display correctly (you get 'object Object'). 'config' at this point is just a vanilla dict, and we should be able to display such dicts in the metadata.

Fixed using a pipe and CSS hackz. There are other ways to do it, but this should work.

>  - At first glance, I was going to comment that a 'Phases:' label would give some context, but i think the problem might be that the underline on the phases/logs/meta selector might not be quite salient enough. perhaps a slightly lighter box around the currently selected one might help?

Made the underline thicker for better visibility.

>  - As a corollary, the inserting of the prompt box seems a bit crowbarred in where it is. not sure i have an action item/suggestion here, maybe smooth the animation a little more and visually attach it to the header rather than having it be a floating bubble between the header and phases/logs/meta pane? By having it floating in between you lose some of the visual association between that underline selector and the content it reflects.

Good insight, but I don't have a good solution yet. For now I'm going to move the prompt above the test header. The down side is that the test header will jump around when prompts come and go, but at least the test header's selector will be directly adjacent to the content it controls. Filed #326 for this.

>  - Logs almost certainly need a log level selector, at the very least enable/disable debug level logs.

I have log-level highlighting for now. I'd like to do actual selection in a later iteration since we have a lot of more important things to get done for PyPI release. Filed #325 for this.

>  - Times don't need leading 0's, i think that's just clutter. the trailing 0's are different because they actually provide information about precision ("1.00" is not the same as "1") and also allow you to right-align and not have it jump all over the place.

Reduced min integer digits from 2 to 1.

>  - It's a bit of a nit, but the fully collapsed phases feel a bit cramped together, maybe just space them out a tiny bit more?

I think this will actually be a _helpful_ thing when you have tests with a ton of phases. I want to implement auto-expansion of the current phase in a later iteration of the frontend, though. Filed #324 for that.

>  - the expanded phase view needs a little more context as to what it is you're looking at, it looks like it's currently got str()'d validator, set value (if any), and pass/fail/unset status. notably absent is the actual measurement name (looks like it works correctly for the currently running phase, so i'm guessing this is a bug). and some headers to indicate what the columns are.

Wow, took me staring at that forever to see that I spelled 'key' as 'keu' hehe. That's why the measurement names were missing. I also added some column headers.

>  - as a followup to that, the expanded phase view ultimately will need some improvements, but these are more feature requests to keep in mind for the future. specifically, a measurement can have multiple validators (which i don't think renders nicely currently), multidimensional values might need some extra work put into how we visualize them, and it would be nice if you could jump to the logs and/or codeinfo relevant to a particular phase.

Filed feature requests #321, #322, #323 for this stuff.

>  - the measurements for the currently running phase appear to default to the measurements for the first phase if it has no measurements itself.

I don't know how that's happening. Are you seeing this in one of the example scripts?

openhtf/exe/phase_executor.py
jettisonjoe: Oops, I often forget that.

openhtf/io/frontend/src/app/utils.ts
jettisonjoe: I'm taking the first suggestion for now. My instinct was towards the CSS way as well, but I wanted to make use of materialize-css's pre-defined colors. In reality, there's no reason we can't take their colors and make our own color scheme based on them, but I'll leave that for a later optimization.

openhtf/io/frontend/src/app/utils.ts
jettisonjoe: Currently true, but I expect to use `reset()` once we have the ability to reliably detect test run completion. I will call it from the constructor to make sure `millisRemaining` is always defined, though.

jettisonjoe:
Addressed all reviewer feedback and rebased. PTAL.

grybmadsci:
still seeing running phase with random measurements:
[Screenshot](https://gyazo.com/a6980b7a624a90e1fc97fe72778011c0)

Code:
```Python
@TestPhase(timeout_s=1200)
def delay(test):
  time.sleep(1000)
```

As you can see, this phase has no measurements at all, but in the UI, it has the measurements from the first phase.  (This is from `all_the_things.py`).

jettisonjoe:
Crap, I don't know why that's happening. I'll take another look.

jettisonjoe:
UPDATE: The measurement display bug doesn't only affect phases with no measurements. It _always_ affects the currently running phase, when the currently running phase switches from one phase to another. The first-currently-running-phase-to-be-displayed-since-page-reload will always have its measurements shown in place of the true currently running phase's measurements.

Reloading the browser will cause the real currently running phase's measurements to be properly displayed. I think this has something to do with the way Angular2 binds names in its templates, and I don't have a good idea for how to work around it.

I'm going to cross my fingers and update to the latest version of Angular2.

jettisonjoe:
No dice with the Angular2 version update. Maybe change detection isn't picking up deep changes in our json-patched station state?

jettisonjoe:
Okay, I think this is finally fixed. Much thanks to @fahhem for the solution. The pipe being used on measurements had to be marked as impure, otherwise [Angular2 tries to be perfomant](https://angular.io/docs/ts/latest/guide/pipes.html):

> Angular ignores changes within (composite) objects. It won't call a pure pipe if we change an input month, add to an input array, or update an input object property.

I missed the pipe as a potential source of the issue because I could force freshness with the pipe, but I was ignoring the fact that what I put into the pipe to force freshness was no longer a composite object.

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -0,0 +1,252 @@
...
+/** A pipe to recurisely stringify objects into html <ul> elements. **/
+@Pipe({name: 'objectToUl'})
+export class ObjectToUl implements PipeTransform {
+  ulify(object: any): string {
+    if (object == undefined || object == null) {
+      return '';
+    }
+    let result = '<ul>';
+    Object.keys(object).forEach(key => {
+      result += `<li><span class="key">${key}:&nbsp</span><span class="value">`;
```
grybmadsci: shouldn't this be `&nbsp;` (missing semicolon)?

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -0,0 +1,252 @@
...
+    keys.forEach(key => {
+      this.result.push({ 'key': key, 'value': input[key] });
+    });
+    return this.result;
+  }
+}
+
+
+/** A pipe to recurisely stringify objects into html <ul> elements. **/
+@Pipe({name: 'objectToUl'})
```
grybmadsci: this pipe may suffer from the same strange issue if it isn't marked as non-pure, since it also works with composite objects.

grybmadsci:
LGTM, pending last couple nits, merge away :)

jettisonjoe:
FIxed last few nits. I'd like to get a consensus on whether we prefer squashing PR's down when we merge them or not. Personally I don't think having all these "address yet more reviewer feedback" commits is adding much value. Any thoughts before I merge this? (In a little while I might just pick an approach and merge anyway to keep us moving towards our planned Thursday release).

PR #247, opened 2016-05-13
Pass WithArgs to monitor phases too

fahhem:
If the monitor func can accept it, we pass it in. That now allows a phase to parameterize it's own code, measurement names, and now even monitor based on those values.


grybmadsci:
Awesome, thanks, and thanks for adding the test while you're at it :)

PR #252, opened 2016-05-20
Allow non-float multidims in mfg_inspector output

fahhem:
In the output munging, a multi-dim measurement can be a string, and its validator is included in a structured format when possible.

### DIFF
--- a/openhtf/io/output/mfg_inspector.py
+++ b/openhtf/io/output/mfg_inspector.py
@@ -131,14 +131,21 @@ def _MangleMeasurement(name, value, measurement, mangled_parameters):
       mangled_name += '_'
     mangled_param = test_runs_pb2.TestParameter()
     mangled_param.name = mangled_name
-    mangled_param.numeric_value = float(current_value[-1])
-    if measurement.units:
-      mangled_param.unit_code = UOM_CODE_MAP[measurement.units.uom_code]
     mangled_param.description = (
         'Mangled parameter from measurement %s with dimensions %s' % (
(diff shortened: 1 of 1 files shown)


openhtf/io/output/mfg_inspector.py
```diff
@@ -131,14 +131,29 @@ def _MangleMeasurement(name, value, measurement, mangled_parameters):
...
+
+    value = current_value[-1]
+    if isinstance(value, numbers.Number):
+      mangled_param.numeric_value = float(value)
+    else:
+      mangled_param.text_value = str(value)
+    # Check for validators we know how to translate.
     for validator in measurement.validators:
-      mangled_param.description += '\nValidator: ' + str(validator)
+      if isinstance(validator, validators.InRange):
```
grybmadsci: You can't ever have InRange or RegexMatcher for these because they are dimensioned (and those validators don't make sense or work on dimensioned measurements).

fahhem: Ah, this change has been lying around since before the multi-dim measurement split. However, I do think they are useful: Let's say I monitor the current as I enable/disable things and I want to ensure it doesn't go over 20A. In any case, I'll remove it from this since it's not relevant at the moment.

fahhem:
Ready for review again

openhtf/io/output/mfg_inspector.py
grybmadsci: Agreed, a couple rudimentary validators for multidim measurements would be handy, but yeah not currently a code path that can happen, so we can add it back in later.

PR #253, opened 2016-05-20
Build and associate an attachment for multidims

fahhem:
Attachment format: {'outcome': str, 'name': str, 'dimensions': [{'uom_suffix': str, 'uom_code': str}, ...], 'value': [...]}

This might conflict with my other patch, ah well.


openhtf/io/output/mfg_inspector.py
```diff
@@ -196,7 +198,17 @@ def _ExtractParameters(record, testrun, used_parameter_names):
           else:
             testrun_param.description += '\nValidator: ' + str(validator)
       else:
-        _MangleMeasurement(name, value, measurement, mangled_parameters)
+        attachment = testrun.info_parameters.add()
+        attachment.name = 'multidim_%s' % name
```
grybmadsci: i believe attachment names may need to be unique.

openhtf/io/output/mfg_inspector.py
```diff
@@ -131,6 +132,7 @@ def _MangleMeasurement(name, value, measurement, mangled_parameters):
       mangled_name += '_'
     mangled_param = test_runs_pb2.TestParameter()
     mangled_param.name = mangled_name
+    mangled_param.associated_attachment = attachment.name
```
grybmadsci: if you only use the name, might as well just pass `attachment_name` instead of the whole attachment object.

openhtf/io/output/mfg_inspector.py
```diff
@@ -196,7 +198,17 @@ def _ExtractParameters(record, testrun, used_parameter_names):
...
-        _MangleMeasurement(name, value, measurement, mangled_parameters)
+        attachment = testrun.info_parameters.add()
+        attachment.name = 'multidim_%s' % name
+        dims = [{
+            'uom_suffix': d.uom_suffix.encode('utf8'), 'uom_code': d.uom_code}
+            for d in measurement.dimensions]
+        attachment.value_binary = json.dumps({
+            'outcome': str(testrun_param.status), 'name': name,
+            'dimensions': dims, 'value': value}, sort_keys=True)
+        attachment.type = test_runs_pb2.MULTIDIM_JSON
```
grybmadsci: hm, if this format is documented somewhere, can you link to it? or reference it somehow?  if it's not, can you put said documentation here?

openhtf/io/output/mfg_inspector.py
fahhem: From what I can tell, attachments aren't dealt with that way, but just in case I combined all the name uniqueness mangling into one function and called that here too

fahhem:
Ready for review again

openhtf/io/output/mfg_inspector.py
fahhem: It's documented internally, I can't really share internal code like that...

openhtf/io/output/mfg_inspector.py
```diff
@@ -196,7 +198,17 @@ def _ExtractParameters(record, testrun, used_parameter_names):
...
       else:
-        _MangleMeasurement(name, value, measurement, mangled_parameters)
+        attachment = testrun.info_parameters.add()
+        attachment.name = 'multidim_%s' % name
+        dims = [{
+            'uom_suffix': d.uom_suffix.encode('utf8'), 'uom_code': d.uom_code}
+            for d in measurement.dimensions]
+        attachment.value_binary = json.dumps({
+            'outcome': str(testrun_param.status), 'name': name,
+            'dimensions': dims, 'value': value}, sort_keys=True)
```
grybmadsci: can you `dumps()` a value directly like this, or do you need to encode it?

openhtf/io/output/mfg_inspector.py
fahhem: Done

openhtf/io/output/mfg_inspector.py
fahhem: `dumps()` takes an object and returns a `bytes`/`str` object. Is the worry that `value` will have strings that need to be encoded?

openhtf/io/output/mfg_inspector.py
grybmadsci: can you copy/paste the JSON format component here, at least?

fahhem: Done

jettisonjoe:
LGTM

PR #259, opened 2016-05-21
add logger for plug, update example

wallacbe:


openhtf/plugs/__init__.py
```diff
@@ -115,6 +117,9 @@ class InvalidPlugError(Exception):
 
 class BasePlug(object): # pylint: disable=too-few-public-methods
   """All plug types must subclass this type."""
+  
+  def __init__(self):
+    self.logger = logging.getLogger(logs.RECORD_LOGGER)
```
fahhem: if you put this at the class level, it should still work and won't require a `super()` call

grybmadsci: hm, as another option, if you do something like this, you get plug type info built into the logger name:
```
@property
def logger(self):
  return logging.getLogger('.'.join((logs.RECORD_LOGGER, 'plugs', type(self).__name__)))
```

wallacbe:
Made changes, PTAL

fahhem:
LGTM

PR #260, opened 2016-05-21
check for GetState before calling GetFinishedRecord, modify stdin flush to allow console output

wallacbe:


fahhem:
LGTM. `stdin.flush()` should work, I'm not sure why @grybmadsci chose `tcflush()` over it?

openhtf/io/user_input.py
```diff
@@ -157,7 +157,7 @@ def run(self):
         print self._message
 
         # Before reading, clear any lingering buffered terminal input.
-        termios.tcflush(sys.stdin, termios.TCIOFLUSH)
```
grybmadsci: This seems like an unrelated issue to the fix above, can you pull this out of this PR?

openhtf/__init__.py
```diff
@@ -178,8 +178,9 @@ def Execute(self, test_start=None, loop=None):
     try:
       self._executor.Wait()
     finally:
-      record = self._executor.GetState().GetFinishedRecord()
-      self.OutputTestRecord(record)
+      if self._executor.GetState():
```
grybmadsci: can you add a comment here that if the framework never transitions from 'INITIALIZING' to 'EXECUTING' then test state never gets set, and there's nothing to output because the test never actually started.

wallacbe:
Reverted stdin change, added additional comment.   PTAL

grybmadsci:
i didn't choose tcflush(), i believe @jettisonjoe was the original author of that, and may have more knowledge of the reasons for doing it that way.

jettisonjoe:
I did choose it, probably because I didn't know about `stdin.flush()`, but also possibly because I thought it wouldn't work on Windows (which also could be wrong). It wasn't a super well-thought-through decision in the first place so I'm not at all opposed to changing it.

Issue #189, opened 2016-03-24
flag a measurement as FAIL if the returned value is 'nan'


song7seas:
Can we flag a measurement as FAIL if the returned value is ‘nan’, if the measurement is assigned the pass/fail limits? Currently it’s marked as “PASS” as the example below. 

"measured_values":{"foo":"nan"},

"foo":{"units":{"uom_suffix":"dBmW","uom_code":"2N"},"outcome":"PASS","docstring":"blah","name":"foo","validators":["0 <= x <= 10"]},

grybmadsci:
That's definitely a bug, someone will take a look when we get some time :)

wallacbe:
Any thoughts on handling -inf and inf?

It will behave like a number within InRange and fail if InRange(1, 4).  However the case where InRange(1, None) inf will pass.  However inf is sometimes an indicator something went wrong, (like in FFTs).  Should we reject inf or treat it like a normal float?

grybmadsci:
I think inf and -inf should pass, it's up to the author to specify an upper
limit if there is one, perhaps using something like sys.maxint if you want
"really large but not infinite upper bound".

fahhem:
sys.float_info.max is the true max, so I think it's reasonable if we provide a shortcut:

`RealNumber = InRange(-sys.float_info.max, sys.float_info.max)`

jettisonjoe:
Maybe fixed by #264. Separate feature request for a `RealNumber` validator?

wallacbe:
Moved RealNumber/inf conversation into:
https://github.com/google/openhtf/issues/266

I verified that https://github.com/google/openhtf/pull/264 fixes this issue. should close

PR #264, opened 2016-05-22
check for nan in measurments

wallacbe:


openhtf/util/validators.py
```diff
@@ -87,6 +87,9 @@ def __init__(self, minimum=None, maximum=None):
       self.maximum = maximum
 
     def __call__(self, value):
+      # Check for nan
+      if value != value:
```
fahhem: `math.isnan(value)`

wallacbe:
Global import of math was causing "AttributeError: 'NoneType' object has no attribute 'isnan".
I can import it within the class.

fahhem:
Yeah, validators is one of those modules madsci did the replace trick. You can import the module at the top and pass it into `__init__`. It's a lot more changes, unfortunately.

openhtf/util/validators.py
```diff
@@ -87,6 +87,10 @@ def __init__(self, minimum=None, maximum=None):
       self.maximum = maximum
 
     def __call__(self, value):
+      # Check for nan
+      import math
```
grybmadsci: copy the usage of `numbers`, `import math` at the top, pass it to `Validators.__init__()`, stash it on self in that constructor, and reference it here as `self.math_module.isnan()`.  Sorry this is tricky, it's done to provide `__getattr__` behavior at the module level, and this is actually the officially suggested way to do this.

wallacbe:
This is tricky.  I passed math to class Validators  but math is needed in the nested class InRange.  I'm reading that the methods of a nested class cannot directly access the instance attributes of the outer class.  I have to pass the math module into InRange, but the constructor doesn't appear to be the correct place to pass it.

openhtf/util/validators.py
fahhem: He makes a good point, the only way to intercept this the way things are is to override `__getattribute__` in `Validators`. I think this is the best place to do the import, considering.

wallacbe:
In the original commit I used the following method to test for nan without math module.  Is this an option given the current issue?

a = 1.0
b = float('inf')
c = 0
d = float('nan')

print a != a
print b != b
print c != c
print d != d

Output:
False
False
False
True

wallacbe:
Made updates with a decorator.  Tested this and it worked PTAL thanks

grybmadsci:
Hm, this is still kinda messy.  The preferred way of doing what you're trying to do is done with the `RegexMatcher`.  Creating a helper function like that starts getting tedious, though; I think you may be able to stash `sys.modules` in `self._modules` in `Validators.__init__()` and then refer to all modules as dict elements therein (`self._modules['re']` or `self._modules['math']`).

wallacbe:
made changes, PTAL

openhtf/util/validators.py
```diff
@@ -147,5 +140,7 @@ def __ne__(self, other):
   def MatchesRegex(self, regex):
     return self.RegexMatcher(regex, self.re_module.compile(regex))
```
grybmadsci: need to switch this to use `self.modules` instead of `self.re_module`

openhtf/util/validators.py
```diff
@@ -3,63 +3,49 @@
 Additional validators may be registered by passing them to the Register()
 method.  They can then be accessed directly as attributes on the validators
 module, and will typically be a type, instances of which are callable:
-
```
grybmadsci: let's leave the spacing of this as-is.

grybmadsci:
couple minor comments but otherwise LGTM

openhtf/util/validators.py
wallacbe: Done, not sure what happened with the formatting.

Done

jettisonjoe:
Uh oh, travis is barfing. Looks like it has to do with the pickling stuff @grybmadsci added.

grybmadsci:
oh, yea validators have to be pickle'able =/ this is going to be a real nuisance.  in the interest of just getting this fix in, you can do the `import math` inline like you had it (feel free to either leave the `sys.modules` conversion in or revert it, the important thing is you can't stash a copy of it on the validator itself).

wallacbe:
math is imported inline, PTAL

jettisonjoe:
Might have fixed #189.

PR #265, opened 2016-05-22
add validators and output to names

wallacbe:
Add validators to names for easy access.  The shorthand version of measures requires importing util.

```
from openhtf.names import *
@measures('inline_kwargs', docstring='This measurement is declared inline!',
units=UOM['HERTZ'], validators=[validators.InRange(0, 10)])
```
becomes
```
from openhtf.names import *
@measures('inline_kwargs', docstring='This measurement is declared inline!',
units=UOM['HERTZ'], validators=[InRange(0, 10)])
```

### DIFF
--- a/openhtf/io/output/__init__.py
+++ b/openhtf/io/output/__init__.py
@@ -1,0 +1,25 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Openhtf output modules."""
+
+import openhtf.io.output.json_factory
+import openhtf.io.output.mfg_inspector
+
+# pylint: disable=invalid-name
+
+# Classes used for output
+OutputToJSON = json_factory.OutputToJSON
+UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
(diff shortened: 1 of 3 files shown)


grybmadsci:
I'd like to reconsider this.  This is a fairly big namespace pollution (importing * from names is already very questionable), and also breaks the fundamental way validators works with custom validators (using `__getattr__`).  I'd be okay with importing the name 'validators' automatically so you wouldn't have to have the `from openhtf.util import validators` line, but your code would still have to reference validators explicitly.

wallacbe:
made change, PTAL

wallacbe:
Added output.  Having output here will really help clean up tester code.

wallacbe:
PTAL

grybmadsci:
LGTM, but @jettisonjoe is the primary on names and namespaces so I'll leave it to him to merge this one in.

I believe `output` used to be in there, but was removed (because json/mfg-inspector got demoted into modules within the output package rather than being in output's `__init__.py`).  You might consider aliasing the top-level names from `json_factory.py` and `mfg_inspector.py` in `io/output/__init__.py` to make `output` a little more useful (as it is, you'd still have to do `output.json_factory.OutputTOJSON`).  Then you could do `output.OutputToJSON()` and `output.UpoadToMfgInspector()` directly.

openhtf/io/output/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Openhtf output modules."""
+
+import json_factory
+import mfg_inspector
```
jettisonjoe: Little functional difference here, but Google's style guide prefers us to use absolute imports for these (`openhtf.io.output.json_factory`) rather than using relative imports.

> Do not use relative names in imports. Even if the module is in the same package, use the full package name. This helps prevent unintentionally importing a package twice.

https://google.github.io/styleguide/pyguide.html#Imports

PR #278, opened 2016-05-25
add indent=2 to json in mfg_inspector

wallacbe:


wallacbe:
PTAL

jettisonjoe:
LGTM, but I want to get a quick LGTM from @grybmadsci or @fahhem before we merge, since I haven't been super involved in the mfg_inspector stuff.

wallacbe:
purpose is to homogenize with the examples and make it more human readable

fahhem:
You're probably going to need to update the tests, but lgtm from me

jettisonjoe:
Oddly enough, the tests are passing right now, and the modded lines are covered according to the log. I could dig into the stringification of protobufs, or I could just not question it...

grybmadsci:
LGTM, i think the tests are passing because i updated the test to exclude certain redundant stuff, like this JSON text.

PR #279, opened 2016-05-25
mfg_inspector: configurable error message and save pb on fail

wallacbe:


wallacbe:
PTAL

openhtf/io/output/mfg_inspector.py
```diff
@@ -381,11 +382,14 @@ def locked_get(self):
     def locked_put(self, credentials):
       self._credentials = credentials
   # pylint: enable=invalid-name,missing-docstring
-
-  def __init__(self, user, keydata, token_uri=TOKEN_URI):
+  
+  def __init__(self, user, keydata, token_uri=TOKEN_URI,
+               error_message=None, filename_pattern_fail=None):
```
jettisonjoe: `filename_pattern_fail` was pretty mysterious until I read the whole thing. Maybe there's a clearer way to say this that hints better at its use, like `fallback_filename_pattern`.

jettisonjoe: What would the use-case be for customizing the error message versus simply logging something like 'Upload failed; falling back to writing file to local filesystem' or the like?

openhtf/io/output/mfg_inspector.py
```diff
@@ -430,18 +434,31 @@ def UploadTestRun(testrun, destination, credentials=None):
...
+          user_agent='OpenHTF Guzzle Upload Client',
+          token_uri=self.token_uri)
+      credentials.set_store(self._MemStorage())
       self.UploadTestRun(testrun, self.DESTINATION_URL, credentials)
-    except UploadFailedError:
+    except:
       # For now, just log the exception.  Once output is a bit more robust,
       # we can propagate this up and handle it accordingly.
+      if self._error_message:
+        sys.stdout.write(''.join([self._error_message, '\n']))
```
jettisonjoe: If it's just one variable going into the string I would consider
```sys.stdout.write('%s\n' % self._error_message)```
instead of the empty string join here.

File: openhtf/io/output/mfg_inspector.py (hunk 1 above)
Comment (wallacbe) [inline openhtf/io/output/mfg_inspector.py]: The idea is being able to trigger an alert for the operators that something is wrong.  Would give the ability to add yellow/bold warning text.  I should add a default message.

Comment (wallacbe) [inline openhtf/io/output/mfg_inspector.py]: Changed to fallback_filename_pattern, added args section

File: openhtf/io/output/mfg_inspector.py (hunk 2 above)

File: openhtf/io/output/mfg_inspector.py [hunk 3]
```diff
@@ -382,10 +383,26 @@ def locked_put(self, credentials):
...
+               fallback_filename_pattern=None):
+    """UploadToMfgInspector constructor.
+
+    Args:
+      user: Google cloud service account for OAuth2Credentials.
+      keydata: Google cloud keydata for OAuth2Credentials.
+      token_uri: Google cloud token URI for OAuth2Credentials.
+      error_message: Error message to display when upload fails.
+      fallback_filename_pattern: A format string specifying the filename to write to,
+      will be formatted with the Test Record as a dictionary.  May also be a
```
fahhem: indent follow-on lines so they're easier to read

openhtf/io/output/mfg_inspector.py
```diff
@@ -430,18 +447,30 @@ def UploadTestRun(testrun, destination, credentials=None):
...
+      credentials = oauth2client.client.SignedJwtAssertionCredentials(
+          service_account_name=self.user,
+          private_key=self.keydata,
+          scope=self.SCOPE_CODE_URI,
+          user_agent='OpenHTF Guzzle Upload Client',
+          token_uri=self.token_uri)
+      credentials.set_store(self._MemStorage())
       self.UploadTestRun(testrun, self.DESTINATION_URL, credentials)
-    except UploadFailedError:
+    except:
```
fahhem: Use `except Exception` if you want it to catch all the errors, not a bare except.

openhtf/io/output/mfg_inspector.py
```diff
@@ -430,18 +447,30 @@ def UploadTestRun(testrun, destination, credentials=None):
...
       # For now, just log the exception.  Once output is a bit more robust,
       # we can propagate this up and handle it accordingly.
+      sys.stdout.write('%s\n' % self._error_message)
+      if self._fallback_filename_pattern:
```
fahhem: I propose making this a separate object:

* ` UploadToMfgInspector` shouldn't catch any errors, since now it can raise an error and it'll get logged at the top level.
* The new object (`UploadOrOutput` for instance) should create two internal objects, `UploadToMfgInspector` and `OutputToJSON`
* In the new object's `__call__`, it should try-except around calling the `Upload...` object and in the except call the `OutputToJSON` object

openhtf/io/output/mfg_inspector.py

openhtf/io/output/mfg_inspector.py

openhtf/io/output/mfg_inspector.py
wallacbe: Done, thanks that's a way cleaner implementation

wallacbe:
PTAL

openhtf/io/output/mfg_inspector.py
```diff
@@ -444,4 +453,31 @@ def __call__(self, test_record):  # pylint: disable=invalid-name
     except UploadFailedError:
```
fahhem: This shouldn't catch the exception anymore, the framework will log the exception for you. That will also allow you to catch it in `UploadOrOutput`

wallacbe: Done, made change

fahhem:
LGTM

openhtf/io/output/mfg_inspector.py
grybmadsci: thanks farz, i was going to make the same comment :)

openhtf/io/output/mfg_inspector.py
```diff
@@ -439,9 +448,31 @@ def __call__(self, test_record):  # pylint: disable=invalid-name
...
     try:
-      self.UploadTestRun(testrun, self.DESTINATION_URL, credentials)
-    except UploadFailedError:
-      # For now, just log the exception.  Once output is a bit more robust,
-      # we can propagate this up and handle it accordingly.
-      logging.exception('Upload to mfg-inspector failed!')
+      sys.stdout.write('Attempting to upload to Mfg Inspector...\n')
+      self._UploadToMfgInspector(test_record)
+    except Exception:
+      sys.stdout.write('%s\n' % self._error_message)
```
grybmadsci: i'd prefer to use `logging` than writing directly to `sys.stdout`, as `openhtf` is aiming to be primarily a library, and well-behaved libraries typically don't touch `stdout` directly.

openhtf/io/output/mfg_inspector.py
```diff
@@ -71,6 +72,7 @@
 # measurement.
 MAX_PARAMS_PER_MEASUREMENT = 100
 
+ERROR_MESSAGE = 'Upload to mfg-inspector failed!'
```
grybmadsci: i don't think this really needs to be declared up top or configurable, it can just be used inline directly.

openhtf/io/output/mfg_inspector.py
```diff
@@ -334,19 +336,26 @@ class OutputToTestRunProto(object):  # pylint: disable=too-few-public-methods
...
+
+  Returns:
+    filename of local file.
   """
 
   def __init__(self, filename_pattern):
     self.filename_pattern = filename_pattern
 
   def __call__(self, test_record):  # pylint: disable=invalid-name
+
```
grybmadsci: don't need this blank line

grybmadsci:
LGTM pending a few nits.  Ideally I'd like a better solution for a generic fallback mechanism for output callbacks, but this is a great short term fix for this particular one, thanks!

openhtf/io/output/mfg_inspector.py
wallacbe: Made changes. Looks like using warning level logs to stdout.

openhtf/io/output/mfg_inspector.py

openhtf/io/output/mfg_inspector.py

wallacbe:
Made changes PTAL.

openhtf/io/output/mfg_inspector.py
grybmadsci: by default, yes, but that's configurable by users.  the first message isn't really a warning, and so should be info, the error one could be warning or error.

PR #282, opened 2016-05-30
Phase Unittesting Utils

grybmadsci:
Still have a lot of work to do on this, but I wanted to get it out for review so people could see the general direction I'm going with it and make requests/comments earlier rather than later.

Please hold off on code-review style comments until it's more complete, I'm specifically looking for usage comments (documentation will be more fleshed out too, for now take a look at `test_test.py` for an example of how to use it.

Definite features to be implemented are ability to test `Test()` instances as well as single phases, and also some utility `assert*` methods on the `TestCase` base class as shortcuts for checking phase results and measurement outcomes and things like that (that would otherwise require some poking at internals of phase records or test records).


grybmadsci:
Specifically, @wallacbe can you take a look?  Thanks!

test/util/test_test.py
```diff
@@ -0,0 +1,56 @@
...
+  phase_data.measurements.fails = 20
+
+
+class TestTest(test.TestCase):
+
+  @test.patch_plugs(mock_plug='.'.join((MyPlug.__module__, MyPlug.__name__)))
+  def test_patch_plugs(self, mock_plug):
+    mock_plug.do_stuff.return_value = 0xBEEF
+
+    phase_record = yield test_phase
```
wallacbe: What does this step do?  Can this happen in the background?

test/util/test_test.py
```diff
@@ -0,0 +1,56 @@
...
+  phase_data.measurements.othr_measurement = 0xDEAD
+  phase_data.measurements.passes = 5
+  phase_data.measurements.fails = 20
+
+
+class TestTest(test.TestCase):
+
+  @test.patch_plugs(mock_plug='.'.join((MyPlug.__module__, MyPlug.__name__)))
+  def test_patch_plugs(self, mock_plug):
+    mock_plug.do_stuff.return_value = 0xBEEF
```
wallacbe: Couple of items:
Phase  outcome - check that a phase passes or fails or errors
Assert raises - check that the correct error was raised

grybmadsci: yep, these are on the list to do, will update the example with how to do them :)

test/util/test_test.py
grybmadsci: this actually executes the phase and gives you back the resulting `PhaseRecord`, not sure what you mean by 'in the background'?  if you did this in a thread then you'd have to wait until it finished before you could do any asserts anyway, so there wouldn't be much benefit to it.

i went with this paradigm because you need to specify which phase to execute somehow, and you might want to do set up before doing so, so including it in the decorator was weird, and you want to be able to do asserts after it as well, so returning it from the function won't work either.  This is set up to kind of mirror the `contextlib.contextmanager` paradigm of "the user needs to provide something in the middle".

test/util/test_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+@plugs.plug(my_plug=MyPlug)
+@measurements.measures('test_measurement', 'othr_measurement')
+@measurements.measures('passes', validators=[validators.InRange(1, 10)])
+@measurements.measures('fails', validators=[validators.InRange(1, 10)])
+def test_phase(phase_data, my_plug):
+  phase_data.logger.error('in phase_data %s', id(phase_data))
+  phase_data.logger.error('in measurements %s', id(phase_data.measurements))
+  phase_data.measurements.test_measurement = my_plug.do_stuff('stuff_args')
+  phase_data.measurements.othr_measurement = 0xDEAD
```
jettisonjoe: :laughing:

jettisonjoe:
I dig it.

wallacbe:
Can we include how to handle configs in the the example?   openhtf.conf.Load(foo='bar')
  
Also add cleanup for the configs to the class?
self.addCleanup(openhtf.conf.Reset)

grybmadsci:
PTAL, the only changes I have planned are docs updates (i need to exhaustively document the assert* methods still).  If there are any other features wanted, let me know :)

I'd rather not include an example of how to temporarily modify conf values here, as that's a conf module thing, but I did mention it in the module level docstring.  We could add a unittest in the examples directory at some point that would use it, but I'd rather do that in a followup PR.

grybmadsci:
What happened to coveralls?  Can has coverage report?

grybmadsci:
All right, docs updated, 100% unit test coverage, ready for full review.

openhtf/__init__.py
```diff
@@ -237,17 +237,13 @@ def CreateArgParser(add_help=False):
...
-  CONTINUE = 'PHASE_CONTINUE'
-  REPEAT = 'PHASE_REPEAT'
-  FAIL = 'PHASE_FAIL'
+# Result of a phase.
+#
+# These values can be returned by a test phase to control what the framework
+# does after the phase.  CONTINUE causes the framework to execute the next
+# phase, REPEAT causes the framework to execute that same phase again, and STOP
+# causes the framework to stop executing.
+PhaseResult = Enum('PhaseResult', ['CONTINUE', 'REPEAT', 'STOP'])
```
grybmadsci: heads up @fahhem @wallacbe ,this change may break some of your tests.

wallacbe:
I ran my existing unit tests with these changes, no breakage.

wallacbe:
How about main, do we need to import unittest or can we use util.test?

if __name__ == '__main__':
  unittest.main()

openhtf/util/test.py
```diff
@@ -0,0 +1,377 @@
...
+@conf.SaveAndRestore decorator (see docs in conf.py, examples below).
+
+A few isolated examples, also see test/util/test_test.py for some usage:
+
+  from openhtf import conf
+  from openhtf.util import test
+
+  import mytest  # Contains phases under test.
+
+  class PhasesTest(test.TestCase)
```
wallacbe: missing colon
class PhasesTest(test.TestCase):

wallacbe:
Can we add example of how to access test_rec in the unit test (for doing assertions)?

grybmadsci:
As a subclass of `TestCase`, you should still use `unittest.main()`.

There are examples in the `test_test.py` file of how to use the test record (it gets yielded back in the same way as the phase record), but I'll add an example or two to the docstring also, good call.

A notable missing feature right now is that `test_start` is a lambda that always provides a dummy `dut_id`, and there's no way currently to inject a different `test_start` when yielding a `Test()` object to be executed.  It's tough to come up with a clean way to do that, so I'd like to punt on it for now until it actually becomes needed.

wallacbe:
I have a unit test requirement of passing specific serials.
I have test phases that require a dut serial of a specific length and format.
Will I be able to override the lambda?
test.Execute(test_start=lambda: 'serialnumber')

grybmadsci:
well, that doesn't work because you don't call `Execute()` directly here.  Could you prepend a phase that sets the dut id manually?  so your first phase would just be something like:
```
def set_dut_id(test):
  test.test_record.dut_id = 'serialnumber'
```
just for purposes of the unittest.

openhtf/exe/phase_data.py
```diff
@@ -115,32 +108,41 @@ def AttachFromFile(self, filename, name=None, mimetype=None):
...
+    """Context manager for the execution of a single phase.
+
+    This method performs some pre-phase setup on self (for measurements), and
+    records the start and end time based on when the context is entered/exited.
+
+    Args:
+      phase: openhtf.PhaseInfo object that is to be executed.
+      running_phase_record: PhaseRecord object tracking the phase execution.
+
+    Yields:
```
jettisonjoe: No longer yields this.

openhtf/util/test.py
```diff
@@ -0,0 +1,387 @@
...
+          openhtf.PhaseInfo.WrapOrCopy(phase_or_test))
+    return phase_or_test, self.last_result
+
+
+def yields_phases(func):
+  """Alias for patch_plugs with no plugs patched."""
+  return patch_plugs()(func)
+
+
+def patch_plugs(**mock_plugs):
```
jettisonjoe: Maybe consider `mock_plugs(**mocked_plugs)`? I was a little mislead by the use of "patch" at first which to me suggested something more open-ended. But maybe that's just me.

openhtf/util/test.py
```diff
@@ -0,0 +1,387 @@
...
+    with phasedata.RecordPhaseTiming(phase, phase_record):
+      try:
+        phase_record.result = phase_executor.PhaseOutcome(phase(phasedata))
+      except Exception as exc:
+        logging.exception('Exception executing phase %s', phase.name)
+        phase_record.result = phase_executor.PhaseOutcome(exc)
+
+    return phase_record
+
+  def _handle_test(self, test):
```
jettisonjoe: These mini baby phase and test executors are so cute, hehe!

jettisonjoe:
I dig it. Like we talked about earlier, it would be nice to get some example unit tests for some of our example OpenHTF Tests. Picked a few nits, but overall LGTM.

wallacbe:
@grybmadsci  where would the following be prepended, to the tester or to the unit test?
```
def set_dut_id(test):
  test.test_record.dut_id = 'serialnumber'
```
My phase is taking the scanned serial number number (unique per device) and doing some some manipulation based off of it (checksum, etc...).  My unit test will need to populate a serial number into the test object and I also have some test cases where I check that the correct exception was raised (InvalidSerialNumber).

Also is assertTestOutcomeCode equivalent to assertRaises ?

wallacbe:
Joe recommended checking out "handle_test" method to do this.

jettisonjoe:
FROM MEETING:
Let's make sure we get an example for `assertTestOutcomeCode` in `examples/` because this will probably be a common use case and it requires a little bit of knowledge of how outcome codes work to use it.

wallacbe:
I tried passing Test object.  I get AssertionError: Measurement foo not set because it looks like set_dut_id doesn't have that measurement.
```
@measures('foo')
def Phase1(test):
  test.measurements.foo = True

def set_dut_id(test):
  test.test_record.dut_id = 'serialnumber'

class PhasesTest(test.TestCase):

  @test.yields_phases
  def testPhase1(self):
    phase_record = yield Test(set_dut_id, Phase1,)
    self.assertMeasured(phase_record, 'foo', True)
    self.assertMeasurementPass(phase_record, 'foo')
```

grybmadsci:
This is actually a bug in the PR, fixing it now.  The handling of measurement assertions on TestRecords is currently asserting that *all* phases must have a matching measurement rather than *any*.

That is the correct usage of serial number setting, though.  Will push a fix shortly.

grybmadsci:
All right, fix pushed, and unit test updated to catch that particular case.  You should be able to pull and give it a try, I'll hold off on merging until I hear from you that it fixed your problem.

wallacbe:
Tried it out, fixed that case thanks, LGTM

PR #284, opened 2016-05-31
phases can take phase_data in *args or **kwargs

fahhem:
```py
def Phase(test, **kwargs):
  pass

getargspec(Phase) == ArgSpec(args=('test',), varargs=None, keywords='kwargs', defaults=None)
```

Which means doing Phase.WithArgs(something=1) makes that phase no longer take the test object, whether it wanted it or not.

I have an inkling there's a case where this isn't a good idea, but I can't make it do the wrong thing at the moment.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -339,8 +339,11 @@ class PhaseInfo(mutablerecords.Record(
     kwargs.update(phase_data.plug_manager.ProvidePlugs(
         (plug.name, plug.cls) for plug in self.plugs if plug.update_kwargs))
     arg_info = inspect.getargspec(self.func)
-    if len(arg_info.args) == len(kwargs) and not arg_info.varargs:
-      # Underlying function has no room for phase_data as an arg. If it expects
-      # it but miscounted arguments, we'll get another error farther down.
-      return self.func(**kwargs)
-    return self.func(phase_data, **kwargs)
+    # Pass in phase_data if it takes *args, or **kwargs with at least 1
+    # positional, or more positional args than we have keyword args.
+    if arg_info.varargs or (arg_info.keywords and len(arg_info.args) >= 1) or (
+        len(arg_info.args) > len(kwargs)):
+      # Underyling function has room for phase_data as an arg. If it doesn't
+      # expect it but we miscounted args, we'll get another error farther down.
+      return self.func(phase_data, **kwargs)
+    return self.func(**kwargs)


grybmadsci:
Hm, I'm curious as to @jettisonjoe 's thoughts on this.  I had considered this behavior initially, but looked into it, and `self` is always passed positionally to objects' methods - you can actually call it anything you want.  So passing the `PhaseData` object in positionally was intentional to follow that pattern.  I'm a little concerned about the naming of the argument and enforcing that it must be 'test'.

grybmadsci:
Specifically, it opens up a class of failure modes where someone has something like this:
```
def foo_phase(foo, bar, test):
  pass
```
and they alter the code and accidentally end up with
```
def foo_phase(foo, bar):
  pass
```
and now the `PhaseData` is getting passed in as `foo` positionally, and that's a pretty big "uh wtf just happened".  I think if we were to do kwarg support, we would want to enforce that it's *always* the same keyword and never pass it positionally.

fahhem:
Ah, I misnamed the pr, it's still passed in as the first positional
argument. I was distracted at the time and didn't realize this, sorry.

The real change is that if a phase takes args or kwargs, extra_args won't
count against it and it'll pass in phase_data anyway. On second thought, I
think I figured out my inkling, which is that the phase has to take at
least one positional arg if it takes kwargs to get phase_data passed in.

fahhem:
I made changes, and the logic is better commented now.

openhtf/__init__.py
```diff
@@ -339,8 +339,12 @@ def __call__(self, phase_data):
     kwargs.update(phase_data.plug_manager.ProvidePlugs(
         (plug.name, plug.cls) for plug in self.plugs if plug.update_kwargs))
     arg_info = inspect.getargspec(self.func)
-    if len(arg_info.args) == len(kwargs) and not arg_info.varargs:
-      # Underlying function has no room for phase_data as an arg. If it expects
-      # it but miscounted arguments, we'll get another error farther down.
-      return self.func(**kwargs)
-    return self.func(phase_data, **kwargs)
+    arbitrary_args = arg_info.varargs or arg_info.keywords
```
grybmadsci: arbitrary_args doesn't look like it's actually used anywhere?

jettisonjoe:
This PR as it currently stands looks okay to me, aside from the unused variable mentioned by @grybmadsci above.

grybmadsci:
yep lgtm too

openhtf/__init__.py
fahhem: Whoops! Done

fahhem:
Rebased and cleaned up, PTAL

PR #285, opened 2016-05-31
add callback for uploading local pb files

wallacbe:


wallacbe:
PTAL

openhtf/io/output/mfg_inspector.py
```diff
@@ -389,10 +395,19 @@ def locked_put(self, credentials):
...
   # pylint: enable=invalid-name,missing-docstring
 
-  def __init__(self, user, keydata, token_uri=TOKEN_URI):
+  def __init__(self, user, keydata,
+               token_uri=TOKEN_URI, destination_url=DESTINATION_URL):
     self.user = user
     self.keydata = keydata
     self.token_uri = token_uri
+    self.destination_url = destination_url
+    self.credentials = oauth2client.client.SignedJwtAssertionCredentials(
```
fahhem: 👍

openhtf/io/output/mfg_inspector.py
```diff
@@ -475,3 +492,24 @@ def __call__(self, test_record):  # pylint: disable=invalid-name
...
+  """
+
+  def __init__(self, user, keydata, directory):
+    self._UploadToMfgInspector = UploadToMfgInspector(user, keydata)
+    self._directory = directory
+
+  def __call__(self, _):  # pylint: disable=invalid-name
+    filenames = glob.glob('%s*.pb' % self._directory)
+    for filename in filenames:
+      if self._UploadToMfgInspector.UploadTestRun(filename):
```
fahhem: do you want to try-except around this or only attempt to upload until the first one fails?

This can get really complicated, such as with things like delaying retries for each file, or delaying between attempts (circuit-breaker pattern-esque, so if you can't upload a file, stop trying for 5 minutes instead of retrying every time, such as if you have a quick 30-second test on a loop). I don't know if we want to provide those things or provide this and say anything more complicated should go outside until there's N+ users with it replicated? Maybe it should go in contrib/ like @jettisonjoe and I talked about before, so there's less expectation of "it should always work" and more of a "if it doesn't work, copy-paste it into your project and change it"?

grybmadsci:
Hm, you're not going to like this, but I'm going to push back pretty hard on this sort of thing.  This is trying to leverage the OpenHTF output callback mechanism to run arbitrary python code, in this case a generic uploader service.

A real uploader service is actually quite complex (as @fahhem has hinted at above), and has a lot of nuances around retry mechanics, fallback mechanisms, etc.  That's all way too much to include in OpenHTF, and that's ignoring the security aspect of someone randomly dropping malicious files in the directory you're walking.

A good rule of thumb for output callbacks is that if they have any other inputs besides the test record passed in, then it's too complex for an output callback.  In this case, you're walking an entire directory, that's a lot of out of band inputs.

That said, this is a common request, so it's something we might roll into a `contrib/` directory, as @fahhem suggests.

wallacbe:
ACK...porting over some existing behavior/implementation.  How about the refactoring in the latest push minus the looping mechanism?  Does this seem acceptable?

fahhem:
I'm in favor of the refactoring

grybmadsci:
Yeah, the refactor LGTM, makes the code a little more reusable too, thanks!

Issue #286, opened 2016-06-01
Tear Down isn't working


wallacbe:
I don't see "Running teardown" in the output

python all_the_things.py
Provide a DUT ID in order to start the test.
test
Hello World!
What's the widget type?
foo
Plug value: 1
Tearing down <ExamplePlug: 5>
Finishing test execution with outcome FAIL.

PR #287, opened 2016-06-01
call test executor only once

wallacbe:
https://github.com/google/openhtf/issues/286


grybmadsci:
LGTM, but we could also do with a closer look at all the places that a call to `.Stop()` can cause weird behavior (here, it can cause `self._executor` to get set to `None` before calling `self._executor.Wait()`, which would lead to an obscure error).

Issue #288, opened 2016-06-01
numeric values of zero aren't logged to codes in failureCodesList


wallacbe:
numeric values of zero aren't logged to codes in failureCodesList.  PR in progress..

PR #289, opened 2016-06-01
fix numeric 0 not added to fail code list

wallacbe:
https://github.com/google/openhtf/issues/288

### DIFF
--- a/openhtf/io/output/mfg_inspector.py
+++ b/openhtf/io/output/mfg_inspector.py
@@ -252,7 +252,7 @@ def _ExtractParameters(record, testrun, used_parameter_names):
         testrun_code = testrun.failure_codes.add()
         testrun_code.code = testrun_param.name
         if measurement.dimensions is None:
-          if testrun_param.numeric_value:
+          if isinstance(testrun_param.numeric_value, float):
             testrun_code.details = str(testrun_param.numeric_value)
           else:
             testrun_code.details = testrun_param.text_value


wallacbe:
PTAL

fahhem:
LGTM

openhtf/io/output/mfg_inspector.py
```diff
@@ -417,6 +417,7 @@ def UploadTestRun(testrun, destination, credentials=None):
       testrun: TestRun proto to upload.
       credentials: An OAuth2Credentials object to use for authenticated uploads.
     """
+
```
grybmadsci: remove extra newline here

openhtf/io/output/mfg_inspector.py
```diff
@@ -252,10 +252,10 @@ def _ExtractParameters(record, testrun, used_parameter_names):
         testrun_code = testrun.failure_codes.add()
         testrun_code.code = testrun_param.name
         if measurement.dimensions is None:
-          if testrun_param.numeric_value:
-            testrun_code.details = str(testrun_param.numeric_value)
-          else:
+          if testrun_param.text_value:
```
grybmadsci: this is now going to be falsey if text_value is an empty string, which will change the behavior of this to make details the string "None" when the text value is an empty string, which is kind of weird and probably not what we want.  It's probably better to change the `if testrun_param.numeric_value:` to `if testrun_param.numeric_value is not None:`.

wallacbe: I tried this initially.  For the string case numeric_value is still populated with 0 and then enters that block.

I'll change it to check for float instead since we cast above for numeric case

openhtf/io/output/mfg_inspector.py

wallacbe:
Made change, PTAL

Issue #291, opened 2016-06-02
SigInt isn't stopping tester configured for looping


wallacbe:
Add loop to all_the_things:
```
 while True:
    test.Execute(test_start=triggers.PromptForTestStart())
```

python all_the_things.py 
Provide a DUT ID in order to start the test.
test
Hello World!
What's the widget type?
^C2016-06-01 19:50:24,605 - ERROR - Received SIGINT.
2016-06-01 19:50:24,605 - ERROR - Stopping Test due to SIGINT
2016-06-01 19:50:24,606 - ERROR - Test state: <TestState: test@wallaceb-glaptop Running Phase: hello_world>
Tearing down <ExamplePlug: 0>
^C2016-06-01 19:50:42,645 - ERROR - Received SIGINT.
^C

wallacbe:
I added sys.exit() at the end of Stop() in openhtf/__init__.py and I see that SIGINT stops the tester and plugs tear down methods are called.

python all_the_things.py 
Provide a DUT ID in order to start the test.
test
Hello World!
What's the widget type?
^C2016-06-01 21:35:43,877 - ERROR - Received SIGINT.
2016-06-01 21:35:43,877 - ERROR - Stopping Test due to SIGINT
2016-06-01 21:35:43,878 - ERROR - Test state: <TestState: Running Phase: hello_world>
Tearing down <ExamplePlug: 0>

fahhem:
Ah, right, it's not http_api, it's the fact that you're doing `while True`. When `Test.Stop()` gets called (by hitting Ctrl-C, we set `self._stopped = True` then if you try to do `test.Execute(...)` we check `_stopped` and refuse to run the test. So you sit in that while loop forever.

I just checked, and if we raise `KeyboardInterrupt` in `Test.Stop()` then it works. I think we should add that so I'm sending a PR for that now.

PR #293, opened 2016-06-02
add sys.exit to Stop method in Test class

wallacbe:
https://github.com/google/openhtf/issues/291

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -146,6 +146,7 @@ class Test(object):
         _LOG.error('Test state: %s', self._executor.GetState())
         self._executor.Stop()
         self._executor = None
+      sys.exit()
 
   def Execute(self, test_start=None, loop=None):
     """Starts the framework and executes the given test.


wallacbe:
PTAL

fahhem:
`sys.exit()` will close the entire program, and should be avoided by libraries that aren't trying to wrap `sys`. This will make using `openhtf` as part of another program harder since `Stop` can kill the whole program rather than just `openhtf`.

The problem you're having is that the http api thread doesn't always die when it's supposed to, so you're stuck waiting for it to die (which it won't) when you hit Ctrl-C. @jettisonjoe said he has changes in the pipeline to fix this, so for now just add `test.Configure(http_api=None)`.

In the meantime, I might make the http api a daemon thread so if it's the last thread running, it won't hold the interpreter up.

fahhem:
I mean `http_port=None`

wallacbe:
II had that set already and have the issue:

```
  test.Configure(http_port=None, teardown_function=teardown)
  while True:
```

fahhem:
I replied in #291 and sent a new PR (#295) that should address this.

PR #295, opened 2016-06-02
Raise KeyboardInterrupt when catching SIGINT

### LINKED ISSUE #291: SigInt isn't stopping tester configured for looping

fahhem:
Fixes #291

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -146,7 +146,10 @@ class Test(object):
         _LOG.error('Test state: %s', self._executor.GetState())
         self._executor.Stop()
         self._executor = None
-    raise KeyboardInterrupt
+    # The default SIGINT handler does this. If we don't, then nobody above
+    # us is notified of the event. This will raise this exception in the main
+    # thread.
+    raise KeyboardInterrupt()
 
   def Execute(self, test_start=None, loop=None):
     """Starts the framework and executes the given test.
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -82,7 +82,6 @@ class Test(object):
     self._test_options = TestOptions()
     self._test_data = TestData(phases, metadata=metadata, code_info=code_info)
     self._lock = threading.Lock()
-    self._stopped = False
     self._executor = None
     # Make sure Configure() gets called at least once before Execute().  The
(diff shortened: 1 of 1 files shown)


wallacbe:
LGTM

openhtf/__init__.py
```diff
@@ -146,6 +146,7 @@ def Stop(self, *_):
         _LOG.error('Test state: %s', self._executor.GetState())
         self._executor.Stop()
         self._executor = None
+    raise KeyboardInterrupt
```
grybmadsci: nit: iirc, style prefers raising instances rather than the exception type like this?

similarly a nit, can you add a comment here as to why you need to raise this, the python docs are actually a bit cryptic, but it looks as though this signal handler gets called in the main thread, between other instructions to the global interpreter?

grybmadsci:
LGTM, pending nits

openhtf/__init__.py

fahhem:
I think with this change, we may want to remove the `_stopped` variable and allow someone to `Execute()` a test again if they caught the `KeyboardInterrupt`?

jettisonjoe:
@fahhem that's a good point. I'm cool with getting rid of `_stopped`. @grybmadsci?

Also, LGTM and will merge once we resolve the point about `_stopped`.

grybmadsci:
I'd like to strongly discourage reusing the test in that case, but that wasn't the primary function of the `_stopped` attribute, that was to avoid a race between executor instantiation and execution, which is no longer an issue with the refactor, so I'm okay with leaving `_stopped` out, but if you see someone catching a KeyboardInterrupt and re-running a test, you should give them a funny look to see if they might reconsider doing that.

PR #296, opened 2016-06-03
add phases and timings to test run proto

wallacbe:


fahhem:
LGTM

I think this is all mfg inspector shows at the moment, but we may later add support for the phase's description

grybmadsci:
LGTM, but you'll have to update the output test files.  Follow the instructions in the docstring to do that (change UPDATE_OUTPUT to True, run the test again, then change it back to FALSE).

grybmadsci:
Looking at the internal thread, i believe this mechanism in mfg-inspector is deprecated, so this isn't going to do what you want?

fahhem:
I actually just sent a CL internally to support this since it's the only way to upload timings in the future.

wallacbe:
I set UPDATE_OUTPUT = True, test is failing

fahhem:
You run it once with UPDATE_OUTPUT=True, then you set it back to False and run
it again. It should now be passing.





On Fri, Jun 3, 2016 5:51 PM, wallacbe notifications@github.com wrote:
I set UPDATE_OUTPUT = True, test is failing

wallacbe:
Hmmm, still failing.  I set to True, pushed, then set back to False and pushed again.

grybmadsci:
"Follow the instructions in the docstring to do that (change UPDATE_OUTPUT to True, run the test again, then change it back to FALSE)."

This is also described in the `output_test.py` module docstring, but more explicitly, you need to:
  - Change `UPDATE_OUTPUT` to `True`
  - Run the test locally by running `python setup.py test`, it will fail.
  - Change `UPDATE_OUTPUT` to `False`
  - Run the test locally again the same way, it should pass.
  - If it doesn't, then look at the output to see what needs to be fixed.
  - Once the test passes, commit the changes and push.

wallacbe:
test fixed, PTAL

PR #297, opened 2016-06-03
Add responsiveness tags to README.md

jettisonjoe:
Add average PR and issue close time tags (from issuestats.com) to our README.md file.


jettisonjoe:
@googlebot I'm an employee of Google, Inc. so I should be covered.

fahhem:
I think that means the email address you used in your commit isn't recognized?

jettisonjoe:
Yeah, it's probably because I used github's interface and it defaulted to my personal email (even though both are linked to my account). I was sort of hoping being a member of the 'Google' github organization would be taken into account, but I guess it's just email addy based.

PR #302, opened 2016-06-04
add plug logging and conf to all the things

wallacbe:


wallacbe:
PTAL

examples/all_the_things.py
```diff
@@ -32,9 +33,15 @@
...
 
 
+conf.Declare('foo', default_value='bar',
+             description='Example of conf value.')
+
+
 @plug(example=example_plug.ExamplePlug)
-def example_monitor(example):
-  return example.Increment()
+def example_monitor(self, example):
```
grybmadsci: don't call it 'self', 'self' very specifically is intended for class methods to refer to the bound instance.  i think we typically call it 'test', but really anything that's not a word that already means something else.

grybmadsci:
Hm, what is this intended to demonstrate?  All you're doing is accessing a conf value in a couple places.  It doesn't appear as though you're using a plug's logger.

wallacbe:
hmm, yeah that's incorrect let me fix

PR #306, opened 2016-06-07
Raise errors from uploading correctly

fahhem:
Previously the UploadFailedError would get caught and re-raised differently right away


jettisonjoe:
Hah, gp.

PR #307, opened 2016-06-07
Fix typo from earlier PR

fahhem:


jettisonjoe:
Dang, I probably read that sentence 5 times in the original PR and didn't notice.

Issue #311, opened 2016-06-08
have output callbacks use .format


wallacbe:
have output callbacks use .format to allow netsted dicts access (like meta data)

PR #312, opened 2016-06-08
use format in output callbacks

wallacbe:
https://github.com/google/openhtf/issues/311


wallacbe:
PTAL

fahhem:
I'd like to keep backwards compatibility if possible, so I think we should use the mechanism from measurements' WithArgs that checks for { in the string and chooses the appropriate interpolation based on that.

wallacbe:
Made change, PTAL

grybmadsci:
LGTM, will merge once there's a second LGTM in case I missed something.

fahhem:
LGTM, thanks Brandon!

PR #319, opened 2016-06-13
Encode unicode before shoving into a bytes field

fahhem:
protos expect `value_binary` to be Python `bytes` or `str` objects. If we pass in `unicode` for any reason, it's understandably expected to be encoded. UTF-8 is as good an encoding as any.


jettisonjoe:
LGTM.

Issue #318, opened 2016-06-13
Output file syncronization


wallacbe:
I made an uploader daemon that monitors files being output to a directory.  There may be a conflict if the daemon tries to access the output file before openhtf output module is done writing.  Should we add a locking mechanism or an atomic write to the output module or should this be handled by the daemon?

fahhem:
There's code in plumbum for atomic file operations:
http://plumbum.readthedocs.io/en/latest/api/fs.html

I don't think this has to be part of openhtf, but we can refactor the
output methods to allow a subclass to override the writing. I can see the
atomic writer going under `contrib` though.

grybmadsci:
Hm, using the plumbum code isn't too terrible of an idea.. a common solution to this problem is actually to write the output to a temp file first, and then os.rename it to the final location when writing is complete (because mv operations are atomic at the filesystem level).  there are a few edge cases where this isn't as cross platform (indeed, in python 3.3, os.replace was introduced to address these), but as long as the destination file doesn't exist, i believe it should work fine.

i think the plumbum code is neat, but rather overkill for this case, where we don't have contention over a shared file (where flock and friends are needed), but rather we want to make sure we write all our output to a single-use file before the next process in the pipeline slurps it up, and a simple os.rename functions for that purpose.

amyxchen:
I have a function "atomic_write" in the util, since we have multiple place to use it, it basically write the file to a tmp file, and rename it after it complete writing to be available to others. Not sure if it will fit this case.

grybmadsci:
That sounds like exactly what we need, could you send out a PR with just that function @amyxchen and @wallacbe can use it in his new output base class?

PR #320, opened 2016-06-14
add save file methods to output modules

wallacbe:
https://github.com/google/openhtf/issues/318

Allows subclass to overwrite the save_file method


wallacbe:
PTAL

openhtf/io/output/json_factory.py
```diff
@@ -51,12 +50,17 @@ def __call__(self, test_record):
...
-      with open(filename, 'w') as f:
-        f.write(self.encode(as_dict))
+        self.filename = self.filename_pattern % as_dict
+      self.file_data = self.encode(as_dict)
+      self.save_file()
     else:
       self.filename_pattern.write(self.encode(as_dict))
   # pylint: enable=invalid-name
+
+  def save_file(self):
```
grybmadsci: there's still a lot of code duplication here.  ideally, the output module (`io/output/__init__.py`) would contain an `OutputToFile` base class, where the `.format()` and file-writing functionality would reside, and then subclasses would implement the test-record-to-file-data mechanism.

in fact, test records are now pickle-able, so the default file output mechanism should be to simply pickle the test record and write that out (a very pure-python way to go about it), but subclasses can override that method to do other output formats (making those subclasses very small and clean).

wallacbe:
Made changes, PTAL

jettisonjoe:
Couple of things:

  * @amyxchen looks to have a good contextmanager-based realization of atomic file creation in the works as a contribution to `util`. Would be nice to use that in your OutputToFile base class.
  * `self.data` isn't guaranteed to exist on your base class, so subclasses have to provide it. I suggest either raising a `NotImplementedError` if you go to output and `self.data` hasn't been set, or setting `self.data` to a sensible default which subclasses can override, such as a pickled test record.
  * Should include more specific instructions in the docstring for how to subclass OutputToFile.

openhtf/io/output/json_factory.py
```diff
@@ -1,13 +1,13 @@
...
 import base64
+import output_to_file
 from json import JSONEncoder
-
 from openhtf.exe import test_state
 from openhtf.util import data
 
 
-class OutputToJSON(JSONEncoder):
+class OutputToJSON(JSONEncoder, output_to_file.OutputToFile):
```
grybmadsci: multiple inheritance can get quirky, better to avoid it when convenient.  I'd just create a JSONEncoder instance in `__init__` and save it on `self` for use in `__call__` (don't forget to pass **kwargs through to it's constructor).  This technique is referred to as 'composition' rather than inheritance: [more reading](https://www.thoughtworks.com/insights/blog/composition-vs-inheritance-how-choose)

openhtf/io/output/json_factory.py
```diff
@@ -44,19 +44,12 @@ def default(self, obj):
   def __call__(self, test_record):
     assert self.filename_pattern, 'filename_pattern required'
     if self.inline_attachments:
-      as_dict = data.ConvertToBaseTypes(test_record)
-      for phase in as_dict['phases']:
+      self.test_record_dict = data.ConvertToBaseTypes(test_record)
```
grybmadsci: generally avoid putting things on `self` when you don't need to, also, any attributes you plan to set on `self` should be initialized in `__init__` (this is a matter of style/convention, it's not technically required by Python).  In this case, you shouldn't save it on `self`.

fahhem:
@jettisonjoe and @grybmadsci, there are many atomic file creators. I wrote one, and it turns out so did half the open source community. On PyPI there are at least 5 different modules, not including 4 more on GitHub or in the Trac project:

* https://pypi.python.org/pypi/atomicwrites
* https://pypi.python.org/pypi/atomictempfile
* https://pypi.python.org/pypi/AtomicWrite
* https://github.com/mahmoud/boltons
* https://github.com/mitsuhiko/python-atomicfile
* https://www.edgewall.org/docs/tags-trac-0.11.7/epydoc/trac.util-pysrc.html
* https://github.com/abarnert/fatomic

Most of those have Windows support; some put the temp file in the same directory as the destination file to avoid cross-disk copies when you 'rename' it; some bring in more deps and some don't; but they've all been used more than one that we write.

I think the latest-and-greatest is 'atomicwrites', since it refers to most of the ones I listed here. It doesn't pull in any dependencies outside the standard library, and it's pretty simple to use. It even has tests that run on both Linux and Windows.

I strongly encourage openhtf to not reinvent this wheel.

openhtf/io/output/json_factory.py
```diff
@@ -44,19 +44,12 @@ def default(self, obj):
...
-      else:
-        filename = self.filename_pattern % as_dict
-      with open(filename, 'w') as f:
-        f.write(self.encode(as_dict))
-    else:
-      self.filename_pattern.write(self.encode(as_dict))
-  # pylint: enable=invalid-name
+      self.test_record_dict = data.ConvertToBaseTypes(test_record, ignore_keys='attachments')
+
+    self.data = self.encode(self.test_record_dict)
```
grybmadsci: stashing values on `self` and then invoking a method that accesses it via `self` is a common mistake (in many languages).  Generally, you want to limit data to as small a namespace as possible (`self` is pretty broad, compared to just the locals of a method).  Passing things in as arguments also keeps a "more functional" interface, where inputs and outputs are more explicit.

the better way to do what you're doing here would be to pass the data into `.Save()`.  Really, though, the even better way is to have the base class' `__call__` method invoke some other method to format the data, and have subclasses override that method.  This reduces code duplication even more (subclasses don't have to all call `Save()` explicitly) and has subclasses override exactly the behavior they want, and nothing more (ideally, you don't want subclasses to have to reimplement whole chunks of logic).

openhtf/io/output/output_to_file.py
```diff
@@ -0,0 +1,32 @@
...
+Output modules inherit from OutputToFile and overload methods
+based on output file requirements.
+
+"""
+
+import os
+import tempfile
+
+
+class OutputToFile(object):
```
grybmadsci: I'd recommend putting this in `io/output/__init__.py`, it's kind of a "base output functionality", not an extension thereof, so deserves to be a first-class citizen, I think.

openhtf/io/output/output_to_file.py
```diff
@@ -0,0 +1,32 @@
...
+        self.filename = self.filename_pattern.format(**self.test_record_dict)
+      else:
+        self.filename = self.filename_pattern % self.test_record_dict
+      self.SaveFile()
+    else:
+      self.filename_pattern.write(self.data)
+    return self.filename
+
+  def SaveFile(self):
+    with tempfile.NamedTemporaryFile(delete=False) as temp:
```
grybmadsci: See other comments, I think @amyxchen has a `util` function for you that will do this.

openhtf/io/output/output_to_file.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+import os
+import tempfile
+
+
+class OutputToFile(object):
+  """Class for saving to file."""
+
+  def Save(self):
```
grybmadsci: see above for the longer version, but basically, you should implement `__call__` here, have it invoke a separate method on self for formatting data (make the default just use pickle), and have subclasses override that method rather than having to implement `__call__` themselves.

you might also pull out the filename_pattern handling into a separate method, so that `__call__` is really just a handful of calls to other methods on `self`.

openhtf/io/output/output_to_file.py
```diff
@@ -0,0 +1,32 @@
...
+    """Save to file or file object."""
+    self.filename = None
```
grybmadsci: this class needs an `__init__`, and filename_pattern needs to be set there.  i'm guessing the multiple inheritance broke your ability to have an __init__ here, so once you fix that it should work like you expect.

base classes need to be fully functional self-contained classes, here you're assuming all sorts of behavior on the part of the subclasses (self.data and self.filename_pattern being set).  at most, a base class can leave some methods as "the subclass must implement this" (referred to as an abstract class), but making assumptions about attributes on `self` that subclasses are going to initialize/set for you makes for very fragile code.

grybmadsci:
Not bad for a first stab at it - this task is notably more complex than most of the stuff you've seen/done, brandon, so it's a great learning opportunity.  I've tried to give real justification, useful vocab words for looking up, and a link or two for context where it seemed worthwhile.  This PR is getting into some legit SWE stuff, so don't feel discouraged at the numerous review cycles and "no don't do it this way" comments :)

openhtf/io/output/output_to_file.py
jettisonjoe: Good point, @grybmadsci. This output module should be a first-class citizen.

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
-# Copyright 2014 Google Inc. All Rights Reserved.
```
grybmadsci: let's get the standard license header in here, with the date updated to 2016

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
...
+"""Output module."""
```
grybmadsci: """
This module contains support for various built-in output mechanisms.

Here, a base OutputToFile class is implemented to provide simple output to
a file via the pickle serialization mechanism.  It can be subclassed to implement
alternative serialization schemes, see json_factory.py and mfg_inspector.py for
examples.
"""

openhtf/io/output/mfg_inspector.py
```diff
@@ -37,7 +36,7 @@
 import threading
 import zlib
 
-from openhtf.io.output import json_factory
```
grybmadsci: you can still leave this as `from openhtf.io.output import json_factory`, `from X import Y` syntax is allowed as long as Y is a module (ie, style forbids importing individual symbols from a within a module that way).

openhtf/io/output/mfg_inspector.py
```diff
@@ -334,49 +325,32 @@ def _TestRunFromTestRecord(record):
...
       will be formatted with the Test Record as a dictionary.  May also be a
       file-like object to write directly to.
-
   Returns:
     filename of local file.
   """
 
   def __init__(self, filename_pattern):
-    self.filename_pattern = filename_pattern
+    super(self.__class__, self).__init__(filename_pattern)
```
grybmadsci: Hm, by convention we use `OutputToTestRunProto` instead of `self.__class__` here, not sure if there's any real technical justification for that.. this might act weird if someone inherits from your class, I think `self.__class__` would then be up one level and your `super()` would infinitely recurse.

openhtf/io/output/json_factory.py
```diff
@@ -30,18 +26,14 @@ class OutputToJSON(JSONEncoder):
...
-
-  # pylint: disable=invalid-name
-  def __call__(self, test_record):
+    self.json_encoder = JSONEncoder(**kwargs)
+  
+  def GetWriteData(self, test_record):
+    return self.json_encoder.encode(self.ConvertToDict(test_record))
+  
+  def ConvertToDict(self, test_record):
     assert self.filename_pattern, 'filename_pattern required'
```
grybmadsci: this assert should move to the base class

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
...
-# Classes used for output
-OutputToJSON = json_factory.OutputToJSON
-UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
-OutputToTestRunProto = mfg_inspector.OutputToTestRunProto
-UploadOrOutput = mfg_inspector.UploadOrOutput
+  def ConvertToDict(self, test_record):
+    return data.ConvertToBaseTypes(test_record)
+
+  def GetWriteData(self, test_record):
+    return str(self.ConvertToDict(test_record))
```
grybmadsci:
```Python
import cPickle as pickle

# ...
  def SerializeTestRecord(self, test_record):
    return pickle.dumps(test_record, -1)
```

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
...
+    if '{' in self.filename_pattern:
+      return self.filename_pattern.format(**test_record_dict)
+    return self.filename_pattern % test_record_dict
+
+  def Save(self, filename, write_data):
+    # TODO(wallacbe): use util atmoic method
+    with tempfile.NamedTemporaryFile(delete=False) as temp:
+      temp.write(write_data)
+      temp.flush()
+      os.rename(temp.name, filename)
```
grybmadsci: `os.rename` needs to be de-dented one level to ensure the temp file has been closed (i believe on windows you can't `os.rename()` an open file).  it's a moot point because you'll be switching to something else instead, but just an FYI.

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
...
+  def ConvertToDict(self, test_record):
+    return data.ConvertToBaseTypes(test_record)
+
+  def GetWriteData(self, test_record):
+    return str(self.ConvertToDict(test_record))
+
+  def Format(self, test_record_dict):
+    if '{' in self.filename_pattern:
+      return self.filename_pattern.format(**test_record_dict)
+    return self.filename_pattern % test_record_dict
```
grybmadsci: let's raise a `ValueError('%-style filename patterns deprecated, use .format() syntax')` in this case.

openhtf/io/output/__init__.py
```diff
@@ -1,26 +1,41 @@
...
-UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
-OutputToTestRunProto = mfg_inspector.OutputToTestRunProto
-UploadOrOutput = mfg_inspector.UploadOrOutput
+  def ConvertToDict(self, test_record):
+    return data.ConvertToBaseTypes(test_record)
+
+  def GetWriteData(self, test_record):
+    return str(self.ConvertToDict(test_record))
+
+  def Format(self, test_record_dict):
```
grybmadsci: rather than handling just the `.format()` here, I'd also handle the possibility of a callable, that way all the 'handle possible inputs of filename_pattern' logic is in the same place.  something like:
```Python
  def open_output_file(self, test_record):
    record_dict = data.ConvertToBaseTypes(test_record)
    if isinstance(self.filename_pattern, basestring):
      if '{' in self.filename_pattern:
        # Replace open() with atomic file open.
        return open(self.filename_pattern.format(**record_dict), 'wb')
      elif '%' in self.filename_pattern:
        raise ValueError('%-style filename patterns deprecated, use .format() syntax')
      return open(self.filename_pattern, 'wb')
    if callable(self.filename_pattern):
      return open(self.filename_pattern(record_dict), 'wb')
    if hasattr(self.filename_pattern, 'write'):
      return self.filename_pattern
    raise ValueError('filename_pattern must be string, callable, or File-like object')

  def __call__(self, test_record):
    with self.open_output_file(test_record) as outfile:
      outfile.write(self.serialize_test_record(test_record))
```

openhtf/io/output/__init__.py
wallacbe: Done

openhtf/io/output/__init__.py
wallacbe: Done

openhtf/io/output/mfg_inspector.py

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,54 @@
...
-import openhtf.io.output.json_factory
-import openhtf.io.output.mfg_inspector
+Here, a base OutputToFile class is implemented to provide simple output to
+a file via the pickle serialization mechanism. It can be subclassed to implement
+alternative serialization schemes, see json_factory.py and mfg_inspector.py for
+examples.
+"""
 
-# pylint: disable=invalid-name
+import base64
```
grybmadsci: clean up imports, there's a few i don't think you need anymore

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,54 @@
...
-# Classes used for output
-OutputToJSON = json_factory.OutputToJSON
-UploadToMfgInspector = mfg_inspector.UploadToMfgInspector
-OutputToTestRunProto = mfg_inspector.OutputToTestRunProto
-UploadOrOutput = mfg_inspector.UploadOrOutput
+
+class OutputToFile(object):
+  """Class for formatting filename and writing to file."""
+
+  def __init__(self, filename_pattern=None):
```
grybmadsci: Hm, I think the default value of None was to use OutputToJSON as a JSONEncoder without doing the actual file output, but that doesn't really make sense in this context, so I think it's safe to require the filename_pattern.

openhtf/io/output/json_factory.py
```diff
@@ -30,33 +26,19 @@ class OutputToJSON(JSONEncoder):
...
-
-  # pylint: disable=invalid-name
-  def __call__(self, test_record):
-    assert self.filename_pattern, 'filename_pattern required'
+    self.json_encoder = JSONEncoder(**kwargs)
+ 
+  def serialize_test_record(self, test_record):
+    return self.json_encoder.encode(self.ConvertToDict(test_record))
+  
+  def ConvertToDict(self, test_record):
```
grybmadsci: `def convert_to_dict()`, we're in a weird spot right now for convention, switching to underscores instead of caps, but let's make the whole file consistent at least (lowercase probably preferable, we're going to do a sweep soon and change over to all lowercase, i think).

openhtf/io/output/mfg_inspector.py
```diff
@@ -129,14 +128,13 @@ def _EnsureUniqueParameterName(name, used_parameter_names):
...
 def _AttachJson(record, testrun):
   """Attach a copy of the JSON-ified record as an info parameter.
-
   Save a copy of the JSON-ified record in an attachment so we can access
   un-mangled fields later if we want.  Remove attachments since those get
   copied over and can potentially be quite large.
   """
   record_dict = data.ConvertToBaseTypes(record, ignore_keys=('attachments',))
   record_json = json_factory.OutputToJSON(
-      sort_keys=True, indent=2).encode(record_dict)
```
grybmadsci: you can call `.serialize_test_record(record)` directly if you pass `inline_attachments=False` to the `OutputToJSON` constructor.  then you're not duplicating the `ConvertToBaseTypes()` call from `json_factory`.

openhtf/io/output/mfg_inspector.py
```diff
@@ -350,33 +340,20 @@ class OutputToTestRunProto(object):  # pylint: disable=too-few-public-methods
...
-      self.filename_pattern.write(serialized)
-    return filename
+  @staticmethod
+  def serialize_test_record(test_record):
+    return _TestRunFromTestRecord(test_record).SerializeToString()
 
 
 class UploadToMfgInspector(object):  # pylint: disable=too-few-public-methods
   """Generate a mfg-inspector TestRun proto and upload it.
-
```
grybmadsci: hm, most of the lines are here intentionally to give visual separation between the short-description and long-description.  can we leave them in?

openhtf/io/output/mfg_inspector.py
```diff
@@ -334,13 +326,11 @@ def _TestRunFromTestRecord(record):
   return testrun
 
 
-class OutputToTestRunProto(object):  # pylint: disable=too-few-public-methods
+class OutputToTestRunProto(output.OutputToFile):  # pylint: disable=too-few-public-methods
```
grybmadsci: I really like the simplification of the subclasses that this change enables, nice job with it!

openhtf/io/output/mfg_inspector.py
wallacbe: When I enter that code block I get an error, unsure yet if it's a bug..investigating

  as_dict = data.ConvertToBaseTypes(test_record, ignore_keys='attachments')
  File "/home/source/openhtf/openhtf/util/data.py", line 113, in ConvertToBaseTypes
    assert not isinstance(ignore_keys, basestring), 'Pass a real iterable!'
AssertionError: Pass a real iterable!

grybmadsci: oh, where you have `ignore_keys='attachments'`, you need to instead have `ignore_keys=('attachments',)`.. and i'm glad you hit that, because it makes me feel good about having he foresight to put that assertion there lol... had it not been there, you would have accidentally ignored the keys 'a', 't', 'c', 'h', 'm', 'e', 'n', and 's', because it would have interpreted 'attachments' as an iterable of individual string keys.

openhtf/io/output/json_factory.py
```diff
@@ -30,33 +26,19 @@ class OutputToJSON(JSONEncoder):
...
     if self.inline_attachments:
       as_dict = data.ConvertToBaseTypes(test_record)
       for phase in as_dict['phases']:
         for value in phase['attachments'].itervalues():
           value['data'] = base64.standard_b64encode(value['data'])
     else:
       as_dict = data.ConvertToBaseTypes(test_record, ignore_keys='attachments')
```
grybmadsci: here is where it needs to be `('attachments',)` instead... looks like it was a pre-existing bug that we just never caught.

wallacbe:
unit test failing.  cStringIO.StringO doesn't have an __exit__ attribute.  Doesn't work with with self.open_output_file(test_record) as outfile in __call__

grybmadsci:
all right, this gets annoying fast.  `cStringIO.StringIO` objects don't have context manager attributes (they pretend to be files in most ways, but not that particular way).  `io.StringIO` objects DO have context manager attributes, but once the object is closed, the value is no longer accessible, which is certainly not what you want if you pass a `StringIO` object in.

i'll make a comment inline as to what i think the best solution is, but sadly it's not quite as clean as before...

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,52 @@
...
+  def serialize_test_record(test_record):
+    return pickle.dumps(test_record, -1)
+
+  @staticmethod
+  def open_file(filename):
+    # Replace open() with atomic file open.
+    # This allows subclasses to override if they want.
+    return open(filename, 'wb')
+
+  def open_output_file(self, test_record):
```
grybmadsci:
```Python
@contextlib.contextmanager
def open_output_file(self, test_record):
    """Open file based on pattern."""
    output_file = None
    record_dict = data.ConvertToBaseTypes(test_record)
    if isinstance(self.filename_pattern, basestring):
      if '{' in self.filename_pattern:
        output_file = self.open_file(self.filename_pattern.format(**record_dict))
      elif '%' in self.filename_pattern:
        raise ValueError(
            '%-style filename patterns deprecated, use .format() syntax')
      else:
        output_file = self.open_file(self.filename_pattern)
    elif callable(self.filename_pattern):
      output_file = self.open_file(self.filename_pattern(record_dict))
    if output_file:
      try:
        yield output_file
      finally:
        output_file.close()
    elif hasattr(self.filename_pattern, 'write'):
      yield self.filename_pattern
    else:
      raise ValueError(
          'filename_pattern must be string, callable, or File-like object')
```

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,79 @@
...
+# TODO(wallacbe): Switch to util
+class Atomic(object):
+  """Class that does atmomic write in a contextual manner."""
+
+  def __init__(self, filename):
+    self.filename = filename
+    self.temp = tempfile.NamedTemporaryFile(delete=False)
+
+  def write(self, write_data):
+    self.temp.write(write_data)
```
grybmadsci: `return self.temp.write(write_data)`

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,79 @@
...
+
+
+class OutputToFile(object):
+  """Class for formatting filename and writing to file."""
+
+  def __init__(self, filename_pattern):
+    self.filename_pattern = filename_pattern
+
+  @staticmethod
+  def serialize_test_record(test_record):
```
grybmadsci: can you throw a quick docstring on these describing why a subclass might override it?  so for this one, something like "Override this to alter how test records are serialize to file data."

and `open_file` might have "Override this to alter how output files are opened or what type of file is used."

openhtf/io/output/__init__.py
```diff
@@ -12,15 +12,79 @@
...
+        yield output_file
+      finally:
+        output_file.close()
+    elif hasattr(self.filename_pattern, 'write'):
+      yield self.filename_pattern
+    else:
+      raise ValueError(
+          'filename_pattern must be string, callable, or File-like object')
+
+  def __call__(self, test_record):
```
grybmadsci: a docstring here describing something like:
```
"""Output the given TestRecord to a file.

Instances of this class are intended to be used as an output callback (see Test.AddOutputCallbacks) to output TestRecord results to a file.  This base implementation outputs the TestRecord by serializing it via the pickle module.  Subclasses may change this by overriding the serialize_test_record() method.  Additionally, subclasses may implement more complex file naming mechanisms by overriding the open_file() method.

Args:
  test_record: The TestRecord to write out to a file.
"""

File: test/io/output/output_test.py [hunk 29]
```diff
@@ -90,3 +90,6 @@ def testTestrun(self):
   def testUpdateOutput(self):
     """Make sure we don't accidentally leave UPDATE_OUTPUT True."""
     assert not self.UPDATE_OUTPUT, 'Change UPDATE_OUTPUT back to False!'
+
+if __name__ == '__main__':
```
grybmadsci: hm, i don't think we include this by convention, because we run the tests with pytest via `python setup.py test`.  It does make the test more standalone-runnable, though.  let's leave it out for now, and if you really like having it there, we can talk about adding this to all the unit tests.

grybmadsci:
LGTM pending a last few nits, @jettisonjoe can merge once we get a second LGTM (i'm going to be afk for a bit, I'll merge it later if needed, we'll definitely get this in today).

wallacbe:
Fixed nits, PTAL

grybmadsci:
missed one docstring comment, and we could use some unit tests for this, but i know you're trying to get this in asap, so let's merge it in and address the last couple nits later..

PR #327, opened 2016-06-15
Atomic write file

amyxchen:


openhtf/util/atomic_write.py
```diff
@@ -0,0 +1,40 @@
...
+
+@contextmanager
+def atomic_write(filename, filesync=False):
+  """ atomically write a file (using a temporary file).
+
+  filename: the file to be written
+  filesync: flush the file to disk
+  """
+
+  tmpfile = filename + '~'
```
grybmadsci: should use a `tempfile.NamedTemporaryFile` with `delete=False` for this to guarantee safe creation of the temp file (you could get collisions if you just append a string to the output filename).

grybmadsci:
The commits in the PR seem to create, then modify, then remove a bunch of unrelated files to the change, can you squash the commits into a single commit to remove that noise in the commit history?

PR #333, opened 2016-06-16
Allow unicode in data.ConvertToBaseTypes

fahhem:


openhtf/util/data.py
```diff
@@ -130,8 +131,10 @@ def ConvertToBaseTypes(obj, ignore_keys=tuple()):
     obj = [ConvertToBaseTypes(value, ignore_keys) for value in obj]
   elif isinstance(obj, tuple):
     obj = tuple(ConvertToBaseTypes(value, ignore_keys) for value in obj)
-  elif obj is not None and not isinstance(obj, numbers.Number):
-    # Leave None as None to distinguish it from "None", and leave numbers alone.
-    obj = str(obj)
+  elif obj is not None:
+    # Leave None as None to distinguish it from "None"
+    if not isinstance(obj, numbers.Number) and not isinstance(obj, basestring):
```
grybmadsci: why the second indentation level?  can't this all be a single conditional?

fahhem: Editing in github loses the 80-char line so I was flying blind. Fixed

PR #334, opened 2016-06-17
Programmatically generate unit codes.

### LINKED ISSUE #298: Switch to UNECE unit codes.
Update `openhtf/util/units.py` to use UNECE units.

jettisonjoe:
Here's a first attempt at generating UNECE codes from the published spec (a .xls file :-\). What do folks think of this approach?

Fixes #298


openhtf/util/units/__init__.py
```diff
@@ -0,0 +1,1170 @@
...
+import collections
+
+
+Unit = collections.namedtuple('Unit', 'name code suffix')
+
+UNITS = {}
+
+# NO_DIMENSION means that there are units set, but they cannot be expressed
+# by a known dimension (such as a ratio)
+UNITS['NO_DIMENSION'] = Unit('NDL', None)
```
fahhem: needs 3 arguments?

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+                'Common\nCode',
+                'Symbol']
+
+# The contents of the lines of the output file between which the generated code
+# will be inserted.
+START_DELIMITER = '######## BEGIN_UNITS ########'
+END_DELIMITER = '######## END_UNITS ########'
+
+SHEET_NAME = 'Annex I'
+REPLACEMENTS = {' ': '_',
```
fahhem: need to add ' to \' to make the tests get pasts that syntax error

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+# limitations under the License.
+
+
+"""Read in a .xls file and generate OpenHTF Unit objects.
+
+UNECE, the United Nations Economic Commision for Europe, publishes a set of
+unit codes for international trade in the form of Excel spreadsheets (.xls
+files) here:
+
+http://www.unece.org/cefact/codesfortrade/codes_index.html
```
fahhem: Mention the xls is in the zip file at the bottom under "Codes for Units of Measurement used in the International Trade". This wasn't obvious to me months ago when I got to that page

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+
+This tool is used to parse those spreadsheets and turn the published standard
+code set into a .py file that OpenHTF can import.
+
+Typical usage looks like:
+
+python ./bin/units_from_xls.py --infile ./Rec20_Rev6e_2009.xls \
```
fahhem: Why 2009? There's a 2014 one in the same zip file

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+  replacements = replacements or {}
+  result = string
+
+  for old, new in replacements.iteritems():
+    result = result.replace(old, new)
+
+  return result
+
+
+def insert_into_file(filepath, code, start, end):
```
fahhem: This start/end stuff seems easy to break. If you just put the top and bottom of the file as a triple-quoted string here, then you could just blow away the old file each time.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+      seen.add(name)
+
+      yield 'UNITS[\'%s\'] = Unit(\'%s\', \'%s\', \'%s\')\n' % (
+          multi_replace(name.upper(), REPLACEMENTS), name, code, suffix)
+
+  except xlrd.XLRDError:
+    sys.stdout.write('Unable to process the .xls file.')
+
+
+def multi_replace(string, replacements=None):
```
fahhem: This function seems useless if replacements isn't provided, why make it an optional argument?

fahhem:
I like the approach of generating it from the source file. It's just unfortunate that the source is an XLS :)

Did you have to clean any of the fields up or was that exactly the XLS that you downloaded?

openhtf/util/units/bin/units_from_xls.py
grybmadsci: or use `'''` as a string delimiter instead of just `'`

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+  """Return the string in uppercase with the specified substitutions."""
+  replacements = replacements or {}
+  result = string
+
+  for old, new in replacements.iteritems():
```
grybmadsci: alternatively, you could use `string.translate()`, but then have a bit more complexity in the building of the translation table.  you could probably do something like this, but it starts getting unreadable quickly, up to you:
```
result = string.translate(''.join(REPLACEMENTS.get(chr(i), chr(i)) for i in xrange(256)))
```

openhtf/util/units/__init__.py
```diff
@@ -0,0 +1,1170 @@
+# coding: iso-8859-1
```
grybmadsci: I would actually prefer to have this generated as part of the build process rather than having it checked in.  Having it checked in means having the same information duplicated in two places, leaving opportunity for it to get out of sync.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+    for row in rows:
+      name = row[col_indices[COLUMN_NAMES[0]]].value
+      code = row[col_indices[COLUMN_NAMES[1]]].value
+      suffix = multi_replace(row[col_indices[COLUMN_NAMES[2]]].value,
+                             {"'": "\\\'", '"': '\\\"'})
+      if name in seen:
+        continue
+      seen.add(name)
+
+      yield 'UNITS[\'%s\'] = Unit(\'%s\', \'%s\', \'%s\')\n' % (
```
grybmadsci: use " around strings that have ' in them, so you don't have to escape them all.  also, i'd recommend using triple quotes for that last field and not worrying about escaping quotation marks elsewhere (yay python).

openhtf/util/units/bin/units_from_xls.py
grybmadsci: Also, the header should have in caps at the very top of the file "THIS FILE IS AUTOMATICALLY GENERATED, DO NOT EDIT"

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+                      help='Where to put the generated .py file.')
+  args = parser.parse_args()
+
+  code = code_generator(
+      xlrd.open_workbook(args.infile).sheet_by_name(SHEET_NAME))
+
+  insert_into_file(args.outfile, code, START_DELIMITER, END_DELIMITER)
+
+
+def code_generator(sheet):
```
grybmadsci: `generate_units_from_sheet`?

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+  Yields: An iterable of lines of Python source code defining OpenHTF Units.
+  """
+  seen = set()
+  try:
+    col_indices = {}
+    rows = sheet.get_rows()
+    
+    # Find the indices for the columns we care about.
+    for idx, cell in enumerate(rows.next()):
+      if cell.value in COLUMN_NAMES:
```
grybmadsci: should take `COLUMN_NAMES` as an argument here and pass it in from the top level, better to reference global constants as far up the stack as possible to keep the lower bits more hermetic.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+        col_indices[cell.value] = idx
+
+    # loop over all remaining rows and pull out units.
+    for row in rows:
+      name = row[col_indices[COLUMN_NAMES[0]]].value
+      code = row[col_indices[COLUMN_NAMES[1]]].value
+      suffix = multi_replace(row[col_indices[COLUMN_NAMES[2]]].value,
```
grybmadsci: you can avoid this escaping by using triple quotes for the suffix strings.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,160 @@
...
+                             {"'": "\\\'", '"': '\\\"'})
+      if name in seen:
+        continue
```
grybmadsci: hm, maybe warn on this?  defaulting to the first occurrence seems reasonable, but you might want to know that entries are being skipped.

openhtf/util/units/bin/units_from_xls.py
grybmadsci: also i would name this more specifically, this isn't a global replacement dict for everything, only the `UNITS` key, maybe `unit_key_replacements` or something.

grybmadsci:
like the overall solution, couple nits, and probably a little bit of work to make it happen at build time rather than having a copy checked in, but i think it's worth doing.  it will avoid keeping two duplicates of the same info that might get out of sync, and will make it easier to update than having to look up invokation instructions for the tool.

openhtf/util/units/__init__.py
```diff
@@ -0,0 +1,1170 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Units of measure for OpenHTF.
+
+Used to retrieve UNECE unit codes/suffixes as follows:
+
+from openhtf.utils.units import UNITS
```
grybmadsci: style guide forbids importing symbols from within modules.  you could probably do sys.modules swap and just make the whole module just the `UNITS` dict, but that might break ability to pickle them because the class definition of `Unit` goes away (just a hunch, try it first).  maybe just put the `Unit` definition in `test_record.py` or something and reference it here, then you can do `sys.modules[__name__] = UNITS` at the end and users can do ```Python
from openhtf.util import units

u = units['RADIAN']
```

File: openhtf/util/units/__init__.py (hunk 8 above)
Comment (fahhem) [inline openhtf/util/units/__init__.py]: There isn't a second place for this, unless you're talking about the proto? The source is that xls file, but he's not checking that in.

File: openhtf/util/units/__init__.py (hunk 1 above)
Comment (jettisonjoe) [inline openhtf/util/units/__init__.py]: Oops, those were holdovers from the old version. Will fix.

**Comment by jettisonjoe** [comment]
Here's something I didn't notice until now. UNECE codes don't include bits/bytes. Do we care? Do we add them into the mix ourselves?

**Comment by fahhem** [comment]
It's in Annex II

**Comment by fahhem** [comment]
Actually, it looks like Annex II is much more complete and organized

**Comment by jettisonjoe** [comment]
Huh. For some reason I thought they were the same set of units just listed in a different order.

File: openhtf/util/units/bin/units_from_xls.py (hunk 13 above)
Comment (jettisonjoe) [inline openhtf/util/units/bin/units_from_xls.py]: There are a lot of entries for "one" in the sheet; there are gonna be a lot of spurious warnings if I do this.

File: openhtf/util/units/bin/units_from_xls.py (hunk 12 above)
Comment (jettisonjoe) [inline openhtf/util/units/bin/units_from_xls.py]: gp. If I use the "Annex II & Annex III" sheet (which _does_ seem more complete, I also need to triple-quote the names, as at least one includes an apostrophe.

File: openhtf/util/units/__init__.py (hunk 14 above)
Comment (jettisonjoe) [inline openhtf/util/units/__init__.py]: It's probably more straightforward to just recommend importing the module and using `units.UNITS` rather than importing `UNITS` from `units`, no?

File: openhtf/util/units/__init__.py (hunk 8 above)
Comment (jettisonjoe) [inline openhtf/util/units/__init__.py]: @grybmadsci, if I don't check in the generated Python, I'm going to need to either check in the .xls file, or download and unzip it from UNECE during the build step. I dislike the first because I don't like republishing someone else's document, and I dislike the second because of the potential volatility of UNECE URL's, .zip archive structure, and spreadsheet naming.

On the other hand, I think it's going to be tricky to make the generated code pass lint without exceptions. Would like to discuss more before moving code generation to build step.

**Comment by jettisonjoe** [comment]
More annoying stuff: there are two 'second's. WIth the current code, one becomes `SECOND_UNIT_OF_ANGLE` and the other becomes `SECOND_UNIT_OF_TIME`. This seems pretty cumbersome when most of the time you just want to use `SECOND` to mean the time one. Suggestions?

**Comment by jettisonjoe** [comment]
Right now I'm leaning towards a lot of regex-based one-off hacking to clean up a bunch of stuff.
  * The `UNIT_OF` items have a bit of special casing as I just mentioned.
  * Many of the BTU-related unit names are excessively long.
  * Some of the this-per-this-per-that unit names would be friendlier with abbreviations.

Question: is it excessive to have a 'name' field when the key is already so similar to the name?

File: openhtf/util/units/__init__.py (hunk 8 above)
Comment (grybmadsci) [inline openhtf/util/units/__init__.py]: ah, I missed that the .xls file wasn't also getting checked in.  hm, it's a little riskier checking in the generated output instead of the xls because if the xls goes away, there's no way to get it back (whereas the other direction is easy).  i don't think lint is really a problem, just add disable comments where needed ;)

It's up to you which you prefer, i have a slight preference for checking in the xls and generating at build-time, but as long as both aren't checked in side by side, i'm okay with it.

File: openhtf/util/units/bin/units_from_xls.py (hunk 13 above)
Comment (grybmadsci) [inline openhtf/util/units/bin/units_from_xls.py]: you could warn once for any given duplicated name but ignore further duplicates of the same name.

File: openhtf/util/units/__init__.py (hunk 14 above)
Comment (grybmadsci) [inline openhtf/util/units/__init__.py]: in my (admittedly limited) experience using this, it gets *real* annoying very quickly to have a long name for this, and if it's a sub-variable in a name, you can't even use import renaming to make it shorter, you have to use an assignment at the top of the file as an alias.

if you're all right with that, then that's fine, but expect a lot of `U = units.UNITS` at the top of files.

**Comment by grybmadsci** [comment]
I think it's pretty reasonable to have (well documented) aliases for ease-of-use purposes, such as `UNITS['SECOND'] = UNITS['SECOND_UNIT_OF_TIME']`, but i think it's wise to avoid much more complexity than that (like special handling of 'UNIT_OF' cases).  This way, we can point at the xls and say "use these (here are a few additions for convenience)" rather than "use uses, except when this, and those ones are annoying so we changed them to this".

There's always room for someone to grab annoyingly long units they use at the top of the file and alias them to shorter names themselves.  This also gives people the freedom to use aliases they feel are most appropriate (you thinking naming things is hard, try re-naming already named things in a way that everyone agrees on).

File: openhtf/util/units/bin/units_from_xls.py (hunk 9 above)
Comment (jettisonjoe) [inline openhtf/util/units/bin/units_from_xls.py]: At least one of the quotation forms will still need to be escaped in the generated code as is done on line 105 with `multi_replace`.  Specifically, the type that's being used to enclose the suffixes, since both `'` and `"` appear as suffixes.

File: openhtf/util/units/bin/units_from_xls.py (hunk 10 above)
Comment (jettisonjoe) [inline openhtf/util/units/bin/units_from_xls.py]: Roger.

**Comment by fahhem** [comment]
I think we can provide a lookup functionality that allows users to use the name as-is, the ALL CAPS version, or the units themselves ('m/s').

Since it's generated now, I think we can restructure it to take the dict out:

```py
UnitInfo = collections.namedtuple('Unit', 'name code suffix')

METERS_PER_SECOND = UnitInfo('meters per second', '...', 'm/s')
...
ALL_UNITS = [
    METERS_PER_SECOND,
    ...,
]
UNITS_BY_NAME = {u.name: u for u in ALL_UNITS}
UNITS_BY_SUFFIX = {u.suffix: u for u in ALL_UNITS}

class UnitLookup(object):
  def __init__(self, lookup):
    self.lookup = lookup
  def __getattr__(self, attr):
    return self.lookup[attr]

UNITS_BY_ALL = {}
UNITS_BY_ALL.update(UNITS_BY_NAME)
UNITS_BY_ALL.update(UNITS_BY_SUFFIX)
Unit = UnitLookup(UNITS_BY_ALL)
```

Now users have this API:

```
from openhtf.util import units

@measures(units.METERS_PER_SECOND)
@measures(units.Unit('m/s'))
@measures(units.Unit('meters per second'))
```

And there's no sys.modules magic to boot.

jettisonjoe:
@fahhem that's a cool idea; I like that test author API a lot. Will pursue and send an update to the PR on Monday.

grybmadsci:
yea, that still leaves potentially a lot of `units.blah`, but the more powerful interface I think is definitely worth it, and the direct attribute access via all-caps allows for import-as semantics to make it much shorter.

awesome idea, @fahhem :)

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -95,21 +99,23 @@ def code_generator(sheet):
...
 
     # loop over all remaining rows and pull out units.
     for row in rows:
-      name = row[col_indices[COLUMN_NAMES[0]]].value
-      code = row[col_indices[COLUMN_NAMES[1]]].value
-      suffix = multi_replace(row[col_indices[COLUMN_NAMES[2]]].value,
-                             {"'": "\\\'", '"': '\\\"'})
+      name = row[col_indices[column_names[0]]].value
+      code = row[col_indices[column_names[1]]].value
+      suffix = row[col_indices[column_names[2]]].value.replace("'", "\\\'")
```
grybmadsci: i don't think you need to escape these if you are using triple quotes in the generated output.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -95,21 +99,23 @@ def code_generator(sheet):
...
       if name in seen:
         continue
       seen.add(name)
 
-      yield 'UNITS[\'%s\'] = Unit(\'%s\', \'%s\', \'%s\')\n' % (
-          multi_replace(name.upper(), REPLACEMENTS), name, code, suffix)
+      yield "UNITS['%s'] = Unit('%s', '''%s''')\n" % (
+          multi_replace(name.upper(),
+                        UNIT_KEY_REPLACEMENTS),
```
grybmadsci: this indentation is super confusing, looks like all 4 lines are args to multi_replace.  i'd put `UNIT_KEY_REPLACEMENTS),` on the previous line

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -95,21 +99,23 @@ def code_generator(sheet):
...
+      code = row[col_indices[column_names[1]]].value
+      suffix = row[col_indices[column_names[2]]].value.replace("'", "\\\'")
       if name in seen:
         continue
       seen.add(name)
 
-      yield 'UNITS[\'%s\'] = Unit(\'%s\', \'%s\', \'%s\')\n' % (
-          multi_replace(name.upper(), REPLACEMENTS), name, code, suffix)
+      yield "UNITS['%s'] = Unit('%s', '''%s''')\n" % (
+          multi_replace(name.upper(),
```
grybmadsci: rather than a generic `multi_replace`, i think it's a bit better encapsulated if you name it `unit_key_from_sheet_name` or something similar, and do the `upper()` and replacement there.  this is very self-contained code, no need for designing for reusability.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: I believe a least one of `'` or `"` needs to be escaped, since both appear by themselves as suffixes. With triple quoting alone, you end up with either `'''''''` or `"""""""`, which parse as an empty string followed by `'` or `"` respectively.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Yes, this was my mistake. I will fix it.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Good call. I was arriving at a similar thought as I was working on this more today. Will refactor.

openhtf/util/units/bin/units_from_xls.py
grybmadsci: ah, right, i hadn't thought of those cases, forgot that single quote characters were suffixes :( yea, i guess you need to escape it.  you could replace with either `"\\'"` or `r'\''`, the triple backslash is unnecessary (this time i actually tried it :) )

jettisonjoe: Oh yeah, I like `r'\''`. that's pretty clean.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Good point.

jettisonjoe:
Made a bunch of fixes.
  * Implemented @fahhem's idea for test author API.
  * Auto-gen the entire `__init__.py` now instead of just a piece of it.
  * Updated to the latest version of the spreadsheet.
  * Use unicode ordinals to solve character replacement headaches (good luck to the person who tries to do this with the next version of the spreadsheet that adds more weird characters :-\
  * Cleaned up structure a bit.

It still fails tests because I didn't update the places where it's used (I might need some help with the `mfg_inspector` output module). But PTAL at what I have.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -104,31 +204,32 @@ def unit_defs_from_sheet(sheet, column_names):
...
-  result = string
+  result = name[:]
 
   for old, new in replacements.iteritems():
     result = result.replace(old, new)
 
+  # Collapse redundant underscores and convert to uppercase.
+  result = re.sub(r'_+', '_', result.upper())
+
   return result
```
fahhem: remove `insert_into_file()`?

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -49,20 +53,107 @@
...
+ALL_UNITS.append(NO_DIMENSION)
+
+NONE = UnitInfo('None', None, None)
+ALL_UNITS.append(NONE)
+'''
+
+POST = '''
+# Convenience aliases.
+MINUTE = MINUTE_UNIT_OF_TIME
+ALL_UNITS.append(MINUTE)
```
fahhem: This doesn't do anything since it's already in `ALL_UNITS` and everything else uses it in a dict, no? I'd `del ALL_UNITS` at the end since it doesn't make sense to use it directly, too.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -49,20 +53,107 @@
...
+
+SECOND = SECOND_UNIT_OF_TIME
+ALL_UNITS.append(SECOND)
+
+
+class UnitLookup(object):
+  """Facilitates user-friendly access to units."""
+  def __init__(self, lookup):
+    self.lookup = lookup
```
fahhem: `_lookup`

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -104,31 +204,32 @@ def unit_defs_from_sheet(sheet, column_names):
...
+      yield "%s = UnitInfo('%s', '%s', '''%s''')\n" % (key, name, code, suffix)
+      yield "ALL_UNITS.append(%s)\n" % key
 
   except xlrd.XLRDError:
     sys.stdout.write('Unable to process the .xls file.')
 
 
-def multi_replace(string, replacements=None):
-  """Return the string in uppercase with the specified substitutions."""
+def unit_key_from_name(name, replacements=None):
```
fahhem: still don't need to allow `replacements` to be `None`.

Don't need to copy name using `[:]`, it's immutable. I'd just use `name` throughout to avoid the extra-but-unused variable.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -104,31 +204,32 @@ def unit_defs_from_sheet(sheet, column_names):
...
+  """Return a legal python name for the given name for use as a unit key."""
   replacements = replacements or {}
-  result = string
+  result = name[:]
 
   for old, new in replacements.iteritems():
     result = result.replace(old, new)
 
+  # Collapse redundant underscores and convert to uppercase.
+  result = re.sub(r'_+', '_', result.upper())
```
fahhem: Do any of them start with `_` after all this replacement?

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Good point, and alright.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Mm, gp.

fahhem:
Might want to add "Fixes #298" to the description so merging this closes that issue

jettisonjoe:
Is `units` a bad name for the module since it's also the name of a kwarg to `Measurement`? I see it being renamed to `uom` a lot. Should we just make it official and call it `uom`?

fahhem:
I don't see a reason why the kwarg name would conflict with the module name, could you explain further?

I think it's only renamed to uom because of the previous module's `UOM` object.

jettisonjoe:
No actual conflict, but I assumed it was being done for clarity, and figured if it's _always_ being renamed, let's not fight the tide. If everyone's cool with keeping it `units` then that works for me.

wallacbe:
Very exhaustive list.  I see sensors and voltage/current readings covered.  Couple issues/questions.

I don't see anything referencing digital counts:
According to the University of Chicago, after a quick google search, "Raw value: The digital number output by the ADC; the units in which raw values are measured are called ADU (analog-to-digital units) or DN (data numbers). " theory.uchicago.edu/~ejm/pix/20d/tests/noise – KyranF Oct 29 '14 at 20:37

Also a generic count, like if we need to record a number of attempts.

Also how should we handle a boolean "status", leave that unitless?

grybmadsci:
hm, it's a shame that DN isn't there, but it's a bit of a stretch to consider that a 'unit' because the actual physical units it corresponds to varies.

as for 'count', that's not really a unit either, that's more a semantic descriptor (ie, HOW the measurement was taken, not WHAT it represents - the units would be what it is you're counting).

in both cases, i think it's pretty reasonable to build that information into the measurement name.  for example, 'retry_count' as a measurement name i think is more useful than a `retry` measurement with a COUNT unit type.  DN I admit would be nice to have a Unit, but you can still do `light_dn`, `dark_dn`, etc.

grybmadsci:
Oh, the conflict there is only because `@monitors` takes a 'units' kwarg, and I didn't want to rename the kwarg because it's a public-facing interface, so I aliased the `units` module as `uom`, but I don't see anywhere else that's done.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,235 @@
...
+import xlrd
+
+
+# Column names for the columns we care about. This list must be populated in
+# the expected order: [<name label>, <code label>, <suffix label>].
+COLUMN_NAMES = ['Name',
+                'Common\nCode',
+                'Symbol']
+
+PRE = '''# coding: iso-8859-1
```
grybmadsci: nit: this coding comment may not be accurate since you `string.encode('utf8')` below.

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,235 @@
...
+      key = unit_key_from_name(name)
+      if key in seen:
+        continue
+      seen.add(key)
+
+      yield "%s = UnitInfo('%s', '%s', '''%s''')\n" % (key, name, code, suffix)
+      yield "ALL_UNITS.append(%s)\n" % key
+
+  except xlrd.XLRDError:
+    sys.stdout.write('Unable to process the .xls file.')
```
grybmadsci: might be worth exiting here so you don't create a partial file?

grybmadsci:
LGTM pending a couple nits inline, and these last questions:

- Looks like it was decided to not make it part of the build process?  I'm all right with that, but wanted to make sure it was intentional.
- I don't think the `units_from_xls.py` script should go in the actual source tree.  It's very much a build-time thing (in fact, the way you have it, it's a meta-build step), and I think it's out of place in the openhtf package tree, maybe throw it in `contrib`?

openhtf/util/units/bin/units_from_xls.py
```diff
@@ -0,0 +1,235 @@
...
+and modern, and because they are recognized internationally. For full details
+regarding where we get the codes from and which units are avaiable, see the
+docstring at the top of openhtf/util/units/bin/units_from_xls.py.
+"""
+
+
+import collections
+
+
+UnitInfo = collections.namedtuple('Unit', 'name code suffix')
```
grybmadsci: Don't forget to change '`Unit'` to `'UnitInfo'` here again, looks like it got reverted and is breaking your tests.

fahhem:
If it doesn't go here, then the source tree can revert to a single `units.py` file. I think it should go in a `bin/` folder rather than `contrib/`

grybmadsci:
Yeah, `contrib/` is a weird spot, I was thinking perhaps `tools/`, but `bin/` works too, and I think is the most conventional place to put it, +1 to a toplevel `bin/`.

openhtf/util/units/bin/units_from_xls.py
jettisonjoe: Less reverted and more I never pushed those changes and have been home sick ever since hehe. I'm still WFH today with a lingering cough, but will get this PR finished up and merged.

jettisonjoe:
Done and should be passing unit tests now. PTAL.

To follow up on a couple of loose threads in the convo

  * For generic 'count' units, maybe use `units.ONE`?
  * Yes it was a conscious decision to check in the generated module and make it a manual metabuild step rather than republishing the UNECE standard. If over time this seems like a bad decision, I'm willing to rethink it.

examples/all_the_things.py
```diff
@@ -76,10 +77,10 @@ def set_measurements(test):
 
 
 @measures(
-    Measurement('unset_dims').WithDimensions(UOM['HERTZ']),
-    Measurement('dimensions').WithDimensions(UOM['HERTZ']),
+    Measurement('unset_dims').WithDimensions(units.HERTZ),
```
fahhem: oh, such a sweet API! Would it make sense to put `units` in `names`?

test/util/measurements_test.py
```diff
@@ -109,12 +111,10 @@ def _SaveResult(test_record):
     test.Configure(http_port=None)
     if self.UPDATE_OUTPUT:
       test.AddOutputCallbacks(_PickleRecord)
```
grybmadsci: oh hey, i almost forgot, @wallacbe was awesome and wrote this into a handy base class in `io/output/__init__.py`, so you can actually do this and remove `_PickleRecord`:
```Python
# At the top (might as well also clean this up):
RECORD_FILENAME = os.path.join(
    os.path.dirname(os.path.abspath(__file__)), 'measurements_record.pickle')

# Replace _LocalFilename('measurements_record.pickle') with RECORD_FILENAME.

# Instead of test.AddOutputCallbacks(_PickleRecord)
  test.AddOutputCallbacks(output.OutputToFile(RECORD_FILENAME))
```

openhtf/bin/units_from_xls.py
```diff
@@ -0,0 +1,250 @@
...
+    # The following three lines are equivalent:
+    @measures(units.METRE_PER_SECOND)
+    @measures(units.Unit('m/s'))
+    @measures(units.Unit('metre per second'))
+
+OpenHTF uses UNECE unit codes internally because they are relatively complete
+and modern, and because they are recognized internationally. For full details
+regarding where we get the codes from and which units are avaiable, see the
+docstring at the top of openhtf/util/units/bin/units_from_xls.py.
+"""
```
fahhem: Maybe add some more "DO NOT EDIT" statements in this docstring? Not a big deal

openhtf/bin/units_from_xls.py
```diff
@@ -0,0 +1,250 @@
...
+      type=str,
+      default=os.path.join(os.path.dirname(__file__), os.path.pardir, 'util',
+                           'units.py'),
+      help='where to put the generated .py file.')
+  args = parser.parse_args()
+
+  if not os.path.exists(args.xlsfile):
+    print 'Unable to locate the file "%s".' % args.xlsfile
+    parser.print_help()
+    print 'HACKABEES', args.outfile
```
fahhem: ahem :)

openhtf/names.py
```diff
@@ -60,4 +59,3 @@
 TestPhase = openhtf.PhaseOptions
 PhaseResult = openhtf.PhaseResult
 Test = openhtf.Test
-UOM = openhtf.util.units.UOM
```
grybmadsci: Did you not want to do `Units = openhtf.util.units.Units` here?

openhtf/util/monitors.py
```diff
@@ -108,7 +108,7 @@ def Wrapper(phase_func):
     @plugs.plug(update_kwargs=False, **monitor_plugs)
     @measurements.measures(
         measurements.Measurement(measurement_name).WithUnits(
-            units).WithDimensions(uom.UOM['MILLISECOND']))
+            units).WithDimensions(uom.MILLISECOND))
```
fahhem: @grybmadsci, did you choose `MILLISECOND` because `SECOND` wasn't in the units list before? Should we switch to seconds now?

grybmadsci:
LGTM pending nits

fahhem:
Same, LGTM

openhtf/util/monitors.py
grybmadsci: i went with millis because all our other timestamps are in millis (`start_time_millis`, `end_time_millis`, etc.).  I'm not particularly attached to it either way.

fahhem: Ah, then never mind.

openhtf/names.py
jettisonjoe: I thought the plan was to move away from `names.py`, but maybe it's still useful as a todo list for when we move to the new way.

openhtf/bin/units_from_xls.py
jettisonjoe: lol oops. That's not the first nor the last time a `print 'HACKABEES'` makes it into one of my PR's accidentally.

jettisonjoe:
Addressed final nits. Will merge as soon as travis tests finish and show passing (passes locally so I expect everything to go smoothly).

jettisonjoe:
Freakin' travis-ci randomly failed again, but then passed when I hit rebuild. Merging now.

fahhem:
Ack! Dictionary ordering dependency!

fahhem:
Please change the MessageToString call to use `use_index_order`: https://developers.google.com/protocol-buffers/docs/reference/python/google.protobuf.text_format-module

Otherwise the test will fail randomly depending on the seed Python randomly chooses at start up.

fahhem:
Oh, that wasn't introduced in this commit. It was added by @grybmadsci as a last-minute no-review change...

jettisonjoe:
Yep, I see it. This PR is becoming pretty overloaded :-\ , but I'll go ahead and fix this since I'm already here.

jettisonjoe:
Crud, that didn't fix it. Latest build just failed with the same problem.

fahhem:
Oh, shoot, it's not even that, it's the order of the config dict stored by the test as an attachment. Okay, go ahead and undo your last commit and merge. I'll fix the actual problem in a separate one

jettisonjoe:
Merging this for real now. @fahhem you might need to do a rebase after I merge this.

Issue #146, opened 2016-03-02
MacOS Support


wallacbe:
Some folks are trying to run on OSX.  Documenting issues
had to homebrew build SWIG
m2crypto was causing issues 

brew install openssl
 brew install swig
 sudo chown <user> /usr/local/bin
 brew link swig
 brew lin pcre
 sudo chown root /usr/local/bin
 sudo chown <user> /usr/local
 brew update
 sudo chown root /usr/local

wallacbe:
m2crypto actually still causing issues (adb usb plug)

jettisonjoe:
I'll try to take a look at this this week.

grybmadsci:
The 'Windows' support I did actually also includes OSX support, I believe.
Do you need to use the ADB USB plug?  If not, you should be able to use
just the openhtf package, without the usb plugs extra, and not depend on
M2Crypto.

Thanks for documenting the brew steps, though, that's super handy and we
should roll it into the docs!

wallacbe:
Don't need to use ADB USB plug.  How do we configure  to not use usb plugs extra?  Thanks

grybmadsci:
Hm, you may need to perform the build step on a linux machine, but the
install process should simply not depend on M2Crypto unless you explicitly
pip install openhtf[usb_plugs].  Try `python setup.py sdist` on a linux
machine, grab the .tar.gz from dist/, and then `pip install <file>` that on
OSX?

Sorry, I'm not too familiar just yet with some of the nuances of python
packaging and dependencies, this is just some of the flailing that I've
done, so maybe it helps.

grybmadsci:
I've successfully run openhtf tests on my Mac now (with usb plugs moved into an extra, those haven't been made to work).  @wallacbe can you give it a go from HEAD and let me know if you encounter any issues?  at some point we should get a clean mac image and document all the brew packages that need to be installed (i don't think there's many) alongside the apt-get equivalents.

wallacbe:
I have a fresh mac that I'm setting up and will update CONTRIBUTING.md.  Go ahead and assign issue to me.

wallacbe:
I was able to build and run hello world on MacOs after these steps:

# Install Xcode

# Install Homebrew Mac

brew install pkg-config libffi

brew install protobuf

git clone https://github.com/google/openhtf.git

sudo easy_install pip

# Make sure pip is up-to-date.
sudo pip install --upgrade pip

# Install virtualenv via pip.
sudo pip install virtualenv

[...]

PR #342, opened 2016-06-27
Added MacOS support on CONTRIBUTING.md

maddychan:


maddychan:
Hey Googlebot, I'm a Google employee and accidentally did the commit with my personal email instead of my google.com email - it's on my github profile though, and I switched it to be my primary email now. Thanks!

jettisonjoe:
Fixes #146.

Thanks for doing this! Inline comments to follow.

CONTRIBUTING.md
```diff
@@ -179,7 +179,36 @@ virtualenv venv
 # Install openhtf into the virtualenv in dev mode.
 python setup.py develop
 ```
+### MacOS
```
jettisonjoe: I'd put an empty line here to make the .md more readable, even if it actually _renders_ the same either way.

CONTRIBUTING.md
```diff
@@ -179,7 +179,36 @@ virtualenv venv
...
 
+Install Xcode and Homebrew Mac. Once that is done:
```
jettisonjoe: I would add a very short explanation of what these are and why they're needed, and hyperlink to their respective landing pages.

CONTRIBUTING.md
```diff
@@ -179,7 +179,36 @@ virtualenv venv
 # Install openhtf into the virtualenv in dev mode.
 python setup.py develop
 ```
+### MacOS
 
+Install Xcode and Homebrew Mac. Once that is done: 
+
+```bash
+#Install dependent packages
```
jettisonjoe: nit: A space between the # and the comment text makes the comment more readable.
nit+: I like full capitalization and punctuation on comments as if they're sentences.

The [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml) doesn't explicitly state these nits, but all its examples are consistent with them.

jettisonjoe: I would just call these "dependencies". I think that's more clear as OpenHTF is on what could be called the "dependent" side of the relation here.

CONTRIBUTING.md
```diff
@@ -179,7 +179,36 @@ virtualenv venv
...
+
+#Change to the openhtf directory 
+cd openhtf
+
+#Create a new virtualenv
+virtualenv venv
+
+#Activate the new virtualenv
+. venv/bin/activate
+```
```
jettisonjoe: Should add a `python setup.py develop` step at the end.

CONTRIBUTING.md
maddychan: It seems that a gcc compiler is needed for Homebrew and pip, and xcode is one way to provide that but not the only way. What do you suggest I include? Just Homebrew and the landing page?

jettisonjoe: Ah, I didn't realize xcode was where we were getting pip from, too. I think it's reasonable to recommend and link to both, as long as we provide a few words explaining why.

maddychan:
Just redid the steps with my newly reimaged mac and it went perfectly. yay!

grybmadsci:
LGTM, looks awesome, thanks!  next up is windows ;) (don't really, it's a nightmare, we'll deal with it later...)

jettisonjoe:
Awesome, merging now!

PR #343, opened 2016-06-27
Initial commit of updated XMLRPC-based station API

grybmadsci:


grybmadsci:
Okay, new PR created for branch on my fork merging into temporary dev branch on the main repo.  Action items left:

- Unit tests
- Doc updates/clean up TODOs
- Solve teardown phase nuances
- Prompt Plug
- Play with various test abort conditions, I believe some threads are hanging around in the station_api.

Done:
- Station API interface
- Update-notifying semantics
- Station polling script for exercising/demoing the station api (mostly done).

grybmadsci:
Merging so @jettisonjoe can start work on updating the frontend server.

openhtf/exe/__init__.py
```diff
@@ -100,22 +74,53 @@ def Stop(self):
...
+    Raises: TestAlreadyFinalized if end_time_millis already set.
+    """
+    if not self._test_state:
+      raise TestStopError('Test Stopped.')
+
+    if not self._test_state.is_finalized:
+      self._test_state.logger.info('Finishing test with outcome ABORTED.')
+      self._test_state.Finalize(test_record.Outcome.ABORTED)
+
+    _LOG.debug('Test completed for %s, saving to history and outputting.',
```
fahhem: This is the log statement that makes no sense, history isn't affected directly here. If this was moved up into `Test` then it would be logical

openhtf/io/station_api.py
```diff
@@ -0,0 +1,574 @@
...
+    self._known_tests = updated_tests
+    return self._known_tests.itervalues()
+
+
+### Server-side objects below here. ###
+
+
+class StationApi(object):
+
+  UID = '%s:%s' % (os.getpid(), util.TimeMillis())
```
fahhem: you probably want `uuid.uuid1()` here, it creates something host-specific, time-specific, with some randomness. You could cause it to overlap if you ran a sufficient number of pythons in the same second, but it has 14 bits of randomness so you'd need to run quite a few.

openhtf/io/station_api.py
```diff
@@ -0,0 +1,574 @@
...
+                            skip_phases=0, skip_logs=0):
+    if (start_time_millis and
+        start_time_millis == state['test_record'].start_time_millis):
+      state['test_record'] = mutablerecords.CopyRecord(state['test_record'])
+      del state['test_record'].phases[:skip_phases]
+      del state['test_record'].log_records[:skip_logs]
+
+    return {
+        'status': state['status'].name,
+        'test_record': saxutils.escape(pickle.dumps(state['test_record'])),
```
fahhem: Might prefer wrapping this as `Binary` rather than escaping it: https://docs.python.org/2/library/xmlrpclib.html#xmlrpclib.Binary

That way `xmlrpclib` will escape it only when necessary

openhtf/util/data.py
```diff
@@ -138,3 +139,30 @@ def ConvertToBaseTypes(obj, ignore_keys=tuple()):
...
+    size = sys.getsizeof(current_obj)
+
+    if isinstance(current_obj, dict):
+      size += sum(map(sizeof, itertools.chain.from_iterable(
+          current_obj.iteritems())))
+    elif (isinstance(current_obj, collections.Iterable) and
+          not isinstance(current_obj, basestring)):
+      size += sum(map(sizeof, iter(current_obj)))
+    elif isinstance(current_obj, records.RecordClass):
+      size += sizeof(current_obj.__slots__)
```
fahhem: `__slots__` is not copied per instance, it's part of the class definition

openhtf/util/data.py
```diff
@@ -138,3 +139,30 @@ def ConvertToBaseTypes(obj, ignore_keys=tuple()):
...
+      size += sum(map(sizeof, current_obj.optional_attributes.itervalues()))
```
fahhem: `optional_attributes` isn't either, only the data in the slots themselves (or else what's the point of `__slots__`). These three lines can simplify to:

```py
size += sum(sizeof(getattr(current_obj, attr)) for attr in current_obj.__slots__)
```

openhtf/exe/__init__.py
grybmadsci: yeah moved up into `Test` and moved the log line with it.

openhtf/io/station_api.py
grybmadsci: thought about this when doing it, but i like that the UID gives a knowledgeable sysadmin some traceability (ie, you can look up the PID), and the randomness doesn't really buy anything since PID is guaranteed unique within a millisecond anyway (you've got bigger problems if you have cycled through all PIDs in <1ms).

openhtf/io/station_api.py
grybmadsci: done, but i think that just base64 encodes it always.

openhtf/util/data.py
grybmadsci: right, thanks, good catch.

openhtf/util/data.py
grybmadsci: ah, yep, good call.

PR #346, opened 2016-06-29
Test API updates

grybmadsci:
Another day, another `station_api` update... The primary focus here is mostly internal refactoring:
- `*Info`/`*Data` renamed to `*Descriptor` (`TestDescriptor` and `PhaseDescriptor`).  These contain static information about the respective thing, and do not change between executions.
- `*State` cleaned up extensively, particularly `PhaseData` -> `PhaseState` + `TestApi`
- `TestApi` established as the canonical object that is passed into test phases' first argument.  It's defined in `openhtf/__init__.py` and is thoroughly documented.  It also does not contain various internal guts like `PhaseData` used to.
- `Measurement` values are now merged into the `measurements list` (no more separate  `measured_values`).
- Monitors (not too many users) were updated a bit to have more consistent polling intervals.
- Lots of little nits updated.

There are still a handful of TODOs, and even more updates left to go before it's completely ready, but it's in a reasonable state right now, all tests pass, examples work, yay.  I'll hold off on merging this into `google/station_api` until tomorrow's update, after that you'll have to make comments elsewhere.


examples/example_plug.py
```diff
@@ -80,4 +80,4 @@ def TearDown(self):
   def Increment(self):
     """Increment our value, return the new value."""
```
jettisonjoe: Docstring no longer accurate, it looks like.

openhtf/__init__.py
```diff
@@ -367,20 +371,78 @@ def WithArgs(self, **kwargs):
...
+      # Underlying function has room for test_api as an arg. If it doesn't
       # expect it but we miscounted args, we'll get another error farther down.
-      return self.func(phase_data, **kwargs)
+      return self.func(
+          test_state if self.options.requires_state else test_state.test_api,
+          **kwargs)
     return self.func(**kwargs)
 
 
+class TestApi(collections.namedtuple('TestApi', [
```
jettisonjoe: Python style for abbrevs _might_ dictate `TestAPI` here instead of `TestApi`.

openhtf/__init__.py
```diff
@@ -367,20 +371,78 @@ def WithArgs(self, **kwargs):
...
+
+  Attributes:
+    attachments: Dict mapping attachment name to test_record.Attachment
+        instance containing the data that was attached (and the MIME type
+        that was assumed based on extension, if any).  Only attachments
+        that have been attached in the current phase show up here, and this
+        attribute should not be modified directly, use TestApi.Attach() or
+        TestApi.AttachFromFile() instead.
+
+    dut_id: This attribute provides getter and setter access to the DUT ID
```
jettisonjoe: Finally `dut_id` is a first-class citizen!

openhtf/exe/__init__.py
```diff
@@ -51,9 +51,9 @@ def __init__(self, test_data, test_start, teardown_function=None):
     super(TestExecutor, self).__init__(name='TestExecutorThread')
 
     self._teardown_function = (
-        teardown_function and openhtf.PhaseInfo.WrapOrCopy(teardown_function))
-    self._test_data = test_data
-    self._plug_manager = plugs.PlugManager()
+        teardown_function
+        and openhtf.PhaseDescriptor.WrapOrCopy(teardown_function))
+    self._test_desc = test_data
```
jettisonjoe: Why not rename the arg, too?

openhtf/exe/test_state.py
```diff
@@ -13,11 +13,18 @@
 # limitations under the License.
 
 
-"""TestState module for handling the lifetime of a test.
+"""Module for handling transient state of a running test.
```
jettisonjoe: Nice, much more clear/focused than the prev iteration.

openhtf/__init__.py
```diff
@@ -213,13 +239,13 @@ def Execute(self, test_start=None, loop=None):
...
+      self._test_desc.metadata['test_name'] = self._test_options.name
+      self._test_desc.metadata['config'] = conf._asdict()
       self.last_run_time_millis = util.TimeMillis()
 
       self._executor = exe.TestExecutor(
-          self._test_data, test_start, self._test_options.teardown_function)
-      _LOG.info('Executing test: %s', self.data.code_info.name)
+          self._test_desc, test_start, self._test_options.teardown_function)
+      _LOG.info('Executing test: %s', self.descriptor.code_info.name)
       self._executor.Start()
```
fahhem: At some point you moved output callbacks being called out of this class, but I think it should remain here. They're attached to the test and it makes sense that `Test` call them. If `executor.Finalize()` returned the final `test_state`, then this would call the callbacks and add to history. There's a log in there that mentions history, but it's the `Test` that adds the `history.append_record` callback in there, so it's really twisted as-is.

openhtf/__init__.py
```diff
@@ -334,19 +364,19 @@ class PhaseInfo(mutablerecords.Record(
 
   @classmethod
   def WrapOrCopy(cls, func):
```
fahhem: now that `PhaseDescriptor.WrapOrCopy` is so long, and it gets used by test authors (to call `WithArgs` on a phase that doesn't use plugs, for instance), could we get a better-named function that calls out to this? Maybe `AsPhase`?

examples/example_plug.py
grybmadsci: done

openhtf/__init__.py
grybmadsci: hm, can you cite a source?  i couldn't find anything in either direction for python, but google c++ style dictates:
`Prefer to capitalize acronyms as single words (i.e. StartRpc(), not StartRPC()).`

openhtf/exe/__init__.py
grybmadsci: ah, meant to, thanks.  also renamed to `test_descriptor` because unnecessary abbreviations are annoying.

openhtf/__init__.py
grybmadsci: good call, moved output callback looping back into here.

openhtf/__init__.py
grybmadsci: hm, i'll defer that to a separate change, could be as simple as an alias in `names.py`

grybmadsci:
going to merge this as it's in a mostly-working state.  i think i've addressed comments from this (and the previous) PR, but again since this is still a side working branch, things are still open for conversation, just keeping the ball rolling.

PR #347, opened 2016-06-29
Edit pylintrc file

maddychan:
Update the regex in some of the fields, make sure the pylint adheres to google style standards as much as possible

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -106,26 +106,26 @@ required-attributes=
 bad-functions=map,filter,apply,input
 
 # Regular expression which should only match correct module names
-module-rgx=(([a-z_][a-z0-9_]*)|([A-Z][a-zA-Z0-9]+))$
+module-rgx=[a-z_][a-z0-9][a-z0-9_]{1,30}$
 
 # Regular expression which should only match correct module level names
 const-rgx=(([A-Z_][A-Z0-9_]*)|(__.*__))$
 
 # Regular expression which should only match correct class names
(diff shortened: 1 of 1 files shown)


maddychan:
Hey Googlebot, I turned off the email privacy on my github account - I think that was the problem.

fahhem:
Can we separate updating pylintrc from making the codebase name-style compliant? This changes the APIs in OpenHTF rather significantly and can cause downstream pain if we don't meter this out.

jettisonjoe:
@fahhem, yeah, we can separate it out. It's likely we're gonna let our PyPI launch slip by a few days, but I'd still like to make a solid attempt at complying with our updated pylint rules in prep for PyPI launch.

grybmadsci:
Agreed, in particular we have a side-development branch in progress right now (station_api) with fairly extensive changes, and doing this right now would cause a ball of merge conflicts, let's wait until `station_api` is merged in to actually update the codebase to by new-pylintrc-compliant.

jettisonjoe:
We'll split it so this PR is just the pylintrc changes. But @maddychan is making great progress on updating the codebase to comply and we should keep momentum going on that. I'll help with the rebase for that separate set of changes when the time comes. I don't think it will be that bad. Plus, git merge conflict resolution practice!

maddychan:
Took out the commits regarding compliance to put in another commit; this now only contains the pylintrc changes

grybmadsci:
LGTM for now, may need to update this as special cases come up, but for now let's get it merged :)

PR #348, opened 2016-06-29
Save the config as a sorted-key json blob

fahhem:
The tests will likely break, so I'll fix those in a bit.


openhtf/io/output/mfg_inspector.py
```diff
@@ -118,7 +118,8 @@ def _PopulateHeader(record, testrun):
   if 'config' in record.metadata:
     attachment = testrun.info_parameters.add()
     attachment.name = 'config'
-    attachment.value_binary = str(record.metadata['config'])
+    attachment.value_binary = json.dumps(
+        record.metadata['config'], sort_keys=True)
```
grybmadsci: can you add `indent=4` also so it gets pretty-printed?

grybmadsci:
LGTM pending nit inline and this one:

Can you add a comment to the output module somewhere that output callbacks should be deterministic?  It's likely to get overlooked accidentally, but wouldn't hurt to have in there.

jettisonjoe:
Ah, looks like no rebase necessary. Didn't realize how small/easy this fix would be.

LGTM. Will merge once @grybmadsci's nits are addressed.

openhtf/io/output/mfg_inspector.py
fahhem: Done

jettisonjoe:
Looks like `record.testrun` needs to be updated with the new indented output.

fahhem:
Fixed and ready to go. Rebased and tested locally, though travis will take its sweet time to catch up

grybmadsci:
@kdsudac pulled this fix into a separate PR that passed tests sooner ;)

PR #355, opened 2016-07-11
Updates to actually trigger wait_for_update().

grybmadsci:
Still remaining are unit tests and fleshing out plugs RPC interfaces.


openhtf/io/station_api.py
```diff
@@ -507,60 +587,78 @@ def wait_for_update(self, test_uid, remote_state, timeout_s):
```
fahhem: the docs say "TODO: Measurement" but this PR does that

grybmadsci: updated, finished cleaning up this doc while i was here.

grybmadsci:
Merging to unblock @jettisonjoe on frontend server work.

PR #359, opened 2016-07-14
Pylint plugin

maddychan:
Add pylint plugin to deal with conf module. Take care of warning messages but doesn't check for whether the dynamically added attributes are instantiated in the same module as they're used yet


maddychan:
googlebot I signed the CLA

openhtf/conf.py
```diff
@@ -238,6 +239,7 @@ def __init__(self, logger, lock, parser, **kwargs):
     self._modules = kwargs
     self._declarations = {}
     self.ARG_PARSER = parser
+    self.__name__ = 'Configuration'
```
grybmadsci: I believe you can actually do type(self).__name__ here, as the `Configuration` type still exists when `__init__` is called (just not when anything else is called).

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+    except AttributeError:
+      pass
+
+    # TODO(maddychan): Figure out how to add a custom failure message here.
+    return func('no-member', line, node, args, **kwargs)
+
+  return add_message
+
+#storing a reference to the openhtf.conf module node
+clss = []
```
grybmadsci: this is one of those rare cases where it's probably better to use `global` than to use lists, so initialize these to `None` and in your `trans_*` methods do `global clss`, etc. so you can assign to it directly.

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+clss_locals = {}
+current_root = []
+
+def trans_expr(node):
```
grybmadsci: a side effect of doing it this way is that i believe it depends on the `conf` module being processed before any that reference it, because otherwise your `clss` and `clss_locals` aren't set.  that might be implied, astroid may parse the modules that a module imports before parsing the module itself.  if you've tried this and it works, then neat.

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+  if (isinstance(node.func, astroid.Attribute) and
+      isinstance(node.func.expr, astroid.Name) and
+      node.func.expr.name == 'conf' and
+      node.func.attrname == 'Declare'):
+
+    #import pdb
+    #pdb.set_trace()
+
+    #keeping track of the current root, refreshing the locals if it changes
+    if current_root == []:
```
grybmadsci: `if not current_root:`

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+    	current_root.append(node.root())
+    elif current_root[0] != node.root():
+    	current_root[0] = node.root()
+    	clss[0].locals = clss_locals
+
+    #adding the conf attribute to the module's locals
+    if node.args and clss:
+    	clss[0].locals[node.args[0].value] = [None]
```
grybmadsci: Does this have to be `[None]` or can it just be `None`?  you just need `node.args[0].value` to be in the `locals`, right?  Also, as a side note, it doesn't *have* to be `args[0]`, it can technically be passed by keyword, hence the annoying logic below.  We don't currently actually do that anywhere, but it would be weird if pylint exploded over it.

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+  
+
+    # madsci's stuff - didn't use it sorry.... 
+    #  
+    # else:
+    #   for keyword in node.keywords:
+    #     if keyword.arg == 'name':
```
grybmadsci: here's where I checked for the `conf.Declare(name='foo')` case.

pylint_plugins/conf_plugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+    #storing reference to this node for future use
+    clss.append(cls)
+    clss_locals.update(cls.locals)
+
+
+def register(linter):
+  MANAGER.register_transform(astroid.Call, trans_expr)
+  MANAGER.register_transform(astroid.Module, conf_transform)
+  linter.add_message = wrap_add_message(linter.add_message)
```
grybmadsci: this line can go away if your locals-swap method works :)

pylintrc~
```diff
@@ -0,0 +1,285 @@
+[MASTER]
```
grybmadsci: whoops, don't forget to revert this file, it's a temp file created by your editor.

pylint_plugins/conf_plugin.py
maddychan: Yeah, I found that at runtime it seems to do the module transform once at the beginning and then only does the call transforms after that. I'm guessing that it has to process the module first in order to check that the call node is calling an actual attribute, and caches it after that. I think this should work? should I add some kind of error message if clss is still None?

pylint_plugins/conf_plugin.py
maddychan: gotcha. and yeah, it has to be [None]; locals is defined with string keys to list values (of references to the nodes), though the lists only ever have one value as far as I've seen... anyways, it dies when I try to do None so I just stuck it in a list.

grybmadsci: ah, right, i forgot this was astroid's internal 'locals' not the actual Python 'locals()' dict.

pylint_plugins/conf_plugin.py
grybmadsci: you could probably just `assert clss, 'openhtf.conf module not parsed, but seen in %s' % node.root()`

also the more I think about it, the more I think this should be encapsulated in a class so you can more cleanly maintain state across these calls, and then `clss` becomes `self.clss`, or, even better `self.conf_module_node`.

kudos for getting the `locals` trick working :)

maddychan: Just tried to stick that assert statement in, and it seems I was wrong - they do run through the class transforms at the beginning before running the module transforms. I think I'll just go into the code if clss is defined;  I don't know if it's possible to tell whether the function was being run before the actual tests or during. I also consolidated the variables in a class. 

and thanks!! ^___^

grybmadsci: did you put the assert lower, around line 42?  it's only important if you already know you're transforming a call to `conf.Declare`

maddychan: yups, that's exactly where I put it actually haha

jettisonjoe:
Superseded by #402.

PR #363, opened 2016-07-15
Optionally disable capturing source code

### LINKED ISSUE #360: Provide an option to disable source code capturing.
Some test authors might consider their tester code "IP" and want to prevent third party CM's from easily viewing it. We should consider having an option to exclude source code from the test record so it's not transmitted in plaintext over the test network, and it's not displayed in the operator interface (frontend).

For the record, I personally don't believe this is a good way to address the imagined problem, but I've had a few requests from test authors provide this option.

fahhem:
This is a quick, simple way to disable source code capturing so it's not even in memory (other than as bytecode).

Fixes #360

### DIFF
--- a/openhtf/io/test_record.py
+++ b/openhtf/io/test_record.py
@@ -29,6 +29,7 @@ from openhtf import util
 from openhtf.util import logs
 
 _LOG = logging.getLogger(__name__)
+CAPTURE_SOURCE_CODE = True  # Override before creating an openhtf.Test()
 
 
 class InvalidMeasurementDimensions(Exception):
@@ -82,6 +83,8 @@ class PhaseRecord(  # pylint: disable=too-few-public-methods,no-init
 
 
 def _GetSourceSafely(obj):
+  if not CAPTURE_SOURCE_CODE:
+    return ''
   try:
     return inspect.getsource(obj)
   except Exception:


fahhem:
It's also a tiny startup performance boost because you're reading fewer files from disk

openhtf/io/test_record.py
```diff
@@ -29,6 +29,7 @@
 from openhtf.util import logs
 
 _LOG = logging.getLogger(__name__)
+CAPTURE_SOURCE_CODE = True  # Override before creating an openhtf.Test()
```
grybmadsci: hm, shouldn't this be a commandline arg?

fahhem: No, those are parsed too late for this, I think. And it tends to be a
program decision rather than a per-invocation decision anyway

grybmadsci: so what's the intended way for a user to 'Override' this?  immediately after import via `test_record.CAPTURE_SOURCE_CODE = False`?

fahhem: Unfortunately, this happens at the time a phase gets defined, so they'd
need to do it before importing other things.

Otherwise we'd have to change it around to late-load the source code after
the test is created and likely after the args are parsed. Is that worth the
complexity?

grybmadsci: hm, i think it is a bit kludgey to have tunable knobs like this (that require poking between imports), but there is precedent for it (with logging, for example, and other modules like GUI tools where you have to configure a backend before importing other bits).

i think it's a pretty close tradeoff in this case, so if you'd prefer to keep it this way, i'm okay with it.  it does need to be clearly documented how to do this, though, and what the constraints are on when that attribute needs to be set.  with updated docs, LGTM.

openhtf/io/test_record.py
```diff
@@ -82,6 +83,8 @@ class PhaseRecord(  # pylint: disable=too-few-public-methods,no-init
 
 
 def _GetSourceSafely(obj):
+  if not CAPTURE_SOURCE_CODE:
+    return ''
```
grybmadsci: maybe make this `return '# Source code capturing disabled.'`?

jettisonjoe:
+1 to @grybmadsci's sentiments. I'm fine with this approach provided we have some documentation and examples to go along with it.

fahhem:
I went about it another way in another PR, so closing this one

PR #369, opened 2016-07-18
Fix typo in 'iff'

johnlunney:


grybmadsci:
This was actually intentional, it's a bit of shorthand from mathematics for "if and only if".

https://en.wikipedia.org/wiki/If_and_only_if

johnlunney:
That has two 'f''s, the one fixed has three :)

(The 'iff' further up in the file is correct)

grybmadsci:
Ah, I looked too quickly, thanks!

PR #370, opened 2016-07-19
first pass at plug rpc plumbing

grybmadsci:
this might break the world, but i want to get it in to unblock progress tomorrow.

### DIFF
--- a/openhtf/plugs/__init__.py
+++ b/openhtf/plugs/__init__.py
@@ -105,10 +105,13 @@ import threading
 import time
 
 import mutablerecords
+import sockjs.tornado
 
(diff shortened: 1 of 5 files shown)


openhtf/plugs/__init__.py
```diff
@@ -193,7 +246,7 @@ def plug(update_kwargs=True, **plugs):
   with @staticmethod.
 
   Args:
-    **plugs: Dict mapping name to Plug type.
+ wait_for_plug_update(   **plugs: Dict mapping name to Plug type.
```
fahhem: typo?

PR #374, opened 2016-07-20
actually fix teardown

grybmadsci:
we could really use an integration test that covers teardown functions... i'm fairly sure this covers all the cases.

### DIFF
--- a/openhtf/exe/phase_executor.py
+++ b/openhtf/exe/phase_executor.py
@@ -92,6 +92,12 @@ class PhaseOutcome(collections.namedtuple('PhaseOutcome', 'phase_result')):
     return isinstance(self.phase_result, (
         Exception, threads.ThreadTerminationError))
 
+  @property
+  def is_terminal(self):
+    """True if this outcome will stop the test."""
+    return (self.raised_exception or self.is_timeout or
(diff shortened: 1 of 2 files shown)


openhtf/exe/phase_executor.py
```diff
@@ -167,22 +173,30 @@ def execute_phases(self, phases):
...
+            self._execute_one_phase(teardown_func, output_record=False)
           yield outcome
-
-          # If there was an error, stop executing now.
-          if outcome.raised_exception:
-            return
-
+  
+          # We shouldn't keep executing if the outcome was terminal.
+          if outcome.is_terminal:
```
fahhem: I'd put this check in `TestState.SetStatusFromPhaseOutcome()` since that has access to `outcome` and is what the logic of `is_terminal` is trying to match. That way if this changes somehow, we'd be pointed at the right function immediately

grybmadsci: ah, good idea, took me a minute to figure out what you were suggesting.. i did hesitate a bit to implement `is_terminal`, since it's largely a duplication of `TestState.SetStatusFromPhaseOutcome()`, but it's a good idea to put this check directly in there so we get an obvious failure place if they get out of sync.

PR #375, opened 2016-07-20
Phase result exception

maddychan:
Give more info on exception and traceback when raised


openhtf/exe/phase_executor.py
```diff
@@ -55,6 +57,16 @@
 class InvalidPhaseResultError(Exception):
   """Raised when a PhaseOutcome is created with an invalid phase result."""
 
+class ExceptionOutcome(collections.namedtuple('ExceptionOutcome', ['exc_type','exc_val','exc_tb'])):
```
grybmadsci: Can we name this `ExceptionInfo` to match the python `exc_info` nomenclature?  Also, line too long :)

openhtf/exe/phase_executor.py
```diff
@@ -55,6 +57,16 @@
...
+
+  def _asdict(self):
+    return_dict = {}
+    excep = traceback.format_exception(self.exc_type,self.exc_val,self.exc_tb)
```
grybmadsci: `self` is a tuple, so you can do `traceback.format_exception(*self)` directly.

openhtf/exe/phase_executor.py
```diff
@@ -55,6 +57,16 @@
 class InvalidPhaseResultError(Exception):
   """Raised when a PhaseOutcome is created with an invalid phase result."""
 
+class ExceptionOutcome(collections.namedtuple('ExceptionOutcome', ['exc_type','exc_val','exc_tb'])):
+
+  def _asdict(self):
+    return_dict = {}
```
grybmadsci: it's kind of a personal style thing, but i usually like inline dict constructs rather than by-key, so:
```
return {
    'exc_type': str(self.exc_type),
    'exc_val': self.exc_val,
    'exc_tb': ''.join(traceback.format_exception(*self)),
}

File: openhtf/exe/phase_executor.py [hunk 4]
```diff
@@ -55,6 +57,16 @@
...
+    excep = traceback.format_exception(self.exc_type,self.exc_val,self.exc_tb)
+    return_dict['exc_type'] = str(self.exc_type)
+    return_dict['exc_val'] = self.exc_val
+    return_dict['exc_tb'] = "".join(a for a in excep)
```
grybmadsci: single instead of double quotes.  also, did you need to do `a for a in excep`, or could you just do `excep`?  I think `join()` takes an iterable directly.

openhtf/exe/phase_executor.py
```diff
@@ -122,7 +134,7 @@ def _ThreadProc(self):
     self._phase_outcome = PhaseOutcome(phase_return)
 
   def _ThreadException(self, exc):
```
grybmadsci: really, `_ThreadException()` should get passed in the `exc_info`, not just the exception instance.  you might modify `threads.py` to do this.  Looks like `threads.py` and this file are the only ones that use it anyway.  then you don't have to do the `sys.exc_info()` here.

openhtf/exe/phase_executor.py
```diff
@@ -69,15 +81,15 @@ class PhaseOutcome(collections.namedtuple('PhaseOutcome', 'phase_result')):
...
+  similarly be used to check for the timeout case. 
 
   The only accepted values for phase_result are None (timeout), an instance
   of Exception (phase raised), or an instance of openhtf.PhaseResult.  Any
   other value will raise an InvalidPhaseResultError.
   """
   def __init__(self, phase_result):
     if (phase_result is not None and
-        not isinstance(phase_result, (Exception, openhtf.PhaseResult))):
+        not isinstance(phase_result, (ExceptionOutcome, openhtf.PhaseResult))):
```
grybmadsci: this is missing some lines, you should rebase on station_api, as that's about to get merged into master any day now.

jettisonjoe:
Superseded by #413.

PR #380, opened 2016-07-21
Enforce the use of UnitDescriptors as units.

### LINKED ISSUE #208: Enforce use of Unit objects.
Right now, people can do things like call `WithUnits()` with a string instead of a `Unit`, and OpenHTF is cool with it. We should convert to `Unit` or throw an error. The multi-dimensional declaration probably needs a check too.

jettisonjoe:
Fixes #208


fahhem:
LGTM

openhtf/util/measurements.py
```diff
@@ -126,14 +127,23 @@ def Doc(self, docstring):
     self.docstring = docstring
     return self
 
-  def WithUnits(self, units):
+  def _maybe_make_unit_desc(self, unit_desc):
+    """Return the UnitDescriptor or convert a string to one."""
+    if isinstance(unit_desc, str):
+      unit_desc = units.Unit(unit_desc)
+    if not (isinstance(unit_desc, units.UnitDescriptor) or unit_desc is None):
```
grybmadsci: prefer factoring the not thru, instead of `not (something or something)`.  alternatively, in the case of None, you might auto-create the NO_UNITS unit or whatever it is.

openhtf/util/measurements.py
```diff
@@ -126,14 +127,23 @@ def Doc(self, docstring):
...
+  def WithUnits(self, unit_desc):
     """Declare the units for this Measurement, returns self for chaining."""
-    self.units = units
+    self.units = self._maybe_make_unit_desc(unit_desc)
     return self
 
   def WithDimensions(self, *dimensions):
     """Declare dimensions for this Measurement, returns self for chaining."""
-    self.dimensions = dimensions
+    self.dimensions = [self._maybe_make_unit_desc(dim) for dim in dimensions]
```
grybmadsci: interestingly, dimensions used to be a tuple (`*args` is a tuple), now it's a list.  shouldn't matter, but i don't see any reason for it to be a list (it never gets modified), so `tuple(...)` might be better.

grybmadsci:
LGTM pending minor nits.

jettisonjoe:
Addressed reviewer feedback and added a tiny unit test. PTAL.

jettisonjoe:
Merging based on verbal conversation with @grybmadsci.

PR #383, opened 2016-07-21
Fix a small bug that prevents the frontend from running.

jettisonjoe:
@grybmadsci, PTA quick L. This is related to your recent refactoring.


grybmadsci:
lgtm, whoops :)

PR #385, opened 2016-07-26
Update output modules to match Measurement restructuring.

### LINKED ISSUE #381: Update output modules in station_api branch due to Measurement restructuring.
This probably broke with measured_values being collapsed into Measurements.

jettisonjoe:
Fixes #381.


openhtf/io/output/json_factory.py
```diff
@@ -41,5 +41,6 @@ def convert_to_dict(self, test_record):
         for value in phase['attachments'].itervalues():
           value['data'] = base64.standard_b64encode(value['data'])
     else:
-      as_dict = data.ConvertToBaseTypes(test_record, ignore_keys=('attachments',))
+      as_dict = data.ConvertToBaseTypes(test_record,
```
fahhem: Why?

jettisonjoe: It was [83 characters long](https://google.github.io/styleguide/pyguide.html#Line_length).

jettisonjoe:
Wat... ...I think maybe this is failing because Travis needs the new mutablerecords. It's passing on my machine.

fahhem:
Oh, did you update mutablerecords via a github checkout? I haven't merged that PR, let alone pushed it to PyPI

jettisonjoe:
Yeah, I'm testing locally using the latest mutablerecords from source.

openhtf/io/output/mfg_inspector.py
```diff
@@ -171,14 +171,14 @@ def _MangleMeasurement(name, value, measurement, mangled_parameters,
   We generate these by doing some name mangling, using some sane limits for
   very large multidimensional measurements.
   """
-  for current_value in value[:MAX_PARAMS_PER_MEASUREMENT]:
+  for coord, val in value.value_dict.iteritems():
```
grybmadsci: probably still want to respect `MAX_PARAMS_PER_MEASUREMENT` here, with `.items()[:MAX_PARAMS_PER_MEASUREMENT]` instead.

openhtf/io/output/mfg_inspector.py
```diff
@@ -232,10 +231,10 @@ def _ExtractParameters(record, testrun, used_parameter_names):
       if measurement.units:
         testrun_param.unit_code = UOM_CODE_MAP[measurement.units.uom_code]
 
-      if name not in phase.measured_values:
+      if not measurement.measured_value:
         testrun_param.status = test_runs_pb2.ERROR
         continue
-      value = phase.measured_values[name]
+      value = measurement.measured_value
```
grybmadsci: if `dimensions` is `None`, then isn't `measurement.measured_value` a `measurements.MeasuredValue` that you still have to get the `.value` of?  Same for dimensioned measurements, I see a lot of places where `value` is used directly, but I suspect it needs to be `value.value`, such as when passed to `_MangleMeasurement()`.  Aliasing `measured_value` to `value` is confusing, too.

openhtf/util/measurements.py
```diff
@@ -143,6 +143,11 @@ def __setattr__(self, attr, value):
     if attr == 'dimensions':
       self._initialize_value()
 
+  def __setstate__(self, state):
```
grybmadsci: can you add a comment as to why this is here?  i think it's the cleanest way to do what you need, but still isn't super obvious why without a comment.

openhtf/util/measurements.py
```diff
@@ -336,14 +340,15 @@ def value(self):
...
-    value, the other elements will be the assocated coordinates.  The tuples
-    are output in the order in which they were set.
+      A list of dicts mapping coordinates to values measured at those
+      coordinates. Elements are output in the order in which they were set.
     """
     if not self.is_value_set:
       raise MeasurementNotSetError('Measurement not yet set', self.name)
-    return [dimensions + (value,) for dimensions, value in
-            self.value_dict.iteritems()]
+    return [{
```
grybmadsci: this is a bit of a contrived return type to suit JSON-friendly output, which isn't really part of this interface.  I'd rather move the dict-ification of these tuples at whatever layer is doing the JSON serialization itself.  Notably, dicts have a nontrivial amount of overhead compared to tuples, and you could potentially have a lot of them here.  also, this return value isn't really suitable for any kind of numerical analysis.  you could return namedtuples of `(dimensions, value)` if you want to avoid the weird semantic of "the last element is the value", but it was intended to represent a point in n-dimensional space, in which case the flat tuple makes the most sense.

openhtf/io/output/mfg_inspector.py
jettisonjoe: Good point.

openhtf/io/output/mfg_inspector.py
jettisonjoe: Alright, I'll clean this up so it's more clear what's what.

openhtf/io/output/mfg_inspector.py
grybmadsci: ah, `value` is a bit of a misnomer here, this parameter should be called `measured_value` for consistency.

openhtf/util/measurements.py
jettisonjoe: Yep, will do.

openhtf/util/measurements.py
jettisonjoe: The more I see the old way the less I like it. But we can keep it the old way for now because I'm out of time to argue about it, and because you might be right.

jettisonjoe:
Addressed feedback. PTAL.

grybmadsci:
LGTM, feel free to merge ignoring that travis breakage, we just can't merge station_api into master until mutablerecords pushes the fix to PyPI.  @fahhem any idea when you'll have a sec to do that/anything we can do to help?

fahhem:
Just did it! mutablerecords 0.3.0 is on PyPI and I didn't even mess up the formatting! If you update the requirements travis should pass

jettisonjoe:
@fahhem awesome, thank you!

PR #386, opened 2016-07-27
change blank dut_id to UNKNOWN_DUT_ID

wallacbe:
allow us to call Execute() without a lambda setting the serial number


grybmadsci:
The blank check shouldn't happen until the end of a test, filling in a sentinel value is not the solution, because then the blank check does nothing (and we explicitly require a DUT ID to be set *at some point*).

Can you send a stacktrace of the error, and someone will take a look at why the check is happening at the wrong time?

PR #387, opened 2016-07-27
Fix regressions in frontend and update prebuilt.

jettisonjoe:
Ignore commit 80de272; I'll drop it and rebase once #385 is merged.


examples/all_the_things.py
```diff
@@ -119,5 +119,6 @@ def teardown(test):
   #  with open('my_private_key.json', 'r') as json_file:
   #    test.AddOutputCallbacks(output.UploadToMfgInspector.from_json(
   #        json.load(json_file)))
-  test.Configure(teardown_function=teardown)
-  test.Execute(test_start=triggers.PromptForTestStart())
+  # test.Configure(teardown_function=teardown)
```
grybmadsci: hm, maybe don't change this file? don't think it matters much, but we should probably leave it in some kind of base-state in the repo and only change it locally for testing.

openhtf/io/frontend/src/app/station.service.ts
```diff
@@ -97,12 +97,12 @@ export class StationService extends SubscriptionService {
    */
   onmessage(msg) {
     this.reachable = true;
-    if (!msg) {
+    let parsedData = JSON.parse(msg.data);
```
grybmadsci: there's no checking on the format of msg here, or that it's not null, might be worth a descriptive console log message?

openhtf/io/frontend/src/app/utils.ts
```diff
@@ -41,6 +41,7 @@ export class SimplifyStatus extends MapPipe {
     'WAITING_FOR_TEST_START': 'WAITING',
     'STOP_WAIT': 'WAITING',
     'TIMEOUT': 'ERROR',
+    'ABORTED': 'ERROR',
```
grybmadsci: Values here are now in `openhtf.exe.test_state.TestState.Status`:
 - WAITING_FOR_TEST_START
 - RUNNING
 - COMPLETED

Not to be confused with `openhtf.io.test_record.Outcome`:
 - PASS
 - FAIL
 - ERROR
 - TIMEOUT
 - ABORTED

examples/all_the_things.py
jettisonjoe: Aaarrrgg. I make this mistake every single time I'm working on the frontend. I just usually catch it before I make a PR.

openhtf/io/frontend/src/app/utils.ts
jettisonjoe: Yeah, but the frontend color schemes are quite a bit more simplistic. Blue for running normally, red for failing normally, green for passing normally, yellow for something went wrong, and grey for something is down/missing. I agree this could use a cleanup, but I'd like to punt on it for now since it's just an elegance of design thing and not user-affecting.

openhtf/io/frontend/src/app/station.service.ts
jettisonjoe: Possibly. Frontend client needs better console debug logging overall. Filed #388 to track general console logging improvements in the web client.

openhtf/io/frontend/src/app/utils.ts
grybmadsci: Hm? i was talking about the string mapping in your status renaming, you have keys in here that  don't exist anymore (won't actually break, but they should be removed), but you're also missing 'RUNNING' and 'COMPLETED', which will probably break, right?

grybmadsci:
LGTM, pending status pipe comment.

openhtf/io/frontend/src/app/station.phases.html
```diff
@@ -26,8 +26,9 @@
     </div>
     <div class="collapsible-body {{phase.measurements | measToPassFail | statusToDark}}">
       <p *ngIf="phase.docstring" class="docstring">{{phase.docstring}}</p>
-      <div class="measurement-container">
-        <table class="measurement-listing white-text">
+      <div class="measurement-container"
+           [innerHTML]="phase.measurements | measToHtml">
+        <!-- <table class="measurement-listing white-text">
```
fahhem: did you mean to leave this commented out?

fahhem:
LGTM as well

openhtf/io/frontend/src/app/utils.ts
jettisonjoe: Ah. The confusion is because this pipe is a little overloaded. It simplifies status, sometimes for more concise human readability and sometimes for color lookups. To make things extra confusing (but more simple for operators), test outcomes and test status are used to fill in the same spot in the UI (and determine the color of the test headers). Outcomes are preferred because if there is an outcome, that's the most import piece of summary info to display. If the test isn't finished yet, we display a simplified version of the status instead. Which is why I say this could all use a cleanup, but it should actually work the way it stands.

grybmadsci: if you say so, but i don't see how this possibly works when the input status is "RUNNING' and there is no 'RUNNING' key in your map, is there no KeyError equivalent in ts?

jettisonjoe: Line 31. If the input isn't in the map, return it un-changed.

jettisonjoe:
Merging per our conversation over gchat. Thanks for the review, folks!

PR #389, opened 2016-07-28
Fix incorrect behavior in argv.StoreInModule

fahhem:
Also document it now that it works :)

Previously, this would always try to set attributes in the openhtf module itself, not openhtf.util.logs or whichever module was correct.

### DIFF
--- a/openhtf/util/argv.py
+++ b/openhtf/util/argv.py
@@ -1,4 +1,20 @@
-"""Utilities for handling command line arguments."""
+"""Utilities for handling command line arguments.
+
+StoreInModule:
+  Enables emulating a gflags-esque API (flag affects global value), but one
+  doesn't necessarily need to use flags to set values.
+  
+  Example usage:
+    DEFAULT_VALUE = 0
+    ARG_PARSER = argv.ModuleParser()
+    ARG_PARSER.add_argument(
+        '--override-value', action=argv.StoreInModule,
+        default=DEFAULT_VALUE, target='%s.DEFAULT_VALUE' % __name__)
+
+  Then in an entry point (main() function), use that parser as a parent:
+    parser = argparse.ArgumentParser(parents=[other_module.ARG_PARSER])
+    parser.parse_args()
+"""
 
 import argparse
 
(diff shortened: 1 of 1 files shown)


openhtf/util/argv.py
```diff
@@ -19,5 +35,7 @@ def __init__(self, *args, **kwargs):
     def __call__(self, parser, namespace, values, option_string=None):
         if hasattr(self, '_proxy'):
             values = self._proxy(parser, namespace, values)
-        setattr(__import__(self._tgt_mod), self._tgt_attr, values)
+        base, mod = self._tgt_mod.rsplit('.', 1)
```
jettisonjoe: This unpacking will break if the target module is already a base module without any dots, but arguably it doesn't make sense to use `StoreInModule` in such a case.

openhtf/util/argv.py
```diff
@@ -1,4 +1,20 @@
...
+    DEFAULT_VALUE = 0
+    ARG_PARSER = argv.ModuleParser()
+    ARG_PARSER.add_argument(
+        '--override-value', action=argv.StoreInModule,
+        default=DEFAULT_VALUE, target='%s.DEFAULT_VALUE' % __name__)
+
+  Then in an entry point (main() function), use that parser as a parent:
+    parser = argparse.ArgumentParser(parents=[other_module.ARG_PARSER])
+    parser.parse_args()
+"""
```
jettisonjoe: :+1:

openhtf/util/argv.py
grybmadsci: hm, yeah it's nonsensical to have no dots, but something one might accidentally do if you forget to do the %s/__name__ business, maybe throw in a quick format check on the inputs of `add_argument()` to fail-fast rather than raise an obscure error here?

grybmadsci:
LGTM, we'll get a merge conflict on station_api from the fix I threw in here, but farz's version is better, I think, mine was just a quick make-it-work fix:
https://github.com/google/openhtf/blob/station_api/openhtf/util/argv.py#L23

jettisonjoe:
I'm gonna merge this now so that the onus will be on me to resolve that conflict when rebasing station_api in prep for the PR to merge _it_ into master. Created #390 to get a better error message if someone sets an unsuitable target.

PR #391, opened 2016-07-28
Merge station_api branch into master.

jettisonjoe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/391)


grybmadsci:
LGTM, but can we add a tag to master's current HEAD before merging this in so we don't lose track of the "last version before PyPI launch"?  I forget what we were calling it, 0.9?  we should probably follow semantic versioning, so, 0.9.0 or somesuch.

jettisonjoe:
Sounds good. I'll add a 0.9.0 tag at master's current HEAD.

fahhem:
oh gawd my browser lags trying to look at the diffs. looking now

contrib/poll_stations.py
```diff
@@ -0,0 +1,205 @@
...
+import logging
+import os
+import Queue
+import socket
+import sys
+import threading
+import time
+
+# TODO(madsci): Set log level from flags.
+logging.basicConfig(level=logging.INFO)
```
fahhem: this should be under the `__main__` check

jettisonjoe:
No idea why travis tests are failing. They pass for me locally, and UOM_CODE_MAP has the key 'NDL' for me locally.

examples/all_the_things.py
```diff
@@ -89,25 +90,28 @@ def dimensions(test):
 
 
 def attachments(test):
-  test.Attach('test_attachment', 'This is test attachment data.')
-  test.AttachFromFile('example_attachment.txt')
+  test.attach('test_attachment', 'This is test attachment data.')
```
fahhem: We need to start a changelog :) This is going to break all my team's testers twice over

jettisonjoe: This is just the beginning; the pylint and restructuring PR's are going to require every test script to get a pretty thorough once-over. Sorry! :-\ I believe it's better to do this in one fell swoop now (just prior to publishing in PyPI) and have the API be fairly stable once we go live with v1.0 on PyPI.

jettisonjoe: I'm down to write a changelog, but I'd like you guys to vet what I did with the examples in the [restructuring PR (on my fork)](https://github.com/jettisonjoe/openhtf/pull/2) so I can make sure what I'm writing up is what we end up using.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py34
```
grybmadsci: did python3 support break, or are you just removing this because we don't officially support python3 currently?

fahhem:
Reviewed 2 of 17 files at r1, 10 of 27 files at r2, 14 of 21 files at r4, 51 of 51 files at r5, 1 of 6 files at r6, 7 of 7 files at r7.
Review status: all files reviewed at latest revision, 33 unresolved discussions.

---

*[openhtf/__init__.py, line 398 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6fiwD1cULo4eVNBsV:-KO6fiwD1cULo4eVNBsW:-725661834) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L398)):*
> ```Python
>     return new_info
> 
>   def __call__(self, test_state):
> ```

Can you document, at least in a comment, what test_state is supposed to be? Since the type isn't in this file it's not immediately obvious.

---

*[openhtf/__init__.py, line 421 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6iHyZOH6EO3-mb4_7:-KO6iHy_rJMvEwwz1ubz:1849133750) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L421)):*
> ```Python
> 
>   Attributes:
>     attachments: Dict mapping attachment name to test_record.Attachment
> ```

If we don't want this modified directly, then I'd put it in a new 'Readonly Attributes' section at the bottom so we can have test authors look at this docstring and generally ignore this attribute

---

*[openhtf/__init__.py, line 468 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6iaHsCvMUNxJUhJ1u:-KO6iaHsCvMUNxJUhJ1v:1090993537) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L468)):*
> ```Python
>     notify_update: Notify any frontends of an interesting update.  Typically
>         this is automatically called internally when interesting things happen,
>         but it can be called by the user (takes no args).
> ```

Why would they want to do this? What changes don't get to frontends immediately?

---

*[openhtf/exe/__init__.py, line 62 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO7wdfZeAaYY5peLAXA:-KO7wdfZeAaYY5peLAXB:556261643) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/__init__.py#L62)):*
> ```Python
>         teardown_function and
>         openhtf.PhaseDescriptor.WrapOrCopy(
>             teardown_function, timeout_s=conf.teardown_timeout_s))
> ```

We don't want to allow the teardown to specify how long it will take?

---

*[openhtf/exe/test_state.py, line 144 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8-XbshL32c4SJ1PoZ:-KO8-XbshL32c4SJ1Po_:-1616579701) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L144)):*
> ```Python
>       if output_record:
>         # Clear notification callbacks so we can serialize measurements.
>         for meas in self.running_phase_state.measurements.values():
> ```

measurements are messed with in `record_timing_context` in a weird way. It makes slightly more sense to be here? Then you'd be pulling them out of a phase and ripping out the notification callbacks to put it into the serializable `test_record` and there's at least some intuition there.

---

*[openhtf/exe/test_state.py, line 225 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO7zloK5anAcY0UCDW0:-KO7zloK5anAcY0UCDW1:2135157707) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L225)):*
> ```Python
>     return self.is_finalized
> 
>   def TestStarted(self, dut_id):
> ```

`MarkTestStarted` (or `mark_test_started`, or `SetTestStarted`)

---

*[openhtf/io/station_api.py, line 164 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO80mnInWQxxVLVWbwN:-KO80mnInWQxxVLVWbwO:2150811) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L164)):*
> ```Python
> 
> 
> class RemotePhase(collections.namedtuple('RemotePhase', [
> ```

Almost feels like we want a proto in here :)

Any chance we could have `PhaseState._asdict()` return a `RemotePhase` (or similarly named, maybe `SerializablePhase`) object? That would keep these files in sync better

---

*[openhtf/io/station_api.py, line 229 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO81burI8lkA_FgTFsj:-KO81bus7JxH27FZrUke:-886065735) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L229)):*
> ```Python
>     if not state or not state.test_record:
>       return 0, 0, state
>     return (len(state.test_record.phases),
> ```

namedtuple? This is used across modules so it needs it more than if it was intra-module

---

*[openhtf/io/station_api.py, line 234 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO81ui1593mxt0ShjZo:-KO81ui1593mxt0ShjZp:652990282) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L234)):*
> ```Python
> 
>   def __str__(self):
>     # Use cached_state because accessing self.state triggers an RPC, and
> ```

What is `self.state`? Is it `self.state_cache_info`?

---

*[openhtf/io/station_api.py, line 245 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO835rClVOpaiTncM_C:-KO835rClVOpaiTncM_D:-1369629250) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L245)):*
> ```Python
> 
>   @property
>   def cached_state_and_swapped_dict(self):
> ```

'swapped' is weird. It seems like you want a state summary. Maybe `cached_state_summary`?

---

*[openhtf/io/station_api.py, line 268 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO82I68hu4wudHRVt_n:-KO82I68hu4wudHRVt_o:631354925) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L268)):*
> if cached_state:
<details><summary><i alt="summary">Quoted 11 lines of code…</i></summary>
>       state = cached_state._replace(
>           test_record=mutablerecords.CopyRecord(cached_state.test_record))
>       state.test_record.phases = len(state.test_record.phases)
>       state.test_record.log_records = len(state.test_record.log_records)
>       swapped_dict = {
>           'status': state.status.name,
>           'running_phase_state':
>               data.ConvertToBaseTypes(state.running_phase_state),
>           'test_record': xmlrpclib.Binary(pickle.dumps(state.test_record)),
>       }
>     return cached_state, swapped_dict
</details>

Looks like all of this does this slightly simpler thing?

```py
if not cached_state:
  return None, None

test_record = mutablerecords.CopyRecord(cached_state.test_record)
test_record.phases = len(test_record.phases)
test_record.log_records = len(test_record.log_records)
return {
    'status': cached_state.status.name,
    'running_phase_state': data.ConvertToBaseTypes(cached_state.running_phase_state),
    'test_record': xmlrpclib.Binary(pickle.dumps(test_record)}
```

---

*[openhtf/io/station_api.py, line 362 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO88C9GbmhHWt97SrvE:-KO84-258reqWkS45dFJ:-931555406) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L362)):*
> ```Python
> 
>   def __new__(cls, host, station_api_port, station_info, proxy=None):
>     cls.STATION_LOCK.acquire()
> ```

I went and wrote a whole comment about what this could be replaced by, then after reading through to `from_host_port` I realized why you did it this way. I guess a comment here that says something to the effect of "Thread-safe way to ensure only one Station for this key is ever generated. Lock is released at the end of __init__()"


This would also save the extra `__new__` call when the `Station` is already there:
```py
  key = (host, station_api_port)
  return (cls.STATION_MAP.get(key) or
          cls.STATION_MAP.setdefault(key, super(Station, cls).__new__(cls))
```

---

*[openhtf/io/station_api.py, line 383 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO87akN26Z1FQD4o1QE:-KO87akN26Z1FQD4o1QF:2133720702) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L383)):*
> ```Python
>       self._history = history.History()
>       self._lock = threading.Lock()
>       self.cached_tests = {}
> ```

Looks like `cached_test` and `_history` should go out of the try-except?

---

*[openhtf/io/station_api.py, line 391 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO89GSgpdad2haBvWeX:-KO89GSgpdad2haBvWeY:-255691708) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L391)):*
> ```Python
>     # blocking short requests with long-running ones.
>     self._shared_proxy = proxy or self.make_proxy()
>     self.STATION_LOCK.release()
> ```

And a comment here that it's acquired in `__new__()`

---

*[openhtf/io/station_api.py, line 524 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO89c4zbmKCIQ26VqGo:-KO89c4zbmKCIQ26VqGp:469059680) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L524)):*
> ```Python
>     """Obtain dict required to make a StationInfo for this station."""
>     _LOG.debug('RPC:get_station_info() -> %s:%s', conf.station_id, self.UID)
>     return {
> ```

Again with the proto-itch :) Maybe namedtuples? All this over-the-wire synchronization feels brittle like this

---

*[openhtf/io/station_api.py, line 680 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8AFPb3Yxq1xWQAGF3:-KO8AFPb3Yxq1xWQAGF4:2056975682) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L680)):*
> ```Python
>     # remote_state_dict.
>     state_dict_counts = dict(state_dict)
>     state_dict_counts['test_record'] = mutablerecords.CopyRecord(
> ```

Isn't this the swapped_dict stuff from above?

---

*[openhtf/io/station_api.py, line 704 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8AYyWMrkEc7YLDLTi:-KO8AYyWMrkEc7YLDLTj:-141676606) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L704)):*
> ```Python
>             data.pprint_diff(remote_state_dict, state_dict_counts,
>                              'remote_state', 'local_state'))
>         _LOG.debug('\n'.join(log_msg))
> ```

if _LOG.isEnabledFor(logging.DEBUG):
  # Then build all that stuff to avoid wasting time when debugging isn't enabled

---

*[openhtf/io/station_api.py, line 721 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8APqNpz67nXSAoL2s:-KO8APqNpz67nXSAoL2t:-794420553) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L721)):*
> ```Python
>     # Grab a fresh copy of the state and return the new info.
>     state = openhtf.Test.state_by_uid(test_uid)
>     return state and self._serialize_state_dict(
> ```

This is a really long function, it could use getting split up

---

*[openhtf/io/station_api.py, line 798 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8O1A7sqCAiV7s1z8b:-KO8O1A7sqCAiV7s1z8c:-1169934558) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L798)):*
> ```Python
> 
> def start_server():
>   # TODO(madsci): Blech, fix this. Suggestions?
> ```

Why not make `API_SERVER` always an instance of `ApiServer`? It's `__init__()` doesn't do much so it's not that expensive.

---

*[openhtf/io/test_record.py, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8OkKLbIL4rn9DtE3N:-KO8OkKLbIL4rn9DtE3O:684796753) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/test_record.py#L49)):*
> ```Python
>   @property
>   def sha256(self):
>     return hashlib.sha256(self.data).hexdigest()
> ```

I'd propose using md5 since this isn't meant to be cryptographic and it's much faster, which can translate to noticeable times when there are several multi-MB attachments (such as camera stations)

---

*[openhtf/io/frontend/__init__.py, line 116 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8PJPJT-FxdORQx3Dy:-KO8PJPJT-FxdORQx3Dz:651604920) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L116)):*
> ```Python
>   def run(self):
>     self._do_callback_for_state(self._test.state)
>     while True:
> ```

There are so many new while-loops that are various amounts of tight. Have you checked the output of top while running both binaries?

---

*[openhtf/io/frontend/__init__.py, line 214 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8PkciqeKZ-cZknsO0:-KO8PkciqeKZ-cZknsO1:1578989709) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L214)):*
> ```Python
>                      message sent to them.
>     """
>     for client in cls.subscribers:
> ```

I think you'll want a lock here? If `cls.subscribers` changes length during iteration, Python will throw an error here

---

*[openhtf/io/frontend/__init__.py, line 363 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8Q8_Pi6akARuU_vmt:-KO8Q8_Pi6akARuU_vmu:86208106) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L363)):*
> ```Python
>     def _make_sockjs_url(rel_url):
>       """SockJS URL is /plugs/<host>/<port>/<test_uid>/<plug_name>"""
>       return r'/plugs/%s/%s/%s/%s' % (hostport + (test_uid, rel_url))
> ```

Shouldn't '/plugs/' be registered in the call to `Application.__init__()`?

---

*[openhtf/io/frontend/src/app/station.html, line 18 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8REuWqY9NNBXro910:-KO8REuWqY9NNBXro911:1314041848) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.html#L18)):*
> ```HTML
>           [currentMillis]="currentMillis"
>           [status]="test.test_record?.outcome
>                     || test.status || 'OFFLINE'">
> ```

Should fit on one line

---

*[openhtf/io/frontend/src/app/station.html, line 34 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8RHat9l1pSx2JVj0s:-KO8RHat9l1pSx2JVj0t:-480160692) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.html#L34)):*
> ```HTML
>   <div id="history">
>     <div class="big-message">
>       History feature coming soon!
> ```

Yay!!

---

*[openhtf/io/frontend/src/app/station.logs.ts, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8RMgtLusPJj-sFNog:-KO8RMgtLusPJj-sFNoh:166874) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.logs.ts#L49)):*
> ```XML
> 
>   ngOnInit() {
>     // $('ul.tabs').tabs();
> ```

?

---

*[openhtf/io/frontend/src/app/station.metadata.ts, line 16 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8RgdCxEfNWUueTkrL:-KO8RgdCxEfNWUueTkrM:-1778374627) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.metadata.ts#L16)):*
> ```XML
> 
> 
> declare var $;  // Global provided by the jquery package.
> ```

This shouldn't be necessary if you remove the commented line at the bottom

---

*[openhtf/io/frontend/src/app/station.phases.html, line 74 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8SyUZ_WTO80wAXAX3:-KO8SyU_U12TB8ICHo52:1886687977) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.phases.html#L74)):*
> ```HTML
>       <div class="measurement-container"
>            [innerHTML]="state.running_phase_state?.measurements | measToHtml">
>

tox.ini
fahhem: I don't think python3 ever worked. I just added it so that we could test it over time, but we're pretty far from it at the moment since we don't use any compat libraries like `six`

fahhem:
Sorry for taking so long, but I did a full review using reviewable (it's pretty dang awesome) which means the comments don't get sent until I hit publish at the end unlike github's method.

It looks like reviewable was made by an xoogler (even mentioned in a testimonial) since it acts suspiciously like our tools :)

grybmadsci:
LGTM, we should plan on both a comprehensive changelog and a published release candidate for people to try, I expect getting some run time will flush out more bugs that are hard to catch in review.

jettisonjoe:
@grybmadsci, I've been thinking about sub-minor versions too like betas, release candidates, etc. but I haven't come up with a good plan for where/how to publish them. If we publish to PyPI it's an actual release rather than a candidate.  If we want to take advantage of travis-ci's auto-publish-to-PyPI features, we can't use GitHub releases for rc's because they'll be auto-published to PyPI, too. GitHub _does_ have a pre-release checkbox, but I think even pre-releases will create tags, which will trigger travis-ci unless we disable the automatic stuff and have a manual process for uploading new releases to PyPI.

jettisonjoe:
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[openhtf/__init__.py, line 468 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6iaHsCvMUNxJUhJ1u:-KOLmKhzl59_O3bOwTEV:1426731943) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L468)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Why would they want to do this? What changes don't get to frontends immediately?
</details>
I think this was put in to support the concept of frontend plugins. We already have some in progress for interfacing with various data systems for history, and we plan to have some for plugs, too.  If you were using openhtf alongside another framework that did test scheduling and machine management, for instance, you might have a frontend plugin to display and interact with stuff at that one-up layer, and you might want to trigger updates to that interface in phases.

---

*[openhtf/exe/__init__.py, line 62 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO7wdfZeAaYY5peLAXA:-KOLrYNdtJV1R5KqpEF2:351093057) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/__init__.py#L62)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> We don't want to allow the teardown to specify how long it will take?
</details>
The reason we're enforcing a timeout on teardown to begin with is to get reasonably fast shutdown on SIGINT. The person likely to care about tweaking this behavior is going to be the test runner, not necessarily always the same person as the test author. Hooking it up to a command line flag lets you do stuff like have super-fast kills during development and debugging, but go back to the full timeout needed to guarantee proper teardown in production, all without modifying the test script.

That said, @grybmadsci was the one to write this logic so I'd like to hear what he has to say on the matter.

---

*[openhtf/exe/test_state.py, line 144 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8-XbshL32c4SJ1PoZ:-KOLw0JrsX_XXvKMP6SE:1103906892) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L144)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> measurements are messed with in `record_timing_context` in a weird way. It makes slightly more sense to be here? Then you'd be pulling them out of a phase and ripping out the notification callbacks to put it into the serializable `test_record` and there's at least some intuition there.
</details>
I guess that depends whether you think multi-dim measurement validation should count toward phase cycle time or not. Personally I think it probably should, especially since normal measurement validation does.

---

*[openhtf/io/station_api.py, line 164 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO80mnInWQxxVLVWbwN:-KOM355KWbuLVyUiL6jd:252264999) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L164)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Almost feels like we want a proto in here :)
> 
> Any chance we could have `PhaseState._asdict()` return a `RemotePhase` (or similarly named, maybe `SerializablePhase`) object? That would keep these files in sync better
</details>
@grybmadsci?

---

*[openhtf/io/test_record.py, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8OkKLbIL4rn9DtE3N:-KOM40n4b4Gvhl2TeNe_:-699500039) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/test_record.py#L49)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> I'd propose using md5 since this isn't meant to be cryptographic and it's much faster, which can translate to noticeable times when there are several multi-MB attachments (such as camera stations)
</details>
Sounds fine to me. @grybmadsci, you were the one who chose sha256. Any objections to md5?

---

*[openhtf/io/frontend/__init__.py, line 363 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8Q8_Pi6akARuU_vmt:-KOMFnnGll__L4yM4SiT:-933101337) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L363)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Shouldn't '/plugs/' be registered in the call to `Application.__init__()`?
</details>
At initialization, we can't yet know which plugs are going to be available and need handlers.

---

*[openhtf/io/frontend/src/app/station.logs.ts, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8RMgtLusPJj-sFNog:-KOMGdTRWkFGxJHdtxGd:-488984935) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.logs.ts#L49)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> ?
</details>
Oops!

---

*[openhtf/io/frontend/src/app/station.phases.html, line 74 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8SyUZ_WTO80wAXAX3:-KOMHDZpSDkQ7S62Rphy:-786840402) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/src/app/station.phases.html#L74)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Want to remove this now?
</details>
Sure. I was keeping it around for reference because I want to make these look nicer in a future change, but I can always look back into the code history for that.

---

*[openhtf/util/argv.py, line 20 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8WB7TK77Kgbv4odH-:-KOMJmWC-AIGZmj-alqs:1428267931) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/util/argv.py#L20)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> ?
</details>
No clue.

---

*[openhtf/util/functions.py, line 50 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8WaD1jKne5vjTrlt3:-KOMK-ecXaN8-Ek9HMnM:-989796615) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/util/functions.py#L50)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> invocations
</details>
But how will people know it works in KDE?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

jettisonjoe:
I think I've addressed everything except the stuff @grybmadsci should weigh in on.

jettisonjoe:
Oh, and the performance thing. Let me try that out.

jettisonjoe:
Review status: 73 of 84 files reviewed at latest revision, 32 unresolved discussions.

---

*[openhtf/io/frontend/__init__.py, line 116 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8PJPJT-FxdORQx3Dy:-KOMRu5T3tPyeKQk7v7R:-217348394) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L116)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> There are so many new while-loops that are various amounts of tight. Have you checked the output of top while running both binaries?
</details>
Okay, this is not my fault! The while loop itself wouldn't be an issue if the station_api wasn't short-circuiting its wait on every call. From what I've gathered so far, it looks like some stuff isn't propagating correctly to the remote state. In particular:

  * The `plugs` dict.
  * `test_name` in the test's metadata.

The station API is finding the same diff in these every `wait_for_update` call, which makes the loop on the remote side run way tighter than it should be.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

grybmadsci:
One thing I'd like to propose we do before PyPI launch is switch prompts over to a plug that is a thin layer to the current mechanism, that way the user-facing API doesn't have to change when we switch it over to be a real plug under the hood.

---

Review status: 73 of 84 files reviewed at latest revision, 34 unresolved discussions.

---

*[openhtf/__init__.py, line 398 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6fiwD1cULo4eVNBsV:-KO9DEK2LMx94kPtMK3t:-1233459873) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L398)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Can you document, at least in a comment, what test_state is supposed to be? Since the type isn't in this file it's not immediately obvious.
</details>
sorry, lots of my code is missing docstrings, joe you can copypasta this:
```
"""Invoke this Phase, passing in the appropriate args.

By default, an openhtf.TestApi is passed as the first positional arg, but if the 'requires_state'
option is set, then a test_state.TestState is passed instead.  If no positional args are expected,
then neither is passed in.  In any case, keyword args are passed in based on extra_kwargs, set
via WithArgs(), combined with plugs (plugs override extra_kwargs).

Args:
  test_state: test_state.TestState for the currently executing Test.

Returns:
  The return value from calling the underlying function.
"""
```

---

*[openhtf/__init__.py, line 468 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO6iaHsCvMUNxJUhJ1u:-KO9FE25kUc45I2r1fIJ:1723759112) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/__init__.py#L468)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Why would they want to do this? What changes don't get to frontends immediately?
</details>
any direct updates to the test_record would not be propagated.  since that's discouraged anyway, this is not likely to be used directly by users, might be worth making a note to that effect here.

---

*[openhtf/exe/__init__.py, line 62 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO7wdfZeAaYY5peLAXA:-KO9FoCeNUFSn-xzdOhu:667693697) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/__init__.py#L62)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> We don't want to allow the teardown to specify how long it will take?
</details>
i'd be okay with not overriding a timeout if one was set explicitly, but we do want this to default to be much shorter than the default phase timeout value.

---

*[openhtf/exe/test_state.py, line 144 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8-XbshL32c4SJ1PoZ:-KO9Hyo-vyMUOORUlmcj:-1490801809) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L144)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> measurements are messed with in `record_timing_context` in a weird way. It makes slightly more sense to be here? Then you'd be pulling them out of a phase and ripping out the notification callbacks to put it into the serializable `test_record` and there's at least some intuition there.
</details>
hm, measurements are fundamentally attached to phases, not the test as a whole, so i think it makes more sense to have the validation business in `PhaseState.record_timing_context` than here in `TestState`, but it is weird to have this call to clear the notification callback here. you could move this into the end of `record_timing_context`, once the measurements are copied to the `PhaseRecord` it no longer makes sense to have notification callbacks for them (regardless of `output_record`).  what do you think, farz?

---

*[openhtf/io/station_api.py, line 164 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO80mnInWQxxVLVWbwN:-KO9J1vePKq_XvseYskf:-500978801) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L164)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Almost feels like we want a proto in here :)
> 
> Any chance we could have `PhaseState._asdict()` return a `RemotePhase` (or similarly named, maybe `SerializablePhase`) object? That would keep these files in sync better
</details>
`_asdict()` must return a base type for `xmlrpclib` to know how to marshal it.  mirroring `namedtuple._asdict()` and then reconstructing with `**kwargs` was the prettiest way to get visual alignment between those two sides that I could think of.  we could probably do a bit better, but as it's entirely internal, I'm not sure it's worth spending too much (more) time on it.  xmlrpc is somewhat limited.  in retrospect, i might have gone with something a bit more featureful in terms of serialization ability.

i'd be okay with creating an issue to consider revamping `station_api` to cleanup this kind of stuff, but I'd rather not block the merge on it because it's a bit of a can of worms.

---

*[openhtf/io/station_api.py, line 210 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO9KFX8tusx4duqNAKX:-KO9KFX8tusx4duqNAKY:616137909) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L210)):*
> ```Python
>     # Timestamps (in milliseconds) that we care about.
>     'created_time_millis', 'last_run_time_millis'], {
>     # Cache last known state so we can detect deltas.
> ```

might be worth updating this comment to reflect the type of `cached_state` to be `RemoteState`.

---

*[openhtf/io/station_api.py, line 229 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO81burI8lkA_FgTFsj:-KO9K3EgJcA0LYABRpc3:529055862) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L229)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> namedtuple? This is used across modules so it needs it more than if it was intra-module
</details>
agreed, this was sloppy of me, good catch.

---

*[openhtf/io/station_api.py, line 245 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO835rClVOpaiTncM_C:-KO9KS852F7W-JGrxAcE:1953734743) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L245)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> 'swapped' is weird. It seems like you want a state summary. Maybe `cached_state_summary`?
</details>
+1, "summary" is better nomenclature than "swapped".

---

*[openhtf/io/station_api.py, line 268 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO82I68hu4wudHRVt_n:-KO9Kim7h19XKTbSt3Kw:-1975564250) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L268)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> > if cached_state:
> >       state = cached_state._replace(
> >           test_record=mutablerecords.CopyRecord(cached_state.test_record))
> >       state.test_record.phases = len(state.test_record.phases)
> >       state.test_record.log_records = len(state.test_record.log_records)
> >       swapped_dict = {
> >           'status': state.status.name,
> >           'running_phase_state':
> >               data.ConvertToBaseTypes(state.running_phase_state),
> >           'test_record': xmlrpclib.Binary(pickle.dumps(state.test_record)),
> >       }
> >     return cached_state, swapped_dict
> 
> Looks like all of this does this slightly simpler thing?
> 
> ```py
> if not cached_state:
>   return None, None
> 
> test_record = mutablerecords.CopyRecord(cached_state.test_record)
> test_record.phases = len(test_record.phases)
> test_record.log_records = len(test_record.log_records)
> return {
>     'status': cached_state.status.name,
>     'running_phase_state': data.ConvertToBaseTypes(cached_state.running_phase_state),
>     'test_record': xmlrpclib.Binary(pickle.dumps(test_record)}
> ```
</details>
if `mutablerecords.CopyRecord()` supports the override kwargs by now, you can use that to override the `phases` and `log_records` attributes inline now too.  don't forget to `return cached_state, {...}` - farz missed the `cached_state` bit, but otherwise nice refactor.

---

*[openhtf/io/station_api.py, line 383 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO87akN26Z1FQD4o1QE:-KO9LZZP4GL9d1x7AWcq:-836235046) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L383)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Looks like `cached_test` and `_history` should go out of the try-except?
</details>
if `self._station_info == station_info`, we don't want to clear `_history` or `cached_tests`, so the clear needs to happen only in the event of a mismatch, or an `AttributeError`, which indicates they have never been set.

---

*[openhtf/io/station_api.py, line 391 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO89GSgpdad2haBvWeX:-KO9Lqu_DP21gYGMWfz7:-314584797) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L391)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> And a comment here that it's acquired in `__new__()`
</details>
lol good call, this was probably super annoying to figure out what was going on, sorry x.x

---

*[openhtf/io/station_api.py, line 524 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO89c4zbmKCIQ26VqGo:-KO9M-501-LjTDik9FvC:-659275973) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L524)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Again with the proto-itch :) Maybe namedtuples? All this over-the-wire synchronization feels brittle like this
</details>
same comment, must be a base type.  perhaps a comment indicating the namedtuple type above to look at for where these keys came from?

---

*[openhtf/io/station_api.py, line 680 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8AFPb3Yxq1xWQAGF3:-KO9MNVM8QFqvGxzHHhj:-741655337) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L680)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Isn't this the swapped_dict stuff from above?
</details>
yep, this is server-side, we need to do the same swap for the comparison below between `remote_state_dict` and `state_dict_counts` (probably better named `state_dict_summary`)

---

*[openhtf/io/station_api.py, line 721 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8APqNpz67nXSAoL2s:-KO9MgLDdoZmtYkwtVqM:2068871385) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L721)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> This is a really long function, it could use getting split up
</details>
make sure helper functions are prefixed with an underscore, or else the xmlrpc magic will export them.

---

*[openhtf/io/station_api.py, line 798 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8O1A7sqCAiV7s1z8b:-KO9MrMZepicrr1uRipO:-1290189215) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L798)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> Why not make `API_SERVER` always an instance of `ApiServer`? It's `__init__()` doesn't do much so it's not that expensive.
</details>
that's a thought, and then it only gets `.start()`ed if these conf values are set.  SGTM

---

*[openhtf/io/test_record.py, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8OkKLbIL4rn9DtE3N:-KO9N6Xc3S-H2KmzZZHP:-1856978179) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/test_record.py#L49)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> I'd propose using md5 since this isn't meant to be cryptographic and it's much faster, which can translate to noticeable times when there are several multi-MB attachments (such as camera stations)
</details>
if you can get bst to agree to it, then sounds fine to me ;)

---

*[openhtf/io/frontend/__init__.py, line 116 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8PJPJT-FxdORQx3Dy:-KO9NJcqYkYeNBEZRbah:-1711776350) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L116)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> There are so many new while-loops that are various amounts of tight. Have you checked the output of top while running both binaries?
</details>
`wait_for_update()` is blocking, so this isn't really a tight loop, per se.  nevertheless, a peek at top isn't a bad idea.

---

*[openhtf/plugs/__init__.py, line 238 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8UmodKqkrkE9wKj83:-KO9Oc1dFD91YAbRvKhf:-1259590926) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/plugs/__init__.py#L238)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> plug_name is created by splitting on '.', so 'plugs.' shouldn't be in plug_name?
</details>
right, I think I was doing `rsplit('.', 1)` before.  this particular bit of code hasn't been tested at all; it's getting a makeover for v1.1

---

*[openhtf/plugs/__init__.py, line 398 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8VfMIfOq3D9V99tYI:-KO9P-pEEaSUFh-5_Hhc:-1477686127) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/plugs/__init__.py#L398)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> This is only because the server isn't stopped/restarted after `InitializePlugs()`, right? This would otherwise be safe before that gets called?
</details>
right, but it's not really a thing that's intended to be done anyway, so it didn't seem worth making it work outside the context of unittests.  you can file a bug if you'd like to get this updated to play nicer, don't want to block joe's merge on it though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391#-:-KO9PPvBjE-tvfhWaLyU:-2047716260)*

grybmadsci:
er, had comments sitting in drafts, sorry!  some may be out of date :(

---

Review status: 73 of 84 files reviewed at latest revision, 33 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391#-:-KOMcZDXO2r9DdfI2qxF:511810367)*

grybmadsci:
Review status: 72 of 84 files reviewed at latest revision, 33 unresolved discussions.

---

*[openhtf/io/test_record.py, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8OkKLbIL4rn9DtE3N:-KOPQFlt3y9dEVuA_iD3:b2arnna) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/test_record.py#L49)):*
<details><summary><i>Previously, grybmadsci wrote…</i></summary>
> if you can get bst to agree to it, then sounds fine to me ;)
</details>
i'd be okay with SHA-1 instead of SHA-256, but MD5 is kind of a nono these days.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

jettisonjoe:
@grybmadsci, on prompt as a thin plug, where do you think the plug should go? In the top level of `plugs/`, or start a new `general/` subdir of plugs or something? Plug organization is something we haven't really discussed deeply yet, but maybe we should take a shot at it now. I would propose vendor directories at the top level for test equipment, but what about general comms stuff like USB, serial, etc? And even more general stuff like user input? Any thoughts?

---

Review status: 72 of 84 files reviewed at latest revision, 33 unresolved discussions.

---

*[openhtf/io/station_api.py, line 229 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO81burI8lkA_FgTFsj:-KOMyNNwE6OP5SyUSpjX:1995635051) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L229)):*
<details><summary><i>Previously, grybmadsci wrote…</i></summary>
> agreed, this was sloppy of me, good catch.
</details>
Fixing. Let me know if this looks like what you had in mind.

---

*[openhtf/io/station_api.py, line 234 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO81ui1593mxt0ShjZo:-KOMyxrrIMUSE0FZNt8l:-2125167762) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/station_api.py#L234)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> What is `self.state`? Is it `self.state_cache_info`?
</details>
A property that updates `self.cached_state` via RPC then returns it.

---

*[openhtf/io/frontend/__init__.py, line 116 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8PJPJT-FxdORQx3Dy:-KORH6SuAMv8PVKR-_YM:567711827) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/frontend/__init__.py#L116)):*
<details><summary><i>Previously, grybmadsci wrote…</i></summary>
> `wait_for_update()` is blocking, so this isn't really a tight loop, per se.  nevertheless, a peek at top isn't a bad idea.
</details>
This is fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391#-:-KOS1BLQM3bf2FPCTLgl:-618702874)*

jettisonjoe:
In other news, I've addressed pretty much all of the reviewer feedback except the addition of a `user_input` plug. PTAL.

---

Review status: 72 of 84 files reviewed at latest revision, 33 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391#-:-KOS2jlswCg6CXQci5d4:-2038544442)*

fahhem:
:lgtm:

---

Reviewed 1 of 1 files at r8, 10 of 10 files at r9, 2 of 2 files at r10, 6 of 6 files at r11.
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[openhtf/exe/test_state.py, line 144 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8-XbshL32c4SJ1PoZ:-KOTkeCzyRONr-NMSat7:b1gtva9) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L144)):*
<details><summary><i>Previously, grybmadsci wrote…</i></summary>
> hm, measurements are fundamentally attached to phases, not the test as a whole, so i think it makes more sense to have the validation business in `PhaseState.record_timing_context` than here in `TestState`, but it is weird to have this call to clear the notification callback here. you could move this into the end of `record_timing_context`, once the measurements are copied to the `PhaseRecord` it no longer makes sense to have notification callbacks for them (regardless of `output_record`).  what do you think, farz?
</details>
I guess my main problem is the whole 'timing context' thing doing measurement stuff too. I won't block this PR for this, I'll probably mess around with the code after it's in and see if I come up with anything PR worthy

---

*[openhtf/io/station_api.py, line 639 \[r11\]](https://reviewable.io:443/reviews/google/openhtf/391#-KOThg4H1bk2124oPnoy:-KOThg4IFPgn811R9-ms:b-mho4k3) ([raw file](https://github.com/google/openhtf/blob/35030dd997ffd6895a14125f5f051c8390ced45b/openhtf/io/station_api.py#L639)):*
> ```Python
>     state_dict, event = state.asdict_with_event()
>     
>     state_dict_summary = {
> ```

This might be simpler as:

```py
state_dict_summary = dict(state_dict)
state_dict_summary.pop('plugs', None)
``` 

---

*[openhtf/io/test_record.py, line 49 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8OkKLbIL4rn9DtE3N:-KOTi551-qifdQ4RmLxh:b-8kq8y1) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/io/test_record.py#L49)):*
<details><summary><i>Previously, grybmadsci wrote…</i></summary>
> i'd be okay with SHA-1 instead of SHA-256, but MD5 is kind of a nono these days.
</details>
MD5 is a nono for cryptographic uses. This is nowhere near cryptographic, we just need a semi-reasonable hash that doesn't collide across a few dozen entries at most. Anyway, sha1 is better than 256 and it's not worth much more debate so I'll leave it alone.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391#-:-KOTk63fXN5jDMGkakP5:bnfp4nl)*

jettisonjoe:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[examples/all_the_things.py, line 93 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KNsHI6TagqSKh-GhDJQ-r7-93:-KOfhN8IW0VS7aMH-6WX:befr6hc) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/examples/all_the_things.py#L93)):*
<details><summary><i>Previously, jettisonjoe (Joe Ethier) wrote…</i></summary>
> I'm down to write a changelog, but I'd like you guys to vet what I did with the examples in the [restructuring PR (on my fork)](https://github.com/jettisonjoe/openhtf/pull/2) so I can make sure what I'm writing up is what we end up using.
</details>
@grybmadsci, I added user_input as a plug. In this PR it's a copy/paste of the existing stuff, plus a wafer-thin plug wrapper. I updated the example tests to use this. If this looks okay to you I'll remove the user_input module from util.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

grybmadsci:
Review status: 83 of 86 files reviewed at latest revision, 21 unresolved discussions.

---

*[openhtf/exe/test_state.py, line 144 \[r7\]](https://reviewable.io:443/reviews/google/openhtf/391#-KO8-XbshL32c4SJ1PoZ:-KOhXja6oJP3O7zEnFdP:bwgq7tj) ([raw file](https://github.com/google/openhtf/blob/9c065db61c2f48cfae65a00ff26578eba352971a/openhtf/exe/test_state.py#L144)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary>
> I guess my main problem is the whole 'timing context' thing doing measurement stuff too. I won't block this PR for this, I'll probably mess around with the code after it's in and see if I come up with anything PR worthy
</details>
ah, sure, i'd be okay with some renaming of things for sure, and potentially moving around as long as we don't lose obvious the attachment of measurements to phases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

grybmadsci:
LGTM, that's exactly what I had in mind for the `UserInput` plug.  No thoughts really on directory layout, you have a better eye for that sort of thing anyway.

fahhem:
Reviewed 3 of 3 files at r12.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/391)*

fahhem:
Looks like reviewable doesn't put the LGTM stamp here. This is ready to be merged from my perspective

grybmadsci:
LGTM.

jettisonjoe:
Awesome, thanks, for the review, guys!

PR #392, opened 2016-07-29
Skip non-supported units in mfg_inspector

fahhem:


jettisonjoe:
There's a second place you probably want to do this on line 235.

fahhem:
Thanks! PTAL

jettisonjoe:
LGTM.

PR #393, opened 2016-07-29
Support non-dotted modules in argv.StoreInModule

### LINKED ISSUE #310: Document util/argv.py
There are no docs on how to use this module.

### LINKED ISSUE #390: Nicer failure mode for misuse of StoreInModule.
See #389.

> ...it's nonsensical to have no dots, but something one might accidentally do if you forget to do the %s/name business, maybe throw in a quick format check on the inputs of add_argument() to fail-fast rather than raise an obscure error here?
@grybmadsci

fahhem:
Fixes #390 

Fixed #310 in the previous PR


jettisonjoe:
LGTM.

PR #395, opened 2016-07-29
Allow @test.patch_plugs to accept a plug class directly

fahhem:
Thoughts?


grybmadsci:
Good idea, I forget why I didn't do it this way in the first place, actually.. Could you also update the docstring with an example of this usage?

fahhem:
done, PTAL

grybmadsci:
don't forget to update the unittest :)

PR #400, opened 2016-08-15
Plugin

maddychan:
Add plugin interface


openhtf/output/web_gui/__init__.py
```diff
@@ -315,7 +316,7 @@ def initialize(self, port):  # pylint: disable=arguments-differ
     def get(self):
       self.render('index.html', host=socket.gethostname(), port=self.port)
 
-  def __init__(self, discovery_interval_s, disable_discovery, http_port,
+  def __init__(self, discovery_interval_s, disable_discovery, http_port, plugins = [],
```
fahhem: plugins should default to None. See https://google.github.io/styleguide/pyguide.html?showone=Default_Argument_Values#Default_Argument_Values

openhtf/output/web_gui/__init__.py
```diff
@@ -331,9 +332,37 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
         (r'/', self.MainHandler, {'port': http_port}),
         (r'/station/(?:(?:[0-9]{1,3}\.){3}[0-9]{1,3})/(?:[0-9]{1,5})(?:/*)',
          self.MainHandler, {'port': http_port}),
-        (r'/(.*\..*)', tornado.web.StaticFileHandler, {'path': path}),
-        (r'/(styles\.css)', tornado.web.StaticFileHandler, {'path': path}),
+        (r'/(.*\..*)', tornado.web.StaticFileHandler, {'path': path})
     ] + dash_router.urls + station_router.urls
+
+    #Add plugins into handler_routes
+    if type(plugins) == list:
```
fahhem: Once you make plugins default to None, you can do this:

```py
for plugin in list(plugins or []):
  pass
```

openhtf/output/web_gui/__init__.py
```diff
@@ -331,9 +332,37 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
...
+              dep_dict['port'] = http_port
+            elif dep == 'path':
+              dep_dict['path'] = path
+            else: 
+              print("Unable to find dependancy "+dep)
+
+          #Add completed route to handler_routes
+          handler.append(dep_dict)
+          handler_routes.append(tuple(handler))
+        except:
```
fahhem: * Put the try-except around the minimal amount of code you [expect to fail]. It looks like you should only put it around the first two lines (`import_module` and `plugin_module.handler` may fail)
* Catch more specific exceptions, such as `ImportError` and `AttributeError`
* Use _LOG.exception to log your reasoning for the error as well as the traceback itself

[expect to fail]: https://google.github.io/styleguide/pyguide.html?showone=Exceptions#Exceptions

openhtf/output/web_gui/__init__.py
```diff
@@ -331,9 +332,37 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
...
+    if type(plugins) == list:
+      for name in plugins:
+        try:
+          #Import plugin package
+          plugin_module = importlib.import_module(name)
+          handler,deps = plugin_module.handler
+
+          #Add in handler dependencies 
+          dep_dict = {}
+          for dep in deps:
```
fahhem: Why not pass in a `namedtuple` with all of these attributes? Then they only have to supply the handler itself

openhtf/output/web_gui/src/README.md
```diff
@@ -0,0 +1,134 @@
...
+	'label':'example', 
+	'icon':'pets'}
+```
+- label: The label for the tab for your plugin
+- icon: Optional, but refers to the icon from Material Icons that you want
+on your tab. (replace any spaces with underscores)
+
+#### Tornado Handler
+
+If you need to pull information from the backend, then you need to create a
```
fahhem: This is the handler mentioned in the variable above? If it's optional, I'd mention that above and combine this section with that one since it's confusing to separate them.

openhtf/output/web_gui/src/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Everything adding to the backend goes in the __init__.py module! It requires 
+three main parts: 
+
+#### Handler Variable
+
+This tuple is what OpenHTF will use to configure and add your tornado handler. 
+A simple example: 
+```
+handler=([r'/example/', ExampleHandler],
```
fahhem: Lists imply that it can be lengthened. For the deps portion, it makes sense since it can be any length, but for the regex/handler portion it should be a tuple. If we remove the deps like suggested, this could allow multiple handlers like so:

```py
handlers = [
  (r'/example/$', ExampleHandler),
  (r'/example/(\d+)/', IntegerHandler),
]
```

openhtf/output/web_gui/src/README.md
```diff
@@ -0,0 +1,134 @@
...
+- For references to styleUrls and templateUrl in the Component wrapper, use the 
+url 'plugin/(NAME OF PACKAGE)/(NAME OF FILE)', 
+i.e, '/plugin/example_plugin/example_plugin.html'
+- For imports of html or css scripts, use urls relative to the plugin folder 
+in your package
+
+## Building the Plugin
+
+Once the contents are finished, it needs to be added to OpenHTF and then built 
+before it can be used. Note: If you are using a virtualenv, this requires npm
```
fahhem: Is this a hard requirement? Is there any chance we can make the plugin provide the data at runtime?

openhtf/output/web_gui/src/app/station.testheader.ts
```diff
@@ -61,6 +68,17 @@ export class TestHeader implements OnChanges {
   }
 
   /**
+   * Propogates star toggle event back to HistoryHeader component.
```
fahhem: there's a lot of unrelated diffs in this commit. Any chance it was merged with another one on accident?

openhtf/output/web_gui/src/package.json~
```diff
@@ -0,0 +1,46 @@
+{
```
fahhem: backup file, please remove

openhtf/output/web_gui/src/update_plugins.py
```diff
@@ -0,0 +1,248 @@
...
+import {PhaseListing} from './station.phases';
+import {Prompt} from './station.prompt';
+import {StationService} from './station.service';
+import {TestHeader} from './station.testheader';
+import {Countdown, ObjectToArray} from './utils';
+import 'file?name=/styles/station.css!./station.css';
+import 'file?name=/templates/station.html!./station.html';
+"""
+
+mid1_ts='''
```
fahhem: Instead of cutting this up in such an odd location (directives opens but doesn't close the list), put a format directive. Also, use `"""` when possible

```py
mid1_ts = """
@Component({
  selector: 'station',
  ...
  directives: [StationHeader, Prompt, ..., Metadata, {extra_directives}],
  providers: [StationService]
})
export class ...
"""
```

openhtf/output/web_gui/src/update_plugins.py
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Create a Station view component.
+   */
+  constructor(private routeParams: RouteParams,
+              private stationService: StationService) {
+    this.currentMillis = setInterval(() => {
+      this.currentMillis = Date.now();
+    }, 250);
+    //This line is automatically generated, can be very long
+    this.plugins='''
```
fahhem: same as above here, `this.plugins` just has an awkwards `=` here now

openhtf/output/web_gui/src/app/station.testheader.ts
maddychan: The changes were probably to accommodate the plugin I wrote to interface with Scout, but since that's an internal tool I couldn't include the plugin itself in this PR.

openhtf/output/web_gui/src/README.md
maddychan: It's a hard requirement right now, but we're willing to consider any suggestions if you have any for doing it in run time!

maddychan:
I did author the commits, googlebot!

openhtf/output/web_gui/src/update_plugins.py
```diff
@@ -99,20 +99,18 @@
 import 'file?name=/templates/station.html!./station.html';
 """
 
-mid1_ts='''
+post_ts="""
 
 /** Station view component. **/
-@Component({
+@Component({{
```
fahhem: Oh, if you used `%(plugin_configs)s` you wouldn't need to double up all these curly braces, sorry!

fahhem:
Can you rebase on master or specify what this should be based on so that I can see only those changes? It's difficult to review this as-is

PR #405, opened 2016-08-18
Fix a bug in mfg-inspector output.

### LINKED ISSUE #404: Mfg inspector output module is adding unset values to failureCodesList
Mfg inspector output module is adding unset values to failureCodesList with details "None".  This is giving incorrect failure codes in mfg inspector.  Older openhtf commit I was running would just skip these.

"failureCodesList": [
    {
      "code": "NoDeviceError",
      "details": "No device found"
    },
    {
      "code": "foo",
      "details": "None"
    },
    {
      "code": "bar",
      "details": "None"
    },

jettisonjoe:
Specifically, prevent the text_value field on measurements from being set when
the measurement itself remains UNSET.

Fixes #404


openhtf/io/output/mfg_inspector.py
```diff
@@ -239,7 +239,7 @@ def _ExtractParameters(record, testrun, used_parameter_names):
         # Just a plain ol' value.
         if isinstance(value, numbers.Number):
           testrun_param.numeric_value = float(value)
-        else:
+        elif value is not None:
```
fahhem: Actually, I'd check `measurement.measured_value.is_value_set` instead, since if the parameter is actually set to None we want that to be an error. We only want purely unset ones to show up with no value set.

Also, the bug this PR references is that the unset measurements show up as a failure code, which means the real bug is that the code up above checking if the measurement has a `measured_value` isn't working. It should only get to here if it was actually set to None.

I'd say that `measurement.measured_value.is_value_set` should be checked rather than `measurement.measured_value` since the `measured_value` will always be there and neither `MeasuredValue` nor `DimensionedMeasuredValue` have `__nonzero__` implemented.

jettisonjoe:
Made some updates. PTAL.

jettisonjoe:
Tests might fail right now because I didn't update test output files yet. I'll do that once we settle on the right fix.

wallacbe:
Tried running this, I'm seeing FAIL instead of UNSET in inspector

wallacbe:
Also I see the following as a failure code.  Before we would just see "IOError"
ExceptionInfo - ExceptionInfo(exc_type=<type 'exceptions.IOError'>, exc_val=IOError(), exc_tb=<traceback object at 0x7f35f50a8f38>)

wallacbe:
One other observation, the unset measurement shows up in the failure code list in mfg-inspector

All failure codes
ExceptionInfo - ExceptionInfo(exc_type=<type 'exceptions.IOError'>, exc_val=IOError(), exc_tb=<traceback object at 0x7f35f50a8f38>)
hello_world_measurement - Unset

openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -238,7 +238,7 @@ def _extract_parameters(record, testrun, used_parameter_names):
         # Just a plain ol' value.
         if isinstance(value, numbers.Number):
           testrun_param.numeric_value = float(value)
-        else:
+        elif measurement.measured_value.is_value_set:
```
grybmadsci: Hm, I think I prefer to tweak this a little farther up, around line 231:
```py
if not measurement.measured_value:
  testrun_param.status = test_runs_pb2.ERROR
  continue
if not measurement.measured_value.is_value_set:
  testrun_param.status = test_runs_pb2.FAIL
  # Don't continue here, we'll generate an 'Unset' detail code below.
else:
  value = measurement.measured_value.value
  if measurement.dimensions is None:
    # ...
    # Indent all the way to line 270.
if testrun_param.status == test_runs_pb2.FAIL:
```
And if you're bored while you're here, you might pull some of these guts out into a helper function, this function is a bit too long to be easily digestible (my fault, sorry :( ).

PR #406, opened 2016-08-18
Dynamically Create Plugs and Plug Placeholders

kdsudac:
This is a proof of concept of a feature I think can improve flexibility of openhtf:

* Ability to pass plugs into phases via `WithPlugs` method.  This makes it easy to use same test logic for the multiple plugs of same parent class.  Typical use case is you have a component Y (say a motor) that you test at individual level before being integrated into an assembly that has N x Y (say 10 X motors).  After assembly you want to run the same functional test on all of the individual components.

I included a `examples/with_plugs.py` to show basic usage.

Can someone please review the general concept and provide feedback?

I still need to clean up documentation, write unit tests, clean up example, check for impact on other openhtf features (monitors, frontend, etc), and test.  But want to make sure concept gets approval before doing the extra work.


kdsudac:
fahhem, convinced me that copy/paste of subclasses is appropriate for most uses cases so extra complexity required for `create_subclass` method and __metaclass__ may not be worth it.

grybmadsci:
Yeah, it's a neat idea, I like the concept but I think it might be a bit overkill for now.  We can definitely look into adding something like this in once we're on PyPI and everything is a bit more stable.

kdsudac:
I still think (and fahhem agreed) that `withPlugs` is a useful addition.  Mind giving that a look over?

I removed the `create_subclass` method and updated pull request description.

openhtf/plugs/__init__.py
```diff
@@ -142,6 +143,15 @@ class BasePlug(object):
   # Allow explicitly disabling remote access to specific attributes.
   disable_remote_attrs = set()
 
+  class Placeholder(object):
```
fahhem: placeholder is a confusing name, but I'm at a loss for better ones :/

However, `self.cls` is really not helping in the naming department. Consider `self.base_class` or similar

Lastly, this can use the new [util.classproperty][0] to become a better API and potentially remove this class altogether

[0]: https://github.com/google/openhtf/blob/master/openhtf/util/__init__.py#L95

openhtf/__init__.py
```diff
@@ -394,6 +394,28 @@ def with_args(self, **kwargs):
...
     return new_info
 
+  def WithPlugs(self, **subplugs):
+    """Substitute plugs for placeholders for this phase."""
+    # Make a copy so we can have multiple of the same phase with different plugs
+    # in the same test.
+    new_info = mutablerecords.CopyRecord(self)
+    
+    for plug_name, plug_class in subplugs.items():
+      placeholders = [(i, p) for i, p in enumerate(self.plugs)
```
fahhem: there can only be one plug with this name, right? This may benefit from taking advantage of that:

```py
for subplug_name, subplug_class in subplugs.items():
  for i, placeholder_plug in enumerate(self.plugs):
    if placeholder_plug.name == subplug_name:
      if not isinstance(placeholder_plug, plugs.BasePlug.Placeholder)):
        raise plugs.InvalidPlugError('Plug named %s was not a placeholder for %s' % (
            subplug_name, subplug_class))
      new_info.plugs[i] = mutablerecords.CopyRecord(placeholder_plug, cls=subplug_class)
      break
  else:
    raise plugs.InvalidPlugError('No plug named %s required for phase %s' % (
        subplug_name, self.name))
```

That might be worse. It definitely needs comments

kdsudac:
@fahhem thanks for the comments.  Implemented suggestions.

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+"""Example OpenHTF test logic.
+
+Run with (your virtualenv must be activated first):
+
+  python with_plugs.py
+
+Afterwards, take a look at the hello_world.json output file.  This will
+give you a basic idea of what a minimal test outputs.
+
+For more information on measurements, see the measurements.py example.
```
grybmadsci: Can you update this to describe the basic intent/mechanism behind `with_plugs()`

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+TODO(someone): Write an output example.
+For more information on output, see the output.py example.
+"""
+import subprocess
+import time
+
+import openhtf as htf
+from openhtf import plugs
+
+class PingPlug(plugs.BasePlug):
```
grybmadsci: examples need to be very heavily docstring'd, every single class, method, and function should have docstring.  i know it's annoying, but it's really intended for an audience who may have little to no python experience.

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+      str(count),
+      self.host,
+    ]
+
+  def run(self, count):
+    command = self.get_command(count)
+    print "running: %s" % command
+    return subprocess.call(command)
+
+class PingGoogle(PingPlug):
```
grybmadsci: docstrings can be omitted on these subclasses, but can you add a comment at the top of this list to indicate what these are and why you need them?

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+
+TODO(someone): Write an output example.
+For more information on output, see the output.py example.
+"""
+import subprocess
+import time
+
+import openhtf as htf
+from openhtf import plugs
+
```
grybmadsci: two blank lines before class def

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+class PingExample(PingPlug):
+  host = 'example.com'
+
+class PingCnn(PingPlug):
+  host = 'cnn.com'
+
+@plugs.plug(pinger=PingPlug.placeholder)
+@htf.measures('total_time', 'retcode')
+def MainTestPhase(test, pinger, count):
```
grybmadsci: we should probably follow function naming convention with phase names too

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+  retcode = pinger.run(count)
+  elapsed = time.time() - start
+  test.measurements.total_time = elapsed
+  test.measurements.retcode = retcode
+
+
+
+if __name__ == '__main__':
+  # We instantiate our OpenHTF test with the phases we want to run as args.
+  ping_plugs = [
```
grybmadsci: I would use tuples here instead of lists, these are immutable.

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+    PingExample,
+    PingCnn
+  ]
+  counts = [3, 5, 10]
+  phases = []
+  for ping_plug in ping_plugs:
+    for count in counts:
+      phases.append(MainTestPhase
+          .with_args(count=count)
+          .with_plugs(pinger=ping_plug))
```
grybmadsci: in the interest of being a clear example, this is fine, but maybe put a comment below that if you're comfortable with list comprehensions, the preferred syntax for a construct like this would be:
```
test = htf.Test(*(MainTestPhase.with_args(count=count).with_plug(pinger=ping_plug) for
                          ping_plug in ping_plugs for count in counts))
```

examples/with_plugs.py
```diff
@@ -0,0 +1,101 @@
...
+if __name__ == '__main__':
+  # We instantiate our OpenHTF test with the phases we want to run as args.
+  ping_plugs = [
+    PingGoogle,
+    PingDnsA,
+    PingDnsB,
+    PingExample,
+    PingCnn
+  ]
+  counts = [3, 5, 10]
```
grybmadsci: comments to describe what's going on here.  what are these variables for, why are you looping over them?

openhtf/__init__.py
```diff
@@ -394,6 +394,33 @@ def with_args(self, **kwargs):
...
 
+  def with_plugs(self, **subplugs):
+    """Substitute plugs for placeholders for this phase."""
+    # Make a copy so we can have multiple of the same phase with different plugs
+    # in the same test.
+    new_info = mutablerecords.CopyRecord(self)
+    
+    for substitute_name, substitute_class in subplugs.iteritems():
+      # self.plugs contains PhasePlug so enumerate and unpack
+      for i, original_plug in enumerate(self.plugs):
```
grybmadsci: O(n^2) :'(  How about iterating over self.plugs and removing keys from subplugs as you find them, then at the end any keys left in subplugs are ones for which you couldn't find a valid placeholder.

openhtf/__init__.py
```diff
@@ -394,6 +394,33 @@ def with_args(self, **kwargs):
     new_info.measurements = [m.with_args(**kwargs) for m in self.measurements]
     return new_info
 
+  def with_plugs(self, **subplugs):
+    """Substitute plugs for placeholders for this phase."""
+    # Make a copy so we can have multiple of the same phase with different plugs
+    # in the same test.
+    new_info = mutablerecords.CopyRecord(self)
```
grybmadsci: hm, i think i have a slight preference for `new_plugs = list(self.plugs)` here, update `new_plugs` directly on line 411, and then `return mutablerecords.CopyRecord(self, plugs=new_plugs)` at the end.  It's a little clearer that plugs are the only thing you're updating.

openhtf/output/web_gui/__init__.py
```diff
@@ -38,7 +38,7 @@
 To start the frontend server, invoke python with the -m flag in a python
 environment where openhtf is installed:
 
-$ python -m openhtf.io.frontend
+$ python -m openhtf.output.web_gui
```
grybmadsci: lol, thanks :) :+1:

openhtf/plugs/__init__.py
```diff
@@ -142,6 +155,11 @@ class BasePlug(object):
   # Allow explicitly disabling remote access to specific attributes.
   disable_remote_attrs = set()
 
+  @classproperty
+  def placeholder(cls):
+    """Returns a PlugPlaceholder for calling class"""
```
grybmadsci: period at end of docstring.

openhtf/plugs/__init__.py
```diff
@@ -259,7 +277,8 @@ def plug(update_kwargs=True, **plugs):
     InvalidPlugError: If a type is provided that is not a subclass of BasePlug.
   """
   for a_plug in plugs.itervalues():
-    if not issubclass(a_plug, BasePlug):
+    is_placeholder = isinstance(a_plug, PlugPlaceholder)
+    if not (is_placeholder or issubclass(a_plug, BasePlug)):
```
grybmadsci: no need for the `is_placeholder` intermediate variable, just collapse these into one line

grybmadsci:
awesome, i really like the simplified version, and i like how closely the `.with_plugs()` user-facing mechanism mirrors the `.with_args()` mechanism.  couple minor changes but no major complaints, thanks! :)

openhtf/__init__.py
```diff
@@ -394,6 +394,33 @@ def with_args(self, **kwargs):
...
+    
+    for substitute_name, substitute_class in subplugs.iteritems():
+      # self.plugs contains PhasePlug so enumerate and unpack
+      for i, original_plug in enumerate(self.plugs):
+        original_name = original_plug.name
+        original_class = original_plug.cls
```
jettisonjoe: I feel these newly assigned names aren't particularly shorter or more clear than just using the originals directly.

jettisonjoe:
Yep, I'm down with this simplified version. Nice work.

openhtf/__init__.py
fahhem: It's O(n*m) where n and m are never expected to be more than, say, 50. I'd avoid any performance-based changes for what will never be more than 2500 iterations, if more than 10.

However, if you really cared, The main reason for the performance hit is the O(n) lookup of a plug with the right name. This will speed up the lookups without making the code have randoms `.pop()` calls:

```py
plugs_by_name = {plug.name: plug for plug in self.plugs}
new_plugs = dict(plugs_by_name)

for name, sub_class in subplugs.iteritems():
  original_plug = plugs_by_name.get(name)
  if (original_plug is None
      or not isinstance(original_plug.cls, plugs.PlugPlaceholder)
      or not isinstance(sub_class, original_plug.cls.base_class)):
    raise plugs.InvalidPlugError('...')
  new_plugs[name] = mutablerecords.CopyRecord(original_plug, cls=sub_class)
return mutablerecords.CopyRecord(self, plugs=new_plugs.values())
```

It seems to have also made it less complex, so it has that going for it rather than O(N*M)

grybmadsci: i'd be okay with that solution too... the point wasn't the performance hit, it was the structure of doing two nested loops to perform what is essentially a set intersection.  any restructure that removes the need for the second loop is fine :)

PR #412, opened 2016-08-23
Add extra_kwargs to name of phases

fahhem:
This makes the name of phases include their args, so if you repeat a phase multiple times with different arguments, the output (and frontend) is much more useful.

This will likely conflict with Joe's reorg-the-world PR, so feel free to get that in then I'll fix this afterward. I just want to get feedback on the idea ahead of time

### DIFF
--- a/openhtf/util/__init__.py
+++ b/openhtf/util/__init__.py
@@ -104,3 +104,27 @@ class classproperty(object):
 
   def __get__(self, instance, owner):
     return self._func(owner)
+
+
+def format_string(target, **kwargs):
+  """Formats a string in any of three ways (or not at all).
+
+  Args:
+    target: The target string to format. This can be a function that takes a
+        dict as its only argument, a string with {}- or %-based formatting, or
+        a basic string with none of those. In the latter case, the string is
+        returned as-is, but in all other cases the string is formatted (or the
+        callback called) with the given kwargs.
+        If this is None (or otherwise falsey), it is returned immediately.
+    **kwargs: The arguments to use for formatting.
+        Passed to string.format, %, or target if it's callable.
+  """
+  if not target:
+    return target
+  if callable(target):
+    return target(kwargs)
+  if '{' in target:
+    return target.format(**kwargs)
+  if '%' in target:
+    return target % kwargs
+  return target
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -300,10 +300,12 @@ PhaseResult = Enum('PhaseResult', ['CONTINUE', 'REPEAT', 'STOP'])
 
 
 class PhaseOptions(mutablerecords.Record('PhaseOptions', [], {
-    'timeout_s': None, 'run_if': None, 'requires_state': None})):
+    'name': None, 'timeout_s': None, 'run_if': None, 'requires_state': None})):
   """Options used to override default test phase behaviors.
 
   Attributes:
+    name: Override for the name of the phase. Can be formatted similar to
+        measurements using with_args() to distinguish multiple similar phases.
     timeout_s: Timeout to use for the phase, in seconds.
     run_if: Callback that decides whether to run the phase or not.
     requires_state: If True, pass the whole TestState into the first argument,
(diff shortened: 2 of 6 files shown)


grybmadsci:
hm, what do you think about promoting these kwargs to being included in the PhaseRecord?

fahhem:
In addition or instead of this? I'd like it to be in the name since that's used in a ton of places that don't benefit from understanding extra_kwargs separately.

grybmadsci:
i was thinking instead of, but what cases do you have in mind that use the name?  it's a little concerning that we'd use the name for anything other than user-facing strings.  once you start name mangling you start making it harder to get back to the original source.  i'm all for adding the args info, because it's definitely valuable, but i'd rather not lose the original name to do so (and making it so you have to string-parse it out is dangerously close to losing it).  any other ideas that allow preserving the original name while still adding the kwargs?

fahhem:
Well, the places I care about are user-facing output. When things get logged, in output callbacks, and the frontend, I have "TestPort" repeating about 10 times when I want "TestPort(port=<something useful>)" with 10 different useful strings.

An alternative would be to have a name that gets formatted by `WithArgs`:

```py
@plugs.plug(ssh=...)
@openhtf.PhaseOptions(name='TestPort({port})')
def TestPort(test, ssh, port):
  pass

openhtf.Test(
    TestPort.WithArgs(port='front_right'),
)
```

The output would then be "TestPort(front_right)"

grybmadsci:
that's a neat idea, i like it - applying the formatting to the name.  it makes name consistent with other ways we use the args (ie measurements), and makes it so the test author is a bit more explicitly in control of it.  SGTM

fahhem:
Okay, the PR grew quite a bit, but I think it's better now. Let me know what you think, especially about the limitation mentioned in the tests. I'll port this to post-Joe's PR once that's in.

openhtf/util/__init__.py
```diff
@@ -104,3 +104,27 @@ def __init__(self, func):
...
+    **kwargs: The arguments to use for formatting.
+        Passed to string.format, %, or target if it's callable.
+  """
+  if not target:
+    return target
+  if callable(target):
+    return target(kwargs)
+  if '{' in target:
+    return target.format(**kwargs)
+  if '%' in target:
```
grybmadsci: hm, at one point i think we deprecated this usage?  i don't feel super strongly about it, but i believe it's a strict subset of the .format() mechanism.

openhtf/util/measurements.py
```diff
@@ -203,16 +204,9 @@ def with_validator(self, validator):
...
-      formatter = lambda x: x.format(**kwargs) if x else x
-    else:
-      # str % {'a': 1} is harmless if str doesn't use any interpolation.
-      # .format is as well, but % is more likely to be used in other contexts.
-      formatter = lambda x: x % kwargs if x else x
-    new_meas.name = formatter(self.name)
-    new_meas.docstring = formatter(self.docstring)
-    return new_meas
+    return mutablerecords.CopyRecord(
+        self, name=util.format_string(self.name),
```
grybmadsci: i think you need to pass kwargs through to format_string?  in the other places you call format_string, too.

grybmadsci:
LGTM overall, pending rebase after Joe's change goes in.

grybmadsci:
nice job on the format_string() refactor btw, between that and the mutablerecords.CopyRecord() update to allow inline overriding, this cleaned up a lot of code :)

openhtf/util/__init__.py
fahhem: It's deprecated in only one place, so I put extra logic there instead of here to maintain compatibility. It's a subset, but I don't see the harm in it now that it's all in the same place. Previously we were maintaining multiple copies of this function and each option we provided was more maintenance.

openhtf/util/measurements.py
fahhem: Whoops!

fahhem:
Okay, ready to get redone after joe's PR :)

openhtf/util/__init__.py
grybmadsci: yeah, agreed, i don't have any strong objection to it, it's much much cleaner now.  @jettisonjoe tends to feel more strongly about the "There should be one-- and preferably only one --obvious way to do it." zen of python philosophy, so he may care more :)

jettisonjoe: I suspect I _do_ care more, but not enough to block this.

jettisonjoe:
LGTM w/ rebase.

PR #413, opened 2016-08-23
Fix *all* outstanding bugs for PyPI.

### LINKED ISSUE #198: Save full exception info in phase result, not just exception instance.
Right now, we only save the exception in the phase result, thereby losing any stacktrace information.  I'd like to include the full `sys.exc_info()` tuple.

This would likely mean a change in the output from just a str()'d Exception to a stacktrace, but I think that's preferable, thoughts?

### LINKED ISSUE #292: tear down function doesn't run if SIGINT recieved
running all_the_things without looping.  Plug tear down runs but tear down function doesn't run.

python all_the_things.py 
Provide a DUT ID in order to start the test.
test
Hello World!
What's the widget type?
^C2016-06-01 21:03:15,698 - ERROR - Received SIGINT.
2016-06-01 21:03:15,698 - ERROR - Stopping Test due to SIGINT
2016-06-01 21:03:15,698 - ERROR - Test state: <TestState: test@wallaceb-glaptop Running Phase: hello_world>
Tearing down <ExamplePlug: 0>

jettisonjoe:
Fixes #292 
Fixed #198


grybmadsci:
LGTM once tests pass.

jettisonjoe:
Passes on my Travis, so merging.

PR #417, opened 2016-08-24
Fix the web gui so it runs from PyPI installs.

### LINKED ISSUE #416: Web gui broken in pypi package.
The web gui doesn't work from PyPI install because I forgot to update the `package_data` path in `setup.py` when restructuring the package.

jettisonjoe:
Fixes #416.

There are three commits in this PR. In order, they:
  * Update `setup.py` with better `package_data` definitions that should no longer exclude important web gui parts.
  * Update the web client's `package.json` file to both elimintate npm install warnings, and to lock Angulars and RxJs versions. This is necessary because Angular2 is in active development, and tracking the latest changes without manually updating peer dependencies keeps resulting in conflicts that make `npm install` impossible.
  * Update the prebuilt web client.


fahhem:
LGTM

grybmadsci:
LGTM

PR #418, opened 2016-08-25
Fix a logic bug in the response path of non-text prompts

fahhem:


jettisonjoe:
Nice catch, thanks! LGTM.

fahhem:
Hmm, this prebuilt thing is gonna be annoying, huh? I almost merged this without updating it, and it looks like there's a bug in the prebuilt that's fixed in src/ (and was seemingly never broken in src/). Unfortunately, having Travis build the prebuilt/ folder won't help as much since we want devs to be able to run the frontend out of their repos without building it.

How hermetic/repeatable is the frontend build? Can we add a check in travis that builds the frontend and checks that the version checked-in is the same as what it just built?

jettisonjoe:
Hmm, I'm not sure. The problem is that building the frontend requires an `npm install` of our package, which has a bunch of external deps for npm to resolve. While we can version-lock our direct deps, we can't as easily control _their_ dep resolution. So we'll likely end up with bundles that roll in different versions of second-order deps, and thus fail any easy-to-implement identity check.

fahhem:
Oh shoot I wanted to update the prebuilts and then I forgot! I tried updating the prebuilts but I only got the vendor.js.map to change so that'll take a while too

Issue #420, opened 2016-08-27
all the things raises AttributeError: 'module' object has no attribute 'UserInput'


wallacbe:
Ran TOT, observe:

  File "all_the_things.py", line 47, in <module>
    @htf.plug(prompts=user_input.UserInput)
AttributeError: 'module' object has no attribute 'UserInput'

PR #422, opened 2016-08-27
fix prompts in all_the_things

wallacbe:
https://github.com/google/openhtf/issues/420


grybmadsci:
lgtm

jettisonjoe:
LGTM, merging. I swear I've fixed this and the other `all_the_things.py` bug at least twice, but I guess I probably reverted them each time after subsequently using `all_the_things.py` to generate output for the unit tests.

Issue #421, opened 2016-08-27
all the things raises AttributeError: 'module' object has no attribute 'output'


wallacbe:
python all_the_things.py 
Traceback (most recent call last):
  File "all_the_things.py", line 80, in <module>
    test.add_output_callbacks(htf.output.callbacks.OutputToFile(
AttributeError: 'module' object has no attribute 'output'

wallacbe:
core, plugs and util from the openhtf directory show up in the directory listing but output does not

>>> import openhtf
>>> print dir(openhtf)
['Enum', 'InvalidTestPhaseError', 'InvalidTestStateError', 'Measurement', 'PhaseDescriptor', 'PhaseOptions', 'PhasePlug', 'PhaseResult', 'Test', 'TestApi', 'TestDescriptor', 'TestOptions', 'TestPhase', 'UnrecognizedTestUidError', '_LOG', '__builtins__', '__doc__', '__file__', '__name__', '__package__', '__path__', '__version__', 'argparse', 'collections', 'conf', 'copy', 'core', 'create_arg_parser', 'data', 'functions', 'functools', 'history', 'inspect', 'itertools', 'json', 'logging', 'logs', 'measures', 'monitors', 'mutablerecords', 'phase_executor', 'plug', 'plugs', 'signal', 'socket', 'station_api', 'sys', 'test_record', 'threading', 'triggers', 'units', 'user_input', 'util', 'weakref']

PR #423, opened 2016-08-28
add output module imports

wallacbe:
https://github.com/google/openhtf/issues/421


wallacbe:
PTAL

fahhem:
Why was this closed? Was this not the desired solution?

wallacbe:
Discussion at beginning of meeting today was to import modules using full path so we don't pollute the __init__

PR #428, opened 2016-08-30
Handle unset measurements correctly in mfg_inspector

fahhem:


fahhem:
Updated test output and it is exactly what we want: the status becomes ERROR and the failure codes go away.

jettisonjoe:
LGTM

wallacbe:
LGTM

PR #430, opened 2016-09-01
Dynamically Create Plugs and Plug Placeholders

fahhem:
Pushing #406 forward while kdsudac is on vacation

### DIFF
--- a/examples/with_plugs.py
+++ b/examples/with_plugs.py
@@ -27,9 +27,10 @@ TODO(someone): Write an output example.
 For more information on output, see the output.py example.
 
 with_plugs() is most useful when you have a test that has to happen in the same
-(or mostly similar) ways across multiple interfaces, such testing connectivity
-on a quad-NIC board. Instead of creating 4 phases with 4 plugs, you create 1 phase
-with 4 subclasses of the same plug and use with_plugs() to create the 4 phases.
+(or mostly similar) ways across multiple interfaces, such as testing
+connectivity on a quad-NIC board. Instead of creating 4 phases with 4 plugs, you
+can create 1 phase with 4 subclasses of the same plug and use with_plugs() to
+end up with the 4 phases you want.
 """
 import itertools
 import subprocess
--- a/examples/with_plugs.py
+++ b/examples/with_plugs.py
@@ -16,7 +16,7 @@
 
 Run with (your virtualenv must be activated first):
 
-  python dynamic_plugs.py
+  python with_plugs.py
 
 Afterwards, take a look at the hello_world.json output file.  This will
 give you a basic idea of what a minimal test outputs.
@@ -29,20 +29,14 @@ For more information on output, see the output.py example.
 import subprocess
 import time
 
-# Import this output mechanism as it's the specific one we want to use.
-from openhtf.io.output import json_factory
+import openhtf as htf
 from openhtf import plugs
-
-# Import a handful of useful names.  If you're worried about polluting
-# your namespace, you can manually import just the things you want, this
-# is just a convenience.  See names.py for an exhaustive list.
-from openhtf.names import *
-
 
 class PingPlug(plugs.BasePlug):
 
-  def __init__(self, host):
-    self.host = host
+  host = None
+  def __init__(self):
+    assert self.host is not None
 
   def get_command(self, count):
     return [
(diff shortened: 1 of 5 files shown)


fahhem:
PTAL

kdsudac:
LGTM

openhtf/__init__.py
```diff
@@ -405,6 +405,22 @@ def with_args(self, **kwargs):
...
+      original_plug = plugs_by_name.get(name)
+      if (original_plug is None
+          or not isinstance(original_plug.cls, plugs.PlugPlaceholder)
+          or not issubclass(sub_class, original_plug.cls.base_class)):
+        raise plugs.InvalidPlugError(
+            'Could not find valid placeholder for substitute plug %s '
+            'required for phase %s' % (name, self.name))
+      new_plugs[name] = mutablerecords.CopyRecord(original_plug, cls=sub_class)
+    return mutablerecords.CopyRecord(self, plugs=new_plugs.values())
+
```
kdsudac: :thumbsup:

fahhem:
@jettisonjoe or @grybmadsci, could either of you take a look today or tomorrow? It's pretty simple and generally accepted, but I'd like at least a cursory review before submitting it :)

jettisonjoe:
I'll look to tomorrow. I was planning to have one more day of vacation but
in fact I'm coming in to work after all.

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+Afterwards, take a look at the hello_world.json output file.  This will
+give you a basic idea of what a minimal test outputs.
+
+For more information on measurements, see the measurements.py example.
+
+TODO(someone): Write an output example.
+For more information on output, see the output.py example.
+
+with_plugs() is most useful when you have a test that has to happen in the same
+(or mostly similar) ways across multiple interfaces, such testing connectivity
```
grybmadsci: such testing, much wow.

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+on a quad-NIC board. Instead of creating 4 phases with 4 plugs, you create 1 phase
```
grybmadsci: you can/may create 1 phase

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+import subprocess
+import time
+
+import openhtf as htf
+from openhtf import plugs
+
+
+class PingPlug(plugs.BasePlug):
+  """This plug simply does a ping against the host attribute."""
+  enable_remote = True
```
grybmadsci: unused?

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+    return [
+      'ping',
+      '-c',
+      str(count),
+      self.host,
+    ]
+
+  def run(self, count):
+    command = self._get_command(count)
+    print "running: %s" % command
```
grybmadsci: probably `''.join(command)` instead of just command? it's a little odd to see a list in the context of "running"

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+  # using only 1 written phase.
+  counts = [3, 5, 10]
+  ping_plugs = [
+    PingGoogle,
+    PingDnsA,
+    PingDnsB,
+  ]
+
+  test = htf.Test(
+      *(test_ping.with_args(count=count).with_plugs(pinger=plug)
```
grybmadsci: would rather not include `.with_args()` in this example.  the examples really should introduce exactly one concept and only use mechanisms absolutely required to demonstrate that concept (ie, you need to use plugs to demonstrate `.with_plugs()` sensibly).  You could remote the whole `count` mechanism and make this all a bit simpler.

fahhem:
Thanks for the review @grybmadsci!

examples/with_plugs.py
```diff
@@ -0,0 +1,114 @@
...
+        for count, plug in itertools.product(counts, ping_plugs)))
```
grybmadsci: no need to use things like `itertools.product` in an example.  we're not trying to teach people fancy python, the target audience should need to know little more than basic python syntax.  the use of `*generator` is pushing the level of complexity for an example, but since that's almost certainly how one would use `.with_plugs()`, it's reasonable to include.

fahhem:
Pretend I replied "Done" to all comments :)

openhtf/plugs/__init__.py
```diff
@@ -258,10 +275,11 @@ def plug(update_kwargs=True, **plugs):
   Raises:
     InvalidPlugError: If a type is provided that is not a subclass of BasePlug.
   """
-  for a_plug in plugs.itervalues():
-    if not issubclass(a_plug, BasePlug):
+  for plug in plugs.itervalues():
```
grybmadsci: doesn't `plug` mask the outer function `plug` here (ie this function)?

grybmadsci:
LGTM pending nits :)

openhtf/plugs/__init__.py
fahhem: ahh, I was wondering why it was `a_plug` before... I looked everywhere but the function I was in :)

fahhem:
Okay, I'll wait for the tests to pass and then submit. Thanks!

jettisonjoe:
LGTM. Nice example test, too!

PR #434, opened 2016-09-08
adb_protocol: Send signed data token for AUTH response

arsharma1:
The AUTH protocol was trying to encrypt the local banner rather than the random token provided by the device.  Fixed this to use the token as expected.

Also, fixed the final WriteMessage AUTH call that sends the public key to the device.


arsharma1:
@fahhem

fahhem:
For the record, google/python-adb does this correctly and is the supported version of this library. Other than that, LGTM. I'll merge this if the tests pass

PR #437, opened 2016-09-15
Make prompts more of a normal plug.

### LINKED ISSUE #431: while Loop that catches BlankDutIdError raises MultiplePromptsError when SIGINT
Sometimes operators hit enter by mistake and don't input a serial.  I added a try block in my main loop to catch this and not kill the tester.

 # Loop
  while True:
    try:
      test.execute(test_start=user_input.prompt_for_test_start(output.PROMPT))
    except htf.core.test_state.BlankDutIdError:
      pass

However If I then CTRL+C to exit I get MultiplePromptsError, and it continues to loop and doesn't exit.
 DisplayPrompt
    raise MultiplePromptsError
MultiplePromptsError

I have an info phase that checks for DUT id validity.  Could we accept a blank ID in the tester and not raise  htf.core.test_state.BlankDutIdError?

jettisonjoe:
Fixes #431

  * Remove old non-plug `user_input` module.
  * Remove singleton `PROMPT_MANAGER` since plugs are per-execute singletons.
  * Update version number in preparation for a tagged release.
  * Remove triggers module, as triggers should usually be in plugs.


jettisonjoe:
Running tox on my local machine produces this:
```
======================== 54 passed, 2 xfailed, 2 pytest-warnings in 2.58 seconds =========================________________________________________________ summary _________________________________________________
  py27: commands succeeded
  congratulations :)
```

Travis-ci, sometimes I hate you so much. :angry:

fahhem:
I'm a little lost. How does this fix #431? Also, does the frontend path work still? I see respondToPrompt is commented out.

jettisonjoe:
This fixes #431 by doing away with the module-level singleton prompt manager. While that existed, calling `text.execute()` in a loop could result in a start trigger trying to create a new prompt while a leftover from the previous (failed) `execute()` call still existed in the singleton prompt manager. By fully changing this to a plug and doing away with the prompt manager, you get a new instance of the `UserInput` plug each `execute()` call, thus no leftover prompts from previous test runs.

Can you make a line-level comment for respondToPrompt? I don't see it.

jettisonjoe:
@grybmadsci @fahhem could I get someone to look at this today or tomorrow? #431 is currently blocking @wallacbe a bit and this should fix it, as well as lay the groundwork for re-wiring prompts through to the frontend.

setup.py
```diff
@@ -158,7 +158,7 @@ def run_tests(self):
 
 setup(
     name='openhtf',
-    version='1.0.2',
+    version='1.0.3',
```
fahhem: I wouldn't update the openhtf versions inside other commits

fahhem:
This new review thing's got me turned around. I didn't realize saying that I requested changes would actually block merging!

openhtf/plugs/user_input.py
```diff
@@ -173,26 +111,67 @@ def run(self):
...
+    self._cond = threading.Condition()
+    
   def prompt(self, message, text_input=False, timeout_s=None):
-    return PROMPT_MANAGER.DisplayPrompt(
-        message, text_input=text_input, timeout_s=timeout_s)
+    """Prompt for a user response by showing the message.
+
+    Args:
+      message: The message to display to the user.
+      text_input: True iff the user needs to provide a string back.
```
grybmadsci: missing timeout_s in docstring

openhtf/plugs/user_input.py
```diff
@@ -173,26 +111,67 @@ def run(self):
...
 def prompt_for_test_start(
     message='Provide a DUT ID in order to start the test.', text_input=True,
     timeout_s=60*60*24):
   """Make a test start trigger based on prompting the user for input."""
   def trigger():  # pylint: disable=missing-docstring
-    return PROMPT_MANAGER.DisplayPrompt(
+    plug = UserInput()
+    response = plug.prompt(
         message, text_input=text_input, timeout_s=timeout_s)
+    plug.tearDown()
```
grybmadsci: we might consider adding some utility to plugs' usage outside of the plug manager, auto-teardown on exit from a context, for example

setup.py
grybmadsci: +1, version bump should happen in a commit where it is the only thing.  not sure if we want to trigger a new version on this change.  is there anything else in the pipe?  might be better to get in the practice of a weekly version-bump-if-needed policy rather than when there's a commit.

grybmadsci:
LGTM in principle, i'm half blind recovering from lasik last week so didn't look too closely at the implementation details.

setup.py
jettisonjoe: Fixing.

openhtf/plugs/user_input.py
jettisonjoe: Fixing.

openhtf/plugs/user_input.py
jettisonjoe: Yeah, that could be good. Probably a separate issue though.

jettisonjoe:

fahhem:
LGTM

PR #441, opened 2016-09-23
Skip handling new plugs when the state goes away

fahhem:
This is triggered when the station is killed while the frontend is waiting for an update.

### DIFF
--- a/openhtf/output/web_gui/__init__.py
+++ b/openhtf/output/web_gui/__init__.py
@@ -368,6 +368,9 @@ class WebGuiServer(tornado.web.Application):
     This handler updates our URL handlers with any newly available plug
     information and notifies the StationPubSub of the event.
     """
+    if state is None:
+      return
+
     def _make_sockjs_url(rel_url):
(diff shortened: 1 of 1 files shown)


openhtf/output/web_gui/__init__.py
```diff
@@ -368,6 +368,8 @@ def handle_test_state_update(self, hostport, test_uid, state):
     This handler updates our URL handlers with any newly available plug
     information and notifies the StationPubSub of the event.
     """
+    if state is None:
+      return
```
jettisonjoe: I feel like this would be better with a blank line between `return` and  `_make_sockjs_url`, but YMMV (the [style guide](https://google.github.io/styleguide/pyguide.html#Blank_Lines) leaves it up to personal judgement).

grybmadsci:
LGTM

openhtf/output/web_gui/__init__.py
fahhem: Done

PR #442, opened 2016-09-23
AsyncCommandHandle: Keep waiting until timeout

arsharma1:
The AsyncCommandHandle's reader thread does not use the timeout passed in the constructor.  This causes it to use the default timeout buried in usb_handle of 5 seconds, which may not be the correct amount of time to wait, terminating the reader thread.


arsharma1:
@fahhem

PR #443, opened 2016-09-23
Trigger refactoring as a prerequisite for re-wiring prompts through to frontend.

jettisonjoe:
This PR represents my current thinking towards making prompts work again in the frontend. Specifically, it moves toward leveraging the station API's plug handlers that @grybmadsci already put the guts in place for, while still allowing start triggers to use plugs. It is currently failing tests because the `openhtf.util.test` module uses the old construct:

```python
test.execute(test_start=lambda: 'TestDutId')
```

I hadn't worked out a way to support those old, simple start triggers yet, while also having all the wiring in place to use plug-based triggers, but am open to ideas and alternatives.

I'm creating this PR today so @fahhem can pick up the torch of making prompts work while I'm on vacation next week.


openhtf/plugs/__init__.py
```diff
@@ -407,6 +421,12 @@ def initialize_plugs(self, plug_types=None):
       self.update_plug(plug_type, plug_instance)
     self._initialize_rpc_server()
 
+  def call_trigger(self, trigger_info):
+    """Call the trigger corresponding to trigger_info on the matching plug."""
+    trigger = getattr(self._plugs_by_type[trigger_info.plug_type],
```
grybmadsci: you might consider catching and reraising here with a more specific error message than AttributeError will have.

fahhem:
Hmm, I'm not sure why we want to do all this instead of allowing test_start to be a phase (in that it has plugs and is invoked using a plug_manager). TestState is even available already.

jettisonjoe:
Start triggers generally don't need access to the whole test state, only a single plug. Also, they shouldn't be counted against the test's cycle time as normal phases are. Definitely willing to discuss further. Were you able to get any farther while I was out last week? I plan to work on OpenHTF stuff tomorrow and Friday and can take back the torch on prompt plumbing if you want. Let me know how you want to play it!

fahhem:
There is no test state anyway, and we don't have to start the test, nor even record the phase's info, to treat it like a normal phase in terms of plugs, logging, etc.

I've assigned this to Kenneth (you'll see him around here once he learns more about Typescript/Angular2) and he'll be available to work on plugins in the UI, since that's a high priority for us. He may not get around to prompts before you though.

jettisonjoe:
Chatted with @grybmadsci a little more about this yesterday and he reminded me of another possibly important issue. The PR as written swaps plug instantiation to before the start trigger gets called. Although it looks superficially like a minor change, this is bad because it means test equipment could potentially get powered on and initialized, then sit there for hours and hours between tests. Ideally we should do test equipment initialization (plug instantiation) _right_ before the test actually runs, not before we start waiting for an arbitrarily long period of time for a start trigger.

Looking into the start-trigger-as-a-phase concept, it suffers from the same issue. With the way the code is currently structured, there's no easy way to run the start trigger as you would a normal phase without initializing all plugs first.

One possibly way to solve this would be lazy plug instantiation. Meaning plugs aren't instantiated until the phase that needs them is being run. That's the avenue I'm investigating now.

fahhem:
Unfortunately, we make some assumptions about the plugs being initialized at the beginning to catch missing/unplugged tester equipment. Otherwise we may need to run halfway through a test to determine that something's not plugged in.

Both the start-trigger-as-phase and start-as-plug-trigger can only initialize the plugs relevant to the start trigger. In your implementation, you have the plug manager initialize the test's plugs as well as the test_start's plugs, but we could easily add a function for only initializing (and potentially caching) the plugs necessary for the start trigger.

jettisonjoe:
Shucks. Alright, then. If you guys need all plugs initialized at test start then I'll drop the lazy plug loading idea and work on work on trigger-as-a-phase with selective plug initialization (and caching).

fahhem:
Well, if there was an option for the plug to decide if it was an eager or
lazy plug, that would be fine too. We can mark the ones that are eager as
such. There are some other plugs with a Start() method for lazy
initialization.

I just meant that it couldn't be forced-laziness

jettisonjoe:
Okay, let's try this again. Ready for you guys to take a look at the start-trigger-as-a-phase approach and see what you think so far. I snuck some other small changes in here, too:
  * Make `test.execute()` return a boolean representing PASS/FAIL
  * Start fixing a bug in `RemotePlug`'s `discover()`.

But the logic remix in `TestExecutor` is what I'd really like you guys to scrutinize the most.

openhtf/__init__.py
```diff
@@ -239,7 +239,10 @@ def execute(self, test_start=None):
             _LOG.exception(
                 'Output callback %s raised; continuing anyway', output_cb)
       finally:
+        result = (self._executor.test_state.test_record.outcome
+                  == test_record.Outcome.PASS)
         self._executor = None
+        return result
```
grybmadsci: returning from inside a finally block is dangerous.  specifically, you swallow any raised exceptions mysteriously and return a value instead.  theoretically this shouldn't raise anyway, but it's pretty risky.

instead, I would de-dent to be outside all the finally blocks, and
```py
return final_state.test_record.outcome == test_record.Outcome.PASS
```

openhtf/core/__init__.py
```diff
@@ -133,28 +140,14 @@ def _thread_proc(self):
...
 
-        # Perform initialization of some top-level stuff we need.
-        executor = self._make_phase_executor(exit_stack)
-
       # Everything is set, set status and begin test execution.  Note we don't
       # protect this with a try: block because the PhaseExecutor handles any
       # exceptions from test code.  Any exceptions here are caused by the
       # framework, and we probably want them to interrupt framework state
       # changes (like the transition to FINISHING).
+      self.test_state.mark_test_started()
```
grybmadsci: this has moved down fairly far, why did it move to be after plug initialization? that's a pretty significant change.  in particular, it means the test start_time_millis no longer includes plug init time, which may be nontrivial.  plug instantiation is "part of the test", so it shouldn't be excluded from the start time.

openhtf/core/phase_executor.py
```diff
@@ -172,6 +172,16 @@ def __init__(self, test_state):
     self.test_state = test_state
     self._current_phase_thread = None
 
+  def execute_start_trigger(self, phase_desc):
+    """Run the start trigger phase, and check that the DUT ID is set after.
+
+    Logs a warning if the start trigger failed to set the DUT ID.
+    """
+    self._execute_one_phase(phase_desc)
+    if not self.test_state.test_record.dut_id:
```
grybmadsci: prefer `is None` check to test for the unset condition.

openhtf/plugs/__init__.py
```diff
@@ -376,15 +379,21 @@ def _initialize_rpc_server(self):
...
-    rather than passed into the constructor (this is used primarily for unit
-    testing phases).
+    Instantiates plug types and saves the instances in self._plugs_by_type for
+    use in provide_plugs().
+
+    Args:
+      plug_types: Additional plug types may be specified here rather than passed
+                  into the constructor (this is used primarily for unit testing
+                  phases).
+      partial: If true, _only_ instantiate the plugs named in plug_types. Skip
```
grybmadsci: blech, passing this as an arg is pretty awkward.  how about making this method always only instantiate the provided ones, and if not provided, default to instantiating `self._plug_types`?  I don't think anywhere relies on the union behavior, but if it does, it should explicitly call `initialize_plugs()` first then initialize the extra ones.

edit: see below, maybe refactor this a little instead.

openhtf/plugs/__init__.py
```diff
@@ -405,7 +414,8 @@ def initialize_plugs(self, plug_types=None):
         self.tear_down_plugs()
         raise
       self.update_plug(plug_type, plug_instance)
-    self._initialize_rpc_server()
+    if not partial:
```
grybmadsci: ugh this was already kind of gross when you got here.  how about `initialize_plugs()` always takes specific plugs to initialize, and there's a separate `initialize()` that calls `self.initialize_plugs(self._plug_types)` and then `self._initialize_rpc_server()`?

openhtf/plugs/user_input.py
```diff
@@ -166,13 +158,12 @@ def respond(self, prompt_id, response):
...
-        message, text_input=text_input, timeout_s=timeout_s)
-    plug.tearDown()
-    return response
-  return trigger
+    message='Enter a DUT ID in order to start the test.', timeout_s=None):
+  """Return an OpenHTF phase for use as a prompt-based start trigger."""
+  @plugs.plug(prompts=UserInput)
+  def trigger_phase(test, prompts):
+    """Test start trigger that prompts the user for a DUT ID."""
+    test.test_record.dut_id = prompts.prompt(message=message, text_input=True,
```
grybmadsci: Awesome, I actually really like how this turned out, it's way more consistent with existing concepts/paradigms.

Kenadia:
The [comment about dut_id in TestApi](https://github.com/google/openhtf/blob/86ca49880cbe270842203f3de25c4f34f031e998/openhtf/__init__.py#L470) should be updated

openhtf/plugs/__init__.py
```diff
@@ -405,7 +414,8 @@ def initialize_plugs(self, plug_types=None):
...
+      self._initialize_rpc_server()
```
fahhem: I thought the RPC server is the whole point of this? Otherwise, how will we get plugs in test start to communicate with the frontend?

openhtf/core/__init__.py
jettisonjoe: I didn't quite realize the importance of some of the subtleties of moving this. I'll put it back up to right after the start trigger finishes.

openhtf/core/phase_executor.py
jettisonjoe: Roger that.

openhtf/plugs/__init__.py
jettisonjoe: Yeah, I think I can just factor out the union now.

jettisonjoe:
@grybmadsci, @fahhem, I addressed all the feedback, plus re-implemented support for lambda-based start triggers in the new paradigm. PTAL.

@fahhem, with this PR, you guys should have everything you need working to access prompt state from the Angular side again.

openhtf/__init__.py
```diff
@@ -217,8 +218,17 @@ def execute(self, test_start=None):
       self._test_desc.metadata['config'] = conf._asdict()
       self.last_run_time_millis = util.time_millis()
 
+      if not any((isinstance(test_start, PhaseDescriptor), test_start is None)):
```
grybmadsci: any() with only 2 things is kind of weird, probably more readable to use a binary operator and factor the outer `not` through.

openhtf/__init__.py
```diff
@@ -217,8 +218,17 @@ def execute(self, test_start=None):
...
+        @TestPhase()
+        def trigger_phase(test):
+          test.test_record.dut_id = test_start()
+        trigger = trigger_phase
+
```
grybmadsci: no blank line needed here, makes the `else:` harder to follow.

openhtf/__init__.py
grybmadsci: Something to consider, by checking for a `PhaseDescriptor` here, the start trigger phase *must* be decorated with something that does the wrapping.  This means that you can't use undecorated functions, which is inconsistent with test phases.

openhtf/__init__.py
```diff
@@ -241,6 +251,8 @@ def execute(self, test_start=None):
       finally:
         self._executor = None
 
+      return (final_state.test_record.outcome == test_record.Outcome.PASS)
```
grybmadsci: no parens

fahhem:
@jettisonjoe, you didn't reply to one of @grybmadsci's flattering comments so github didn't mark it as resolved, so that's why I dismissed it. If that wasn't the right thing to do, sorry.

LGTM!

openhtf/__init__.py
grybmadsci: this return is still inside the outermost `finally:`, should be de-dented one more level.

jettisonjoe:
@fahhem thanks, I think it just means I need to get into the habit of replying to even the flattering comments.

openhtf/core/__init__.py
```diff
@@ -107,18 +108,24 @@ def _thread_proc(self):
...
-        # Initial setup of exit stack and final cleanup of attributes.
         self._exit_stack = exit_stack
-
-      self.test_state = test_state.TestState(self._test_descriptor)
-      # Wait here until the test start trigger returns a DUT ID.  Don't hold
-      # self._lock while we do this, or else calls to stop() will deadlock.
-      # Create plugs while we're here because that may also take a while and
-      # we don't want to hold self._lock while we wait.
-      self.test_state.mark_test_started(self._wait_for_test_start())
+        exit_stack.callback(executor.stop)
```
grybmadsci: hm, you probably still want to `stop()` the executor *before* tearing down plugs.  technically there may still be a phase running if the executor isn't stopped.  This means pushing it onto the `exit_stack` after `tear_down_plugs` below.

openhtf/core/test_state.py
```diff
@@ -235,11 +235,8 @@ def set_status_from_phase_outcome(self, phase_outcome):
           self.is_finalized, phase_outcome)
     return self.is_finalized
 
-  def mark_test_started(self, dut_id):
-    """Set the TestRecord's dut_id and start_time_millis fields."""
-    assert self._status == self.Status.WAITING_FOR_TEST_START
```
grybmadsci: maybe assert that `start_time_millis is None`?  this assert was here to make sure we didn't accidentally blow away a previous `start_time_millis`, as that would be a particularly difficult bug to track down and diagnose.

openhtf/plugs/__init__.py
```diff
@@ -239,13 +241,14 @@ def discover(cls, host, port, timeout_s=xmlrpcutil.DEFAULT_PROXY_TIMEOUT_S):
...
-    proxy = xmlrpcutil.TimeoutProxyMixin((host, port))
+    proxy = xmlrpcutil.TimeoutProxyServer('http://%s:%s' % (host, port),
+                                          timeout_s=timeout_s, allow_none=True)
     seen = set()
     for method in proxy.system.listMethods():
       if not method.startswith('plugs.'):
         continue
       try:
-        _, plug_name, _ = method.split('.')
+        plug_name = method.rsplit('.', 1)[0].split('.', 1)[1]
```
grybmadsci: oh god, can you add a comment with an example of what `method` looks like to make this a bit more human parseable later?  i realize this was my derpy doing in the first place, sorry about that x.x

openhtf/plugs/user_input.py
```diff
@@ -166,13 +172,12 @@ def respond(self, prompt_id, response):
...
-    plug = UserInput()
-    response = plug.prompt(
-        message, text_input=text_input, timeout_s=timeout_s)
-    plug.tearDown()
-    return response
-  return trigger
+    message='Enter a DUT ID in order to start the test.', timeout_s=None):
+  """Return an OpenHTF phase for use as a prompt-based start trigger."""
+  @plugs.plug(prompts=UserInput)
+  def trigger_phase(test, prompts):
```
grybmadsci: I think you're still going to want to pass a timeout, via normal phase options, otherwise this is going to have the default phase timeout (5 min i think, if not overridden via flag).

grybmadsci:
that's right, accept my flattery and LIKE IT.

few more comments, mostly nits, but 1 or 2 non-nit considerations :) should
be good to go once those are addressed.

~madsci

openhtf/plugs/__init__.py
```diff
@@ -193,6 +194,7 @@ def __init__(self, host, port, plug_name, session):
     # Grab exactly the attrs we want, the server has all plugs', not just ours.
     for method in self.system.listMethods():
       name, attr = method.rsplit('.', 1)
+      name = name.split('.', 1)[1]
```
fahhem: so method here is also `plugs.<plug_class>.<attr>`? Maybe a function that parses this into a namedtuple that's called here and below on line 251 would be the better option?

openhtf/__init__.py
fahhem: Yeah to @grybmadsci, this is much cleaner:

```py
if not isinstance(test_start, PhaseDescriptor) and test_start and is None:
```

openhtf/plugs/user_input.py
fahhem: Actually, this could become a single phase that `prompt_for_test_start()` just returns `WithArgs()` of, for backwards compatibility

openhtf/plugs/__init__.py
jettisonjoe: Hmm, alright, let's give that a shot.

openhtf/plugs/user_input.py
jettisonjoe: Can you clarify the backwards compatibility bit?

jettisonjoe:
Damn, -1.1% coverage. OpenHTF, IOU some test coverage.

PR #446, opened 2016-10-05
Support for string formatting of name and docstring when using with_plugs

kdsudac:
I realized that passing variables to `PhaseOptions.name`, `Measurement.name` and `Measurement.docstring` for string formatting was not supported by with_plugs.

The previous implementation of `PhaseDescriptor.with_args` also assumed that all necessary variables would be present when it called otherwise string.format raises a KeyError.

Changes: 
* `safe_format` in util that allows partial substitution of variables into the template string, similar to [string.Template.safe_substitute](https://docs.python.org/2/library/string.html#string.Template.safe_substitute).
* Renamed method names on `PhaseOptions` and `Measurement` from `with_args` -> `format_strings`.  Reason: I didn't want to call `with_args` from within `with_plugs` and all these two methods were doing were string formatting so `with_args` was a misnomer.
* Added unit test for with_plugs
* Updated examples/with_plugs.py

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -303,8 +303,8 @@ class PhaseOptions(mutablerecords.Record('PhaseOptions', [], {
   """Options used to override default test phase behaviors.
 
   Attributes:
-    name: Override for the name of the phase. Can be formatted in many different
-        formats defined in util.format_string.
+    name: Override for the name of the phase. Can be formatted in several
+        different ways as defined in util.format_string.
     timeout_s: Timeout to use for the phase, in seconds.
     run_if: Callback that decides whether to run the phase or not.
     requires_state: If True, pass the whole TestState into the first argument,
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -303,8 +303,8 @@ class PhaseOptions(mutablerecords.Record('PhaseOptions', [], {
   """Options used to override default test phase behaviors.
 
   Attributes:
-    name: Override for the name of the phase. Can be formatted similar to
-        measurements using with_args() to distinguish multiple similar phases.
+    name: Override for the name of the phase. Can be formatted in many different
+        formats defined in util.format_string.
     timeout_s: Timeout to use for the phase, in seconds.
     run_if: Callback that decides whether to run the phase or not.
     requires_state: If True, pass the whole TestState into the first argument,
@@ -322,7 +322,8 @@ class PhaseOptions(mutablerecords.Record('PhaseOptions', [], {
       pass
   """
 
-  def with_args(self, **kwargs):
+  def format_strings(self, **kwargs):
+    """String substitution of name."""
     return mutablerecords.CopyRecord(
         self, name=util.format_string(self.name, **kwargs))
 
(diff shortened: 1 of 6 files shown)


openhtf/util/__init__.py
```diff
@@ -106,6 +108,23 @@ def __get__(self, instance, owner):
...
+  for tag, var in re.findall(r'(\{(.*?)\})', output):
+    root = var.split('.')[0]  # dot notation
+    root = root.split('[')[0]  # dict notation
+    if root in kwargs:
+      output = output.replace(tag, tag.format(**{root: kwargs[root]}))
+
+  return output
+
 def format_string(target, **kwargs):
   """Formats a string in any of three ways (or not at all).
```
kdsudac: Note: util.format_string and util.safe_format don't support string.format via unnamed *args.

I think this is appropriate since it's only used by the framework on kwargs passed.

i.e. Using standard python string format, this works:
```
'{} {}'.format('foo', 'bar') # returns 'foo bar'
```

But this won't:
```
util.format_string('{} {}', 'foo', 'bar')  # TypeError: format_string() takes exactly 1 argument (3 given)
```

test/util/util_test.py
```diff
@@ -43,3 +44,31 @@ def test_time_expired_false(self, mock_time):
...
+    res = util.safe_format(res, pear=pear)
+    self.assertEqual('Apples are red and sweet. Pears are green and tart. '
+      'Oranges are {orange_color} and {orange_taste}.', res)
+
+    # Format rest of string
+    res = util.safe_format(res, orange_color='orange', orange_taste='sour')
+    self.assertEqual('Apples are red and sweet. Pears are green and tart. '
+      'Oranges are orange and sour.', res)
+
+    #  The original text has not changed
```
jettisonjoe: If you want to test that the original text indeed hasn't changed, I think an explicit check against a copy is better than a check for inequality here. If some but not all of the formatting steps mutate the original, this test as written would still pass.

openhtf/util/__init__.py
```diff
@@ -106,6 +108,23 @@ def __get__(self, instance, owner):
     return self._func(owner)
 
 
+def safe_format(target, **kwargs):
```
jettisonjoe: Minor nit about naming here. To me the term "safe" in the context of string formatting hints towards the escaping of special characters (such as in preventing attacks like SQL injection). I'm not sure I have a great alternative here (maybe `partial_format`), but also don't want users to see this and get the wrong idea. Even if it would technically be their own fault for using code they don't understand.

openhtf/__init__.py
```diff
@@ -303,8 +303,8 @@ class PhaseOptions(mutablerecords.Record('PhaseOptions', [], {
   """Options used to override default test phase behaviors.
 
   Attributes:
-    name: Override for the name of the phase. Can be formatted similar to
-        measurements using with_args() to distinguish multiple similar phases.
+    name: Override for the name of the phase. Can be formatted in many different
+        formats defined in util.format_string.
```
jettisonjoe: wording nit: s/formatted in many different formats/formatted several different ways/ (or the like).

test/util/util_test.py
kdsudac: Done

openhtf/__init__.py
kdsudac: Done

openhtf/util/__init__.py
kdsudac: Done. Original name was following `safe_substitue` method... but agree it's usually associated with escaping.  Went with partial_format.

PR #450, opened 2016-10-14
Don't mess with stdin when it's not a terminal

fahhem:


openhtf/plugs/user_input.py
```diff
@@ -96,14 +97,32 @@ def run(self):
...
-              response = sys.stdin.readline().rstrip()
-              self._callback(response)
+              new = os.read(sys.stdin.fileno(), 1024)
+              if not new:
+                # Hit EOF!
+                if not sys.stdin.isatty():
+                  # We're running in the background somewhere, so the only way
+                  # to respond to this prompt is the UI. Let's just wait for
+                  # that to happen now. We'll give them a week :)
+                  print "Waiting for a non-console response."
```
jettisonjoe: Where will the print go if we're in the background? Is the (undocumented) assumption that the user will redirect stdout? Maybe consider logging instead of printing?

openhtf/plugs/user_input.py
```diff
@@ -96,14 +97,32 @@ def run(self):
...
+                  time.sleep(60*60*24*7)
+                else:
+                  # They hit ^D (to insert EOF). Tell them to hit ^C if they
+                  # want to actually quit.
+                  print "Hit ^C to exit."
```
jettisonjoe: Personal pref, but I like `CTRL+c` for this as I find it more readable than `^C`.

openhtf/plugs/user_input.py
fahhem: We printed the message itself, so I'd rather this went right under that message than potentially redirected to go with logging messages somewhere else.

We only checked that stdin isn't a TTY, so stdout might go to a terminal still, or it might not.

openhtf/plugs/user_input.py
fahhem: Will do

openhtf/plugs/user_input.py
jettisonjoe: Alright, that's a fair point.

Issue #456, opened 2016-11-11
Web Frontend throws exception at end of test


agdeepak:
When the test finishes, there is an exception thrown in the console running the Web front-end. The frontend running in browser seems to be working though. 

To reproduce it, 
Run local front end
Run hello world py test in /example
Finish the hello world py test
Exception is thrown in the frontend console window

Screenshot is attached for reference. Issue was produced on Win7 but not validated on Linux yet.
![openhtf-frontend-example-endtestexception](https://cloud.githubusercontent.com/assets/16326136/20201003/b6610422-a769-11e6-938b-3bea19b45295.PNG)

PR #458, opened 2016-11-11
Fix bug when prompt is none (#456)

Kenadia:
See #456 for the repro steps for the bug.

Note: This took me a while to debug the first time I ran into it. Take a look at the stack trace in the screenshot referenced by #456—it's hard to tell where the error really occurred. Is there something we can do to better surface bugs in RPC calls?


jettisonjoe:
Thanks and good catch. I agree that errors that happen on the far side of an RPC are pretty tricky to debug right now and should be improved.

Issue #382, opened 2016-07-21
Finish plumbing prompts through to the web gui using station_api.


jettisonjoe:

fahhem:
Aw dang, I did a bunch of work to migrate to 1.0, then remembered this is broken in 1.0... I'll have to fix this before I can send that out and it changes every single file in my team's purview...

jettisonjoe:
I already started a decent portion of the work for this. I can finish it, or you can take it over. I'm on vacation next week, so I likely wouldn't submit my solution until sometime the week after. I'm cool either way. Up to you.

fahhem:
Can you send a pr with your initial work so I don't start from scratch?
I'll take over while you're on vacation

jettisonjoe:
I sent #443. Turns out it's a small amount of code, even though mentally it felt like a lot of deliberation to arrive at that approach. Open to feedback and alternatives as well.

jettisonjoe:
@fahhem I think you guys should have all you need in order to make this work from the Angular side. Hit me up if I'm wrong about that or if you need a hand getting it done.

fahhem:
@jettisonjoe Thanks. We were going to start this last week, but something last-minute came up. Hopefully in the next week that'll settle and then we can fix this from the Angular2 side and finally have prompts in 1.0 working :)

fahhem:
@Kenadia fixed this a while back

PR #460, opened 2016-11-16
Fix remote plugs and frontend prompts (#382)

Kenadia:
This PR does the remaining work to fix prompts in the frontend, using remote plugs. (#382)

A few things were broken in remote plugs that had to be fixed first. The most significant change I made was to use HTTP POST requests instead of web sockets for communicating with remote plugs (see `openhtf/plugs/__init__.py`). My reasoning is below—let me know if I misunderstood something and we should be using web sockets instead.
- RemotePlug was designed to create a separate endpoint for each plug instance. Connecting to those endpoints with web sockets entails creating a new SubscriptionService instance on the frontend for every endpoint. Managing all those subscriptions, and identifying when to subscribe and unsubscribe, adds a lot of complexity and overhead to the frontend.
- We already have a web socket open to each station, which keeps the test state up to date, including plug information. There seems to be no need for plugs to independently send messages to the frontend.
- In the web socket model, it was unclear how to handle remote plug errors: see the two TODOs in `openhtf/plugs/__init__.py`. With HTTP it is clear how to handle this.

## Example

OpenHTF test code: http://pastebin.com/ijrDSqWM

prompts


openhtf/output/web_gui/src/app/station.service.ts
```diff
@@ -98,7 +102,9 @@ export class StationService extends SubscriptionService {
   onmessage(msg) {
     this.reachable = true;
     let parsedData = JSON.parse(msg.data);
-    let test_uid = parsedData.test_uid, state = parsedData.state;
+    let test_uid = parsedData.test_uid;
+    let state = parsedData.state;
+    state.test_uid = test_uid  // Add test_uid to the test state.
```
jettisonjoe: Not a huge fan of stashing this directly on state, because it loses the parity between state here and the state that was sent to us. But I don't have a better suggestion short of adding a layer above both. Just a note, not necessarily a blocker.

jettisonjoe:
I'm not sure if I broke this earlier or you broke it here, but my web GUI isn't getting consistently updated anymore when the test state changes. In fact, it's only updating a small fraction of the time. Has anyone else noticed this?

jettisonjoe:
I'm also not seeing pending phases rendered in the web GUI anymore. And in fact, the prompt isn't showing for me either in the prebuilt, even though I'm on your `prompts` branch.

Kenadia:
@jettisonjoe I just cloned a fresh copy of the repo, pulled this PR, installed using setup.py, and ran the frontend and hello_world scripts and everything looked fine (using `prebuilt`).

Are you getting any errors in any of the frontend or backend consoles?

Is this what you meant by “pending phases”? I'm using `examples/hello_world.py`

screen shot 2016-11-18 at 12 47 52 pm

fahhem:
A couple notes on the overall change:

* Any chance we could put the prompt above the phases to match 0.9 behavior? By keeping it in the same place between phases, operators know exactly where to look instead of needing to scroll or look around.
* Would it be possible to use one sockjs connection for everything? Instead of hitting a new URL or a new sockjs connection for plugs, you instead hit the same sockjs connection but routed by what is currently the URL? That way we keep things snappy (when websockets work)

jettisonjoe:
In particular, there should be a `hello_world` phase underneath `trigger_phase`, but grayed-out. I pulled and re-ran from your fork, and I'm seeing prompts now and updates are working (I'm not sure what I screwed up earlier), but for sure there are no more pending phases shown.

openhtf/plugs/__init__.py
```diff
@@ -196,52 +197,70 @@ def tearDown(self):
...
         if hasattr(type(self), remote_attr.name):
-          _LOG.warning('Skipping predefined attribute "%s" for remote access.',
+          _LOG.warning('Skipping preexisting remote plug attribute "%s".',
                        remote_attr.name)
         else:
-          setattr(self, remote_attr.name,
-                  functools.partial(self.__request, method))
+          setattr(self, remote_attr.name, getattr(self, method))
+
+  def list_methods(self):
```
fahhem: Why is this function here? `self.system` should work since `__getattr__` won't get invoked for attributes that exist so this should be the same as just doing `self.system.listMethods()` in other functions directly

fahhem: Ah, it looks like this does caching? Can we store the cache as `self._cached_methods` to make that clear?

openhtf/output/web_gui/src/app/station.html
```diff
@@ -26,6 +26,9 @@
           <metadata id="metadata" [metadata]="test.test_record.metadata">
           </metadata>
         </div>
+        <prompt [prompt]="test.plugs.plug_states['openhtf.plugs.user_input.UserInput']"
```
fahhem: Can we put this above either `<phase-listing>` or `div.test-content`?

openhtf/output/web_gui/__init__.py
```diff
@@ -341,52 +344,49 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
...
+    Notifies the StationPubSub of the state update, and updates the server's URL
+    handlers with any newly available remote plugs.
     """
     if state is None:
       return
 
-    def _make_sockjs_url(rel_url):
-      """SockJS URL is /plugs/<host>/<port>/<test_uid>/<plug_name>"""
-      return r'/plugs/%s/%s/%s/%s' % (hostport + (test_uid, rel_url))
+    def make_plug_url(plug_name):
```

Kenadia:
Confirmed that the missing pending phase issue is present on master, opened as issue #468

openhtf/plugs/__init__.py
Kenadia: 1. That's actually not what happens, and this was a bug before. `system` does not exist as an attribute directly on the proxy server, and so `self.system` results in a call to `__getattr_. I'll add a comment to mention this.

2. I'll rename the variable to `self._cached_methods`

openhtf/output/web_gui/src/app/station.service.ts
Kenadia: I've added a commit `Keep test UID separate from state` Let me know what you think!

fahhem:
I'd like to wait for @jettisonjoe or @grybmadsci to look again before I merge this, but it seems good

fahhem:
Woohoo!

PR #461, opened 2016-11-17
have enable_station_discovery config control servers in station_api

wallacbe:


jettisonjoe:
I think the problems you're having might be solved by just explicitly setting `station_api_port` to `None` if you want to disable the station API.

wallacbe:
setting null instead of None in yaml results in correct behavior

PR #464, opened 2016-11-23
Allow station_api_port to be 0

fahhem:
If you set it to 0, then it means "auto-choose unused port" to most socket APIs

We should also consider making 0 the default instead of 8888


grybmadsci:
LGTM

jettisonjoe:
btw, I'm in favor of making 0 the default.

PR #467, opened 2016-11-25
Lazy load test code based on the new capture_source option

### LINKED ISSUE #360: Provide an option to disable source code capturing.

fahhem:
Fixes #360 

I had to make the default False because `configure()` gets called in `Test.__init__()` and I didn't want to make tests for authors who didn't want the source captured load all the source files, make copies of all the strings, then throw it all away in the span of 2 function calls.


openhtf/__init__.py
```diff
@@ -159,21 +164,27 @@ def add_output_callbacks(self, *callbacks):
...
+      # First, we copy the phases with the real CodeInfo for them.
+      phases = [
+        mutablerecords.CopyRecord(
+            phase, code_info=test_record.CodeInfo.for_function(phase.func))
+        for phase in self._test_desc.phases]
+
+      # Then we replace the TestDescriptor with one that stores the test
+      # module's CodeInfo as well as our newly copied phases.
+      self._test_desc = self._test_desc._replace(
+          code_info=self.code_info, phases=phases)
```
jettisonjoe: I have two qualms with using `_replace` on the descriptor as done here. First, I think it would be better style-wise to explicitly export `_replace` as `replace` on the `TestDescriptor` class.

Second, and perhaps more concerning, is that any references to `self._test_desc` that are grabbed or given out before this `_replace` will be defunct after it. Right now anyone can grab it through the `descriptor` property, and there's nothing to stop someone grabbing the original one after instantiation, but before `configure()` is called. One thought is to maybe use a mutable record here instead, but I think we specifically chose not to because the `TestDescriptor` is intended to be static between runs of the same test.

fahhem: `namedtuple` exports it as `_replace` to not conflict with potential attribute names (since they're enforced as not starting with `_`). We could separate out the code info into separate objects instead of inside the `PhaseDescriptor`/`TestDescriptor`?

We have some other options:
* Make `capture_source` a kwarg to `Test` so we know whether to grab the code info as early as that.
* Store a `TestDescriptor` in an unconfigured state, maybe in `self._initial_test_desc` then only after calling `configure()` they get access to it. `descriptor` is documented as not changing across `execute()` calls, not `configure()` calls.
* Separate the `CodeInfo` to be stored in a separate set of objects than from `TestDescriptor`, though it does make sense to put it in the 'static test information' :/

fahhem:
Yeah, sometimes the captured modules/functions are wrong for various reasons. I'm not sure how much we want to step into that rabbit hole though

openhtf/__init__.py
fahhem: After looking at it some more, storing it in an unconfigured state isn't possible since we call `configure()` inside of `__init__()` anyway. I'd rather not separate out `CodeInfo` since it's pretty relevant.

I think our options are instead:

* Make `capture_source` a special `metadata` attribute rather than a `TestOptions` attribute, which sort of makes sense since `metadata` is relevant to the `TestDescriptor` while `TestOptions` is relevant to a particular test setup.
* Make `capture_source` a config key instead and require it to be set before `Test` is created.
* A few more that I didn't like but I just thought of the config key and I think it's the best one so I'm going to do that

fahhem:
Okay, ready for another look

PR #469, opened 2016-11-30
Add back pending phases

### LINKED ISSUE #468: Pending/upcoming phases are missing from the frontend
Example:

screen shot 2016-11-29 at 5 25 18 pm

We expect to see a pending phase “hello_world” underneath “trigger_phase”.

fahhem:
Fixes #468


fahhem:
This also makes the pending phases into more normal `PhaseRecord`s so you can see the measurements it's going to set and such.

openhtf/output/web_gui/prebuilt/templates/station.phases.html
```diff
@@ -94,20 +94,22 @@
...
-          {{phase[0]}}
-        </div>
-        <div class="cycle-time">
-        </div>
-        <div class="status">
-          PENDING
-        </div>
+        <div class="name">{{phase.name}}</div>
+        <div class="cycle-time"></div>
+        <div class="status">PENDING2</div>
```
jettisonjoe: _PENDING2: The Future's Revenge_

grybmadsci:
Ahh, I'm inclined to agree with joe's hesitation around tracking the sequence-o-phases in two places and keeping them in sync, and feel pretty strongly about not mucking with core internals for the sake of the frontend unless needed.

As an alternative, can this logic be moved into the frontend server?  We can expose more information via the station_api if we need to (i don't recall if we currently provide visibility into "all phases"), but as written this is a bit much of a change to the core code.

openhtf/core/test_record.py
```diff
@@ -71,16 +71,15 @@ def add_outcome_details(self, code, description=''):
 class PhaseRecord(  # pylint: disable=no-init
     mutablerecords.Record(
         'PhaseRecord', ['name', 'codeinfo'],
-        {'measurements': None,
-         'start_time_millis': None, 'end_time_millis': None,
-         'attachments': dict, 'result': None})):
+        {'start_time_millis': None, 'end_time_millis': None,
+         'measurements': dict, 'attachments': dict, 'result': None})):
```
grybmadsci: hm, why is this changing? seems unrelated?  measurements now defaults to a dict?

openhtf/core/test_record.py
```diff
@@ -53,8 +53,8 @@ class TestRecord(  # pylint: disable=no-init
         'TestRecord', ['dut_id', 'station_id'],
         {'start_time_millis': None, 'end_time_millis': None,
          'outcome': None, 'outcome_details': list,
-         'code_info': None,
-         'metadata': dict,
+         'code_info': None, 'metadata': dict,
+         'pending_phases': list,
```
grybmadsci: TestRecord is intended to be an append-only output mechanism, not a place to store transient state.  I'd rather see a station_api mechanism for including a list of all the phase names in the test metadata, and then have the frontend server use that in coordination with phase completion notifications to internally track pending phases.

openhtf/output/web_gui/prebuilt/scripts/app.bundle.js
```diff
@@ -41193,7 +41193,7 @@ webpackJsonp([0],[
 	    Prompt.prototype.sendResponse = function (input, id) {
 	        var ip = this.routeParams.get('ip');
 	        var port = this.routeParams.get('port');
-	        var response = input.value || '';
+	        var response = input ? input.value : '';
```
grybmadsci: seems like an unrelated change?  can we pull unrelated bugfixes out?

test/output/callbacks/record.json
```diff
@@ -41,7 +41,7 @@
       "station_api_bind_address": "0.0.0.0", 
       "station_api_port": 8888, 
       "station_discovery_string": "OPENHTF_DISCOVERY", 
-      "station_id": "jethier.mtv.corp.google.com", 
+      "station_id": "fahhem.mtv.corp.google.com",
```
grybmadsci: we should probably not be exporting google's internal network names to an open github repo.  i don't think it realllly matters, but can we anonymize this string?

test/output/callbacks/record.json
```diff
@@ -282,15 +298,15 @@
...
         "phase_result": "CONTINUE"
       }, 
-      "start_time_millis": 1471981975705
+      "start_time_millis": 1480507463461
     }
   ], 
-  "start_time_millis": 1471981967848, 
-  "station_id": "jethier.mtv.corp.google.com"
+  "start_time_millis": 1480507454324, 
+  "station_id": "fahhem.mtv.corp.google.com"
```
grybmadsci: same

grybmadsci:
I'm not too attached to putting this logic in the frontend server, per se.  Mostly, I'd like to avoid modifications to the core code, especially addition of secondary data structures that have to be kept in sync with existing ones.  I think this logic could move to a few other places, I'll defer to you and joe on where else makes the most sense, but it could be the test-side station_api, frontend server, or all the way into the frontend client.

jettisonjoe:
I think the best solution in terms of separation of concerns (but also possibly the most work) is to put the logic to compute pending phases all the way in one of the frontend client components. Between the full list of phases and the test record with completed phases, the frontend client should have everything it needs to make that computation. But since you guys are the ones with the cycles to do the work right now, I'm pretty fine with wherever it ends up, as long as @grybmadsci is cool with it too.

openhtf/output/web_gui/prebuilt/scripts/app.bundle.js
fahhem: That's because the prebuilt is old. Also, you're one to talk about unrelated changes in a PR :P

openhtf/core/test_record.py
fahhem: This is documenting the type, really. This becomes a dict after the phase has executed, which is right around the time this record is created anyway, and nobody checks the measurements for being `None` (or at all). Rather, I was confused as to the type of measurements (is it a list here like it is elsewhere?) and wanted to document that

openhtf/output/web_gui/prebuilt/scripts/app.bundle.js
grybmadsci: haha guilty as charged ^.^ trying to reign in on that all around, myself included, now that we're all official-like :)

fahhem:
This is a regression that I'm fixing. Can we have the philosophical changes done after this is in? I'm not against them in principle, but I can't upgrade to 1.0 until we're at least feature parity on the frontend, and this is a pretty egregious regression (along with prompts). I promise after February (our next milestone/build) at the latest March to come back and rip this out. I'm likely going to propose some changes for test/phase execution at that point, and potentially a way to replace core pieces this year.

fahhem:
Reverting this in favor of the frontend having all the PhaseDescriptors

PR #475, opened 2016-12-06
Add RemotePhaseDescriptor to provide remote access to phase descriptors.

grybmadsci:
This is required to support display of pending phases in the frontend.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -138,10 +138,10 @@ class Test(object):
     station_api.start_server()
 
   @classmethod
-  def state_by_uid(cls, test_uid):
-    """Get TestState of a test by UID.
+  def from_uid(cls, test_uid):
+    """Get Test by UID.
 
-    Returns: TestState of currently running test, given by test_uid.
+    Returns: Test object, given by UID.
 
     Raises:
       UnrecognizedTestUidError: If the test_uid is not recognized.
@@ -149,7 +149,7 @@ class Test(object):
     test = cls.TEST_INSTANCES.get(test_uid)
     if not test:
       raise UnrecognizedTestUidError('Test UID %s not recognized' % test_uid)
-    return test.state
+    return test
 
   @property
   def uid(self):
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -428,6 +428,14 @@ class PhaseDescriptor(mutablerecords.Record(
     retval.options.update(**options)
     return retval
 
+  def _asdict(self):
+    asdict = {
+        key: data.convert_to_base_types(getattr(self, key), ignore_keys=('cls',))
+        for key in self.optional_attributes
+    }
+    asdict.update(name=self.name, doc=self.doc)
+    return asdict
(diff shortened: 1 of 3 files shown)


openhtf/core/station_api.py
```diff
@@ -213,7 +221,7 @@ class RemoteTest(mutablerecords.Record('RemoteTest', [
     # Timestamps (in milliseconds) that we care about.
     'created_time_millis', 'last_run_time_millis'], {
         # Cache last known state (a RemoteState) so we can detect deltas.
-        'cached_state': None})):
+        'cached_state': None, 'cached_phase_descriptors': None})):
```
fahhem: optional: If you set it to `'cached_phase_descriptiors': list` then the type is documented. An extra empty list is 'created' is the only downside I see

openhtf/core/station_api.py
```diff
@@ -561,6 +579,11 @@ def list_tests(self):
     _LOG.debug('RPC:list_tests() -> %s results', len(retval))
     return retval
 
+  def get_phase_descriptors(self, test_uid):
+    """Get phase descriptors for the given test UID, or None if invalid UID."""
+    phases = openhtf.Test.from_uid(test_uid).descriptor.phases
+    return [{'name': phase.name, 'doc': phase.doc} for phase in phases]
```
fahhem: why not return the phase descriptors themselves? They're not that heavyweight and the information can be useful to the frontend, such as measurement names

fahhem: well, the `_asdict` version of the descriptors

openhtf/core/station_api.py
grybmadsci: done

grybmadsci:
PTAL, non-trivial update to include all optional fields of `PhaseDescriptor` in `RemotePhaseDescriptor`

openhtf/__init__.py
```diff
@@ -428,6 +428,14 @@ def wrap_or_copy(cls, func, **options):
     retval.options.update(**options)
     return retval
 
+  def _asdict(self):
```
fahhem: don't we have something for mutablerecords already? If the only reason for all of this is to ignore the non-serializable cls key, I think we should just add a `if isinstance(obj, type)` and skip all class definitions in `convert_to_base_types`

grybmadsci: also need to remove `func` and add `name` and `doc` instead.  we could move this up a level though since these semantics are specific to the `station_api`?

i'd be okay putting the type-skipping in `convert_to_base_types` but that only really removes the `ignore_keys=('cls',)` bit.

fahhem: Ah, I didn't realize that name and doc weren't direct attributes. I could see `func` and `cls` getting removed because they're not serializable in `convert_to_base_types`, so that would make this function's behavior more obvious:

```py
def _asdict(self):
  asdict = data.convert_to_base_types(self)
  asdict.update(name=self.name, doc=self.doc)
  return asdict
```

Issue #483, opened 2017-01-07
The web GUI is not notified of changes to plug states


Kenadia:
This issue was initially detected as part of issue #479.

The Station API is not notified of changes to plug states. Therefore the frontend web GUI is not notified of changes to plug states. The result is that whether or not a prompt shows up on the front end is a race condition. If the prompt is at the start of a phase, it will usually show up.

Note that `StationAPI` has a `wait_for_plug` method but it's not currently used anywhere.

Below is a short script that demonstrates the problem. If you run this script, the first prompt should show up, but the second prompt should not show up due to the delay. If you refresh the page the second prompt should then show up.

```python
import openhtf
import time

from openhtf.plugs import user_input


@openhtf.plugs.plug(user_input=user_input.UserInput)
def delayed_prompt_phase(test, user_input):
  test.test_record.dut_id = 'my-dut-id'
  print 'Waiting one second.'
  time.sleep(1)
  result = user_input.prompt('Prompt: enter some input', text_input=True)


if __name__ == '__main__':
  openhtf.Test(delayed_prompt_phase, delayed_prompt_phase).execute()
```

fahhem:
This is basically because the plug has no `notify_update` mechanism. The `wait_for_plug` method is a way to spend 100% CPU on it, by asking the plug to serialize it's state in a tight loop, but I think we can do better.

What if we created a `FrontendAwareBasePlug` or similar plug so that it was opt-in, but if they did opt-in they now had to call `self._state_updated.set()` every time the state is updated. Then, the base-class would have `self._state_updated = threading.Event` and a way for the Station API to subscribe to it, in the same way it subscribes to the event underlying `notify_update` in the test executor.

Issue #479, opened 2016-12-19
Have to refresh browser to see prompt, get timeout when entering text


wallacbe:
Cloned at commit  6981216f1df2b8d6c094b917e6c6b74d2d210b77
Started webgui and then started all_the_things.py

Observed that I'm able to enter barcode through gui to start the test.  I then have to refresh the web browser to see the "What's the widget type" prompt.  I then try to enter text but I then see a timeout in the web server logs:

python -m openhtf.output.web_gui
Starting openhtf web gui server on http://localhost:12000.
ERROR:tornado.application:Uncaught exception POST /plugs/172.19.60.136/8888/13161:1482189226832:139848655111696/openhtf.plugs.user_input.UserInput (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:12000', method='POST', uri='/plugs/172.19.60.136/8888/13161:1482189226832:139848655111696/openhtf.plugs.user_input.UserInput', version='HTTP/1.1', remote_ip='127.0.0.1', headers={'Origin': 'http://localhost:12000', 'Content-Length': '73', 'Accept-Language': 'en-US,en;q=0.8', 'Accept-Encoding': 'gzip, deflate, br', 'Host': 'localhost:12000', 'Accept': '*/*', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36', 'Connection': 'keep-alive', 'Referer': 'http://localhost:12000/station/172.19.60.136/8888', 'Content-Type': 'application/json'})
Traceback (most recent call last):
  File "/home/wallaceb/source/venv/local/lib/python2.7/site-packages/tornado-4.3-py2.7-linux-x86_64.egg/tornado/web.py", line 1443, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "openhtf/plugs/__init__.py", line 228, in post
    result = getattr(self.remote_plug, method)(*args)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1233, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1587, in __request
    verbose=self.__verbose
  File "/usr/lib/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1303, in single_request
[...]

PR #484, opened 2017-01-07
Do not re-initialize PlugManager's XML-RPC server

Kenadia:
This PR changes PlugManager so that it reuses its RPC server instead of shutting it down when initialize_plugs is called a second time. This addresses the first problem described in https://github.com/google/openhtf/issues/479#issuecomment-271060500.

This partially addresses #479 to the extent that `all_the_things.py` and other tests that use a test start and prompts should *usually* run as expected with the caveat that there is a race condition described in #483. Tests with more complicated front-end plug use would expect more serious problems.

### DIFF
--- a/openhtf/core/station_api.py
+++ b/openhtf/core/station_api.py
@@ -38,16 +38,16 @@ This API is based around the following hierarchy of abstractions:
       +------------------------------------------------------------------+
 
 A station discovery mechanism is implemented to provide multicast-based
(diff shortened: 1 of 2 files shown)


fahhem:
LGTM

Looks like you'd re-register plugs that were used in test start and in phases, but it's not a big deal since the name and method is exactly the same and it's not repeated (only one duplication per PlugManager instance)

Issue #485, opened 2017-01-10
Add dut_serial upper option in prompt_for_test_start - feature request


wallacbe:
Add dut_serial upper option in prompt_for_test_start to convert to uppercase

PR #486, opened 2017-01-10
add serial to upper option to prompt_for_test_start

wallacbe:
https://github.com/google/openhtf/issues/485

### DIFF
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -194,13 +194,23 @@ class UserInput(plugs.BasePlug):
 
 
 def prompt_for_test_start(
-    message='Enter a DUT ID in order to start the test.', timeout_s=60*60*24):
-  """Return an OpenHTF phase for use as a prompt-based start trigger."""
+    message='Enter a DUT ID in order to start the test.', timeout_s=60*60*24,
+    validator=lambda sn: sn):
+  """Return an OpenHTF phase for use as a prompt-based start trigger.
+
+    Args:
+      message: The message to display to the user.
+      timeout_s: Seconds to wait before raising a PromptUnansweredError.
+      validator: Function used to validate or modify the serial number.
+  """
+
   @PhaseOptions(timeout_s=timeout_s)
   @plugs.plug(prompts=UserInput)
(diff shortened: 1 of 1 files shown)


wallacbe:
PTAL

openhtf/plugs/user_input.py
```diff
@@ -194,13 +194,15 @@ def respond(self, prompt_id, response):
 
 
 def prompt_for_test_start(
-    message='Enter a DUT ID in order to start the test.', timeout_s=60*60*24):
+    message='Enter a DUT ID in order to start the test.', timeout_s=60*60*24,
+    upper=False):
```
fahhem: I'd prefer a validation hook that's a little more general and you can provide `string.upper` but in the future you may want to validate serials more thoroughly.

Take `validator=None` as an argument and put this below:

```py
if validator is None:
  validator = lambda sn: sn
test.test_record.dut_id = validator(dut_id)
```

jettisonjoe:
My thinking on this is that it seems a bit random to add an 'upper' feature to just this start trigger. Mainly it seems like a project-specific need, and `prompt_for_test_start` is already a _very_ thin wrapper around the prompt mechanism to begin with. Some alternate approaches I think would be better:

1. Define your own start trigger that wraps prompting just like `prompt_for_test_start` does, but `upper`'s the text (as if your `upper` arg here were true).
2. Use `prompt_for_test_start` as-is, but reach into the test record later and upper the `dut_id` by hand.
3. Instead of a specific `upper` argument, add a `format_callback`. If the arg is `None` (the default), the start trigger does what it currently does. Otherwise, the start trigger sets `dut_id` to the result of `format_callback(raw_dut_id)`. I can also imagine variations on this where `format_callback` is set as an attribute of the start trigger.
4. Some pipe-style or other more general dut_id formatting approach even more clever than 3 that I haven't thought of.

wallacbe:
PTAL

wallacbe:
Done, added docstring

wallacbe:
Thanks, PTAL maintainer

fahhem:
LGTM

fahhem:
Oops, I forgot to rename the commit!

PR #488, opened 2017-01-18
Make validators.InRange work with None value

kdsudac:


openhtf/util/validators.py
```diff
@@ -85,6 +85,8 @@ def __init__(self, minimum=None, maximum=None):
       self.maximum = maximum
 
     def __call__(self, value):
+       if value is None:
```
fahhem: got the number of spaces off by one :)

kdsudac: Fixed.  That's what I get for using the built-in github editor for a 2 line change. :)

Issue #489, opened 2017-01-23
Writing out log files take too long due to filename formatting


arsharma1:
Writing out the log files takes too long because the filename formatting converts the entire test_record (including all the phases and their possibly large attachments) into dictionaries an extra time.  In `openhtf/output/callbacks/__init__.py`'s `OutputToFile.open_output_file`, the call to `data.convert_to_base_types` should at least include `ignore_keys=('code_info', 'phases', 'log_records')`.

PR #490, opened 2017-01-23
output callbacks: Restrict keys used for filename

arsharma1:
Restrict the test_record keys used for formatting filenames to prevent multiple conversions of large data records.

#489


arsharma1:
@fahhem Can you merge this?

arsharma1:
@jettisonjoe

PR #493, opened 2017-01-31
Enable async prompting

Kenadia:
Main idea (from @fahhem) was to split `UserInput.prompt` into two methods that could be called separately in order to do non-blocking prompts.

Example usage:

```python
import openhtf
import time

from openhtf.plugs import user_input


@openhtf.plugs.plug(user_input=user_input.UserInput)
def async_prompt_phase(test, user_input):
  test.test_record.dut_id = 'my-dut-id'
  user_input.start_prompt('Prompt: enter some input', text_input=True)

  print 'Doing a lot of important work.'
  time.sleep(10)
  print 'Done with the important work.'

  result = user_input.wait_for_prompt()
  print 'User responded with: ' + result


if __name__ == '__main__':
  openhtf.Test(async_prompt_phase).execute()
```

### DIFF
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -58,26 +58,21 @@ Prompt = collections.namedtuple('Prompt', 'id message text_input')
 
 
 class ConsolePrompt(threading.Thread):
-  """Thread that displays a prompt to the console and waits for a response.
+  """Thread that displays a prompt to the console and waits for a response."""
 
-  Args:
-    prompt_id: The prompt manager's id associated with this prompt.
-  """
   def __init__(self, message, callback):
     super(ConsolePrompt, self).__init__()
     self.daemon = True
     self._message = message
     self._callback = callback
     self._stopped = False
+    self._answered = False
 
   def Stop(self):
-    """Mark this ConsolePrompt as stopped.
-
-    If this prompt was already stopped, do nothing.
-    """
-    if not self._stopped:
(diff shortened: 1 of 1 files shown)


openhtf/plugs/user_input.py
```diff
@@ -58,34 +58,31 @@ class PromptUnansweredError(Exception):
...
-  Args:
-    prompt_id: The prompt manager's id associated with this prompt.
-  """
   def __init__(self, message, callback):
     super(ConsolePrompt, self).__init__()
     self.daemon = True
     self._message = message
     self._callback = callback
     self._stopped = False
+    self._answered = False
```
Kenadia: Using this to distinguish between a response that came from elsewhere and a response that came asynchronously. (It was confusing to get the message "Nevermind; prompt was answered from elsewhere." after answering the prompt asynchronously in the console.)

fahhem:
There seems to be a bug that may or may not be due to this PR that makes subsequent runs with the same frontend server and without reloading in chrome not show the prompt. Once that's been fixed, I'll review this PR

Kenadia:
A gif of the bug is below, along with a still of the end state. I think that it's a race condition unrelated to this PR, that probably needs to be fixed in the frontend Angular code. I'm still looking into it.

race2

race_still

fahhem:
It looks like this bug was there before, so I'll merge this and wait for a separate PR to fix that race condition

PR #494, opened 2017-02-03
Fix reference to matches_regex in validators.py

DPJacques:
This method was renamed matches_regex, but references to _matches_regex still existed.


jettisonjoe:
Oops! Nice catch. Let's merge once tests finish and show passing.

DPJacques:
Sorry for the weird close/re-open, still getting used to GitHub's UI.  Looks like all the tests pass.

fahhem:
Thanks for the reminder :)

PR #495, opened 2017-02-07
Fix frontend prompt 405 error

Kenadia:
## Bug description and explanation

Issue: Sometimes, when submitting a prompt on the web frontend, the prompt would not submit and the frontend server would show a 405: Method not found, for the POST corresponding to the response submission.

The underlying problem seems to be that the Angular frontend can receive the state of a test before the plug handlers have been initialized on the server.

I think this occurs in the following situation:
- Angular has a link on the dashboard pointing to an old station
- Clicking that link triggers a subscribe to StationPubSub, which causes the server to publish the current test state
- The problem is that the TestWatcher might not have been started, because the TestWatcher is only started by the discovery loop in `StationStore._discover()`

## Repro steps and GIF

1. Start the test (`hello_world.py` works fine as an example).
2. Start the frontend server with a high discovery interval (`apy -m openhtf.output.web_gui --discovery_interval_s 15`), then immediately open the frontend GUI in Chrome. The station should appear on the dashboard. (You might have to wait if your timing was wrong?)
3. Finish the test in Chrome, then return to the dashboard.
4. Start the test again.
5. Go into the test in Chrome and try to respond to the prompt.

Bug: Trying to submit the prompt results in an HTTP 405 error.

The bug does not occur if the discovery interval triggers during steps 3–5. Repeat steps 3–5 if the bug does not occur the first time.

Note that the bug is easy to run into with a lower discovery interval as well, e.g. 3s (the default), which is how it was found when testing #493.

promptbug

## Proposed solution

When someone subscribes to StationPubSub, make sure that we start the TestWatchers, which will set up the plug handlers (and also ensure the Angular frontend receives updates without delay).

### DIFF
--- a/openhtf/core/station_api.py
+++ b/openhtf/core/station_api.py
@@ -362,8 +362,8 @@ class Station(object):
   which is the same dict mapping as 'tests', but does not trigger any RPCs to
   update the local state.
 
-  The 'reachable' attribute can be used as a lightweight check if the host is
-  currently responding to RPC requests (
+  The 'reachable' attribute can be used as a lightweight check of whether the
+  host is currently responding to RPC requests.
   """
 
   STATION_MAP = {}  # Map (host, port) to Station instance.
@@ -380,26 +380,25 @@ class Station(object):
     self.host = host
     self.station_api_port = station_api_port
 
-    try:
-      if self._station_info != station_info:
-        _LOG.warning(
-            'Reusing Station (%s) with new StationInfo: %s, clearing History.',
-            self._station_info.station_id, station_info)
-        self._history = history.History()
-        self.cached_tests = {}
-    except AttributeError:
+    if not hasattr(self, '_station_info'):
       logging.debug('Creating new Station (%s) at %s:%s',
                     station_info.station_id, host, station_api_port)
       # We're not reusing a shared instance, initialize things we need.
       self._history = history.History()
+      self.cached_tests = {}
       self._lock = threading.Lock()
+
+    elif self._station_info != station_info:
+      _LOG.warning(
+          'Reusing Station (%s) with new StationInfo: %s, clearing History.',
+          self._station_info.station_id, station_info)
+      self._history = history.History()
       self.cached_tests = {}
 
     self._station_info = station_info
     # Shared proxy used for synchronous calls we expect to be fast.
-    # Long-polling calls should use the 'proxy' attribute directly instead, as
-    # it creates a new ServerProxy object each time, avoiding the danger of
-    # blocking short requests with long-running ones.
+    # Long-polling calls should use make_proxy() to create a new ServerProxy
+    # object each time, to avoid blocking short requests with long-running ones.
     self._shared_proxy = proxy or self.make_proxy()
     # Lock was acquired at the beginning of __new__().
     self.STATION_LOCK.release()
--- a/openhtf/output/web_gui/__init__.py
+++ b/openhtf/output/web_gui/__init__.py
@@ -80,7 +80,6 @@ conf.declare('stations',
              default_value=[],
              description='List of manually declared stations.')
 
-
 Hostport = collections.namedtuple('Hostport', ['host', 'port'])
 
 
@@ -143,6 +142,7 @@ class TestWatcher(threading.Thread):
     self._wait_timeout_s = wait_timeout_s
     self._lock = threading.Lock()  # Used by threads.synchronized.
     self._plug_watchers = {}
+    self.handle_state_update()
     self.start()
 
   @threads.synchronized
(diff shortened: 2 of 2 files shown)


openhtf/core/station_api.py
```diff
@@ -362,8 +362,8 @@ class Station(object):
   which is the same dict mapping as 'tests', but does not trigger any RPCs to
```
Kenadia: The changes in this file are just refactoring

openhtf/output/web_gui/__init__.py
```diff
@@ -223,7 +222,7 @@ def __getitem__(self, hostport):  # pylint:disable=invalid-name
     """Provide dictionary-like access to the station store."""
     if not isinstance(hostport, Hostport):
       raise ValueError('StationStore key must be a Hostport instance.')
-    return self.stations.setdefault(hostport, None)
+    return self.stations.get(hostport)
```
Kenadia: This was unrelated, but it looked like a potential bug

openhtf/output/web_gui/__init__.py
```diff
@@ -357,21 +359,24 @@ def make_msg(cls, test_uid, remote_state):
...
+
     if hostport not in self._store.stations:
       _LOG.debug('Client tried to subscribe to unknown station. This can '
                  'happen as a result of the web gui server being restarted.')
       return
+
     try:
+      self._store.watch_tests(hostport)
       for test_uid, remote_test in self._store[hostport].tests.iteritems():
         self.send(self.make_msg(test_uid, remote_test.state))
```
Kenadia: Technically this is still a race condition—the Angular frontend may receive the test state before the TestWatchers are started—but I don't think it will be an issue in practice

openhtf/output/web_gui/__init__.py
```diff
@@ -80,7 +80,6 @@
              default_value=[],
              description='List of manually declared stations.')
 
-
 Hostport = collections.namedtuple('Hostport', ['host', 'port'])
```
fahhem: As mentioned in person, `TestWatcher.run` is full of race conditions. To be tackled later, but just an FYI to the other maintainers that a read-wait-loop has a race condition between the read and the wait. It should be, in some way, a subscribe-read-wait-loop (where wait returns immediately if a change happened after subscribe, even if it happened before the read). There's a slight chance of increased work, but 0 chance of missing an update.

This will be fixed in a similar manner to how the test binary did it: `threading.Event`

openhtf/output/web_gui/__init__.py
fahhem: Why was this a bug? It looks like the difference is that now, `self.stations` won't get a None for hostport's that weren't already in it. `setdefault()` does the same as a `get()`, but also writes the default back if it used the default

openhtf/output/web_gui/__init__.py
fahhem: if `TestWatcher`'s `__init__` called `self.handle_state_update` itself, then this wouldn't be a race condition

openhtf/output/web_gui/__init__.py
Kenadia: Right—I couldn't figure out why we'd ever want to store `None` in `self.stations`.

It has the *potential* to cause an insidious bug:

Currently the default doesn't matter because it's never actually triggered. But suppose it were, because the way StationStore is implemented suggests that it's okay to access it with missing keys. For example, in `StationPubSub.on_subscribe` change
```python
if hostport not in self._store.stations:
```
to
```python
if not self._store.stations[hostport]:
```

Then the server will crash because `DashboardPubSub.publish_discovery_update` assumes the values are never `None`. That's a good thing. Suppose instead we thought `None` values are okay. For example, in `DashboardPubSub.publish_discovery_update` add
```python
if station is None:
  continue
```

Then it will fail silently on certain edge cases because you can't discover a station if its hostport is already assigned to `None`.

openhtf/output/web_gui/__init__.py

openhtf/output/web_gui/__init__.py
Kenadia: (Acknowledged)

jettisonjoe: For a little bit of context, I believe when we originally intended to take in stride occasional missed updates as a fact of life. That's why we _send_ complete state instead of deltas, even though the deltas are computed and applied based on the complete state we have and the complete state we receive.

Kenadia:
Responded to CR

openhtf/output/web_gui/__init__.py
fahhem: Yeah, but if you don't properly listen to those complete state objects, then you're still missing updates. If a test sends an update then sits waiting for a prompt, but the frontend missed the update to show the prompt, then we're in no better shape.

PR #496, opened 2017-02-07
Make sure the test record is marked as started in tests

fahhem:
This allows phases that depend on `start_time_millis` being set have it.


jettisonjoe:
Good point. LGTM!

PR #499, opened 2017-02-09
Ignore errors in htf.util.data.total_size

fahhem:
It's only meant to be approximate anyway

### DIFF
--- a/openhtf/util/data.py
+++ b/openhtf/util/data.py
@@ -156,6 +156,13 @@ def total_size(obj):
   """Returns the approximate total memory footprint an object."""
   seen = set()
   def sizeof(current_obj):
+    try:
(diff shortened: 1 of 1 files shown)


fahhem:
This has been pulled into our local version and fixes the issues. Specifically, we were using a numpy array in a `.equals()` validator and this was attempting to iterate through it recursively. Instead of catching numpy arrays directly (without making it a dependency of openhtf), I decided to take this route. This is an approximate function anyway, and a 0-d array is a fake thing numpy creates anyway so it's not even part of the memory footprint here. In other cases, tracebacks are purely worse than any number, and here we return 4 or 8 (the size of a pointer) since we assume whatever we followed to this object had at least that.

PR #501, opened 2017-02-10
Ensure PhaseExecutor and PlugManager are torn down on exit

fahhem:
Since we use the PhaseExecutor and PlugManager to execute the start trigger now, they need to be notified of cancel events properly

### DIFF
--- a/openhtf/core/__init__.py
+++ b/openhtf/core/__init__.py
@@ -114,6 +114,10 @@ class TestExecutor(threads.KillableThread):
       # Any access to self._exit_stack must be done while holding this lock.
       with self._lock:
         self._exit_stack = exit_stack
+        # We need to exit the PhaseExecutor before we tear down the plugs
+        # that a phase may be executing, so we add them in reverse order.
+        exit_stack.callback(self.test_state.plug_manager.tear_down_plugs)
+        exit_stack.callback(executor.stop)
 
       # Have the phase executor run the start trigger phase. Do partial plug
       # initialization for just the plugs needed by the start trigger phase.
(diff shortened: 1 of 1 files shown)


fahhem:
ping @grybmadsci

openhtf/core/__init__.py
```diff
@@ -114,6 +114,10 @@ def _thread_proc(self):
       # Any access to self._exit_stack must be done while holding this lock.
       with self._lock:
         self._exit_stack = exit_stack
+        # We need to exit the PhaseExecutor before we tear down the plugs
+        # that a phase may be executing, so we add them in reverse order.
+        exit_stack.callback(self.test_state.plug_manager.tear_down_plugs)
+        exit_stack.callback(executor.stop)
```
grybmadsci: hm, by doing this here, the `PlugManager` may have `tear_down_plugs()` called before `initialize_plugs()`.  are you trying to address that `executor.stop()` is never called if the start trigger is broken?  can you not add `executor.stop()` to the block below (line 137 in the old version)?

fahhem: Well, if `initialize_plugs()` is never called, `tear_down_plugs()` does nothing, right? The problem I'm getting is if we're in `test_start` and attempt to exit, nothing here gets torn down.

grybmadsci: ah, right, there should be an `exit_stack.callback(self.test_state.plug_manager.tear_down_plugs` between the `initialize_plugs()` and `executor.execute_start_trigger()`, but then it gets a bit sloppy to handle all the branching.  moving it up here is all right as long as you've double checked that both `executor.stop()` and `tear_down_plugs()` can be safely called if the test isn't started (ie, immediately upon exiting this `with self._lock:` block).

fahhem: `PhaseExecutor.stop()` will only kill its current phase, if any. If it's called right after this block, there will be no phases so nothing happens. `PlugManager.tear_down_plugs()` won't do anything when it hasn't initialized any plugs, and will only tear down those that it has initialized. There is no lock though, so if you hit at the right time, there's a possible race condition that we'll need to address some other way. I propose that we move `self.kill()` above `self._exit_stack.close()` in `stop()` above (in another PR).

PR #502, opened 2017-02-10
Don't wait for the multicast listener to stop when stopping the StationApi server

fahhem:


fahhem:
The `join()` is there in case we use it elsewhere and actually want to block

jettisonjoe:
Yeah, makes sense.

PR #506, opened 2017-02-11
Also log plug init exceptions to the test record.

### LINKED ISSUE #492: Include plug init tracebacks in test record
When tester fails for equipment due to equipment not being present I'm not seeing the stack trace available in the log:

Phase trigger_phase at 08:43:14.548 for 1 minute 3.299 seconds
08:44:18.130	Finishing test with outcome ABORTED.

And finishes with status ERROR.

jettisonjoe:
Fixes #492.

Also:
  * Fixes a bug in station_api.py that was causing unit tests to fail.
  * Replaces the object() logger in plugs_test.py with a MagicMock.

### DIFF
--- a/openhtf/plugs/__init__.py
+++ b/openhtf/plugs/__init__.py
@@ -482,7 +482,11 @@ class PlugManager(object):
           # Now the instance has its own copy of the test logger.
           plug_instance.logger = self._logger
       except Exception:  # pylint: disable=broad-except
-        _LOG.error('Exception instantiating plug type %s', plug_type)
+        _LOG.error('Exception instantiating plug type %s', plug_type,
+                   exc_info=True)
+        # Also log to the test record.
+        self._logger.error('Exception instantiating plug type %s', plug_type,
+                           exc_info=True)
         self.tear_down_plugs()
         raise
       self.update_plug(plug_type, plug_instance)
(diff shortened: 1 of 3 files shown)


openhtf/plugs/__init__.py
```diff
@@ -482,7 +482,11 @@ def initialize_plugs(self, plug_types=None):
           # Now the instance has its own copy of the test logger.
           plug_instance.logger = self._logger
       except Exception:  # pylint: disable=broad-except
-        _LOG.error('Exception instantiating plug type %s', plug_type)
+        _LOG.error('Exception instantiating plug type %s', plug_type,
```
fahhem: _LOG.exception?

openhtf/plugs/__init__.py
```diff
@@ -482,7 +482,11 @@ def initialize_plugs(self, plug_types=None):
...
+                   exc_info=True)
+        # Also log to the test record.
+        self._logger.error('Exception instantiating plug type %s', plug_type,
+                           exc_info=True)
```
fahhem: What if we only logged to this logger? Doesn't it show up in stdout too?

wallacbe:
I tested this branch out, I am able to see the stack trace in the test run record.

openhtf/plugs/__init__.py
jettisonjoe: Confirmed, this logger does have console logging (dependent on command line flags), so logging once is better.

openhtf/plugs/__init__.py
jettisonjoe: Will do. I didn't realize this existed, but will start using it now that I know!

jettisonjoe:

jettisonjoe:
NOTE: Also fixes #507.

wallacbe:
LGTM

PR #512, opened 2017-02-14
Make PhaseState._asdict return a non-zero start_time_millis

fahhem:
This makes the frontend show the right start time for the phase

### DIFF
--- a/openhtf/core/test_record.py
+++ b/openhtf/core/test_record.py
@@ -51,7 +51,7 @@ class Attachment(collections.namedtuple('Attachment', 'data mimetype')):
 class TestRecord(  # pylint: disable=no-init
     mutablerecords.Record(
         'TestRecord', ['dut_id', 'station_id'],
-        {'start_time_millis': None, 'end_time_millis': None,
+        {'start_time_millis': int, 'end_time_millis': None,
          'outcome': None, 'outcome_details': list,
          'code_info': None,
          'metadata': dict,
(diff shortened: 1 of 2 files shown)


openhtf/core/test_record.py
```diff
@@ -72,7 +72,7 @@ class PhaseRecord(  # pylint: disable=no-init
     mutablerecords.Record(
         'PhaseRecord', ['name', 'codeinfo'],
         {'measurements': None,
-         'start_time_millis': None, 'end_time_millis': None,
+         'start_time_millis': int, 'end_time_millis': None,
```
jettisonjoe: What's the motivation to change this from `None` besides being able to cast it in `_asdict()` without a check? It would be really weird, but technically _possible_ for someone to use timestamps that weren't epoch-based, and thus for zero to be a valid timestamp. Effectively defaulting this to zero would create ambiguity in that (admittedly freakish) case.

Aside from that, is the result of calling `int()` part of the definition of Python, or is it implementation specific?

fahhem: https://docs.python.org/2/library/functions.html#int (it says `int(x=0)`)

Are you saying a platform we may support in the future may use non-epoch times? Or that someone may attempt to use `TestRecord` without our framework and throw non-epoch times in there? Because we call `time.time()` for all our timestamps.

jettisonjoe: I'm saying that OpenHTF is historically un-opinionated about where and how you use it (and its parts).

fahhem:
Yeah, I see why we say OpenHTF is unopinionated about certain things, but not about "how time works". We do call these fields "start_time_**millis**" after all.

grybmadsci:
Wait, why can't the frontend detect the None case?  Using 0 as a sentinel value in Python is wrong, period.  If it's unset, it should be None.  If the test actually started at time 0, then it should be 0.

fahhem:
Right now, the frontend always gets 0, which is wrong, so I fixed that, and I don't have time to wax philosophical about what sentinel value to use. Feel free to open an issue and assign it to me for when I get this all working

PR #513, opened 2017-02-15
Alter the RemotePlug management to accept the latest connection instead of only caching them

fahhem:
This is because the remote PlugManager may go away and come back between test state updates, but everything stay the same, such as when the test is being looped within the same process.

### DIFF
--- a/openhtf/output/web_gui/__init__.py
+++ b/openhtf/output/web_gui/__init__.py
@@ -394,11 +394,33 @@ class WebGuiServer(tornado.web.Application):
     """Main handler for OpenHTF frontend app.
 
     Serves the index page; the main entry point for the client app."""
+
     def initialize(self, port):  # pylint: disable=arguments-differ
       self.port = port
 
     def get(self):
       self.render('index.html', host=socket.gethostname(), port=self.port)
+
+  class PlugsHandler(tornado.web.RequestHandler):
+    """Handler for all plugs from the OpenHTF frontend app."""
+
+    def initialize(self, remote_plugs):
+      self._plugs = remote_plugs
+
+    def post(self):
+      host, port_str, test_uid, plug_name = self.request.path.split('/')[2:]
+      plug = self._plugs.get((host, int(port_str), test_uid, plug_name))
+      if plug is None:
+        self.set_status(500)
+        self.write('Plug "%s" not found in "%s"' % (
+            self.request.path, self._plugs.keys()))
+      try:
+        response = plug.respond(self.request.body)
+      except Exception as e:  # pylint: disable=broad-except
+        self.set_status(500)
+        self.write('Remote plug error: %s' % repr(e))
+      else:
+        self.write(response)
 
   def __init__(self, discovery_interval_s, disable_discovery, http_port,
                dev_mode=False):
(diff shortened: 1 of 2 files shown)


fahhem:
This fixes the bug where, when looping a test, the frontend won't replace the plug handler (the host, port, test_uid, and plug_name are all the same), even though the opposing `PlugManager`'s server went away and came back.

openhtf/output/web_gui/__init__.py
```diff
@@ -441,23 +455,19 @@ def handle_test_state_update(self, hostport, test_uid, state):
...
-          in plugs.RemotePlug.discover(hostport.host, plugs_port)
-          if not self.has_handler_for_url(make_plug_url(plug_name))
-      ]
-      self.add_handlers('.*$', handlers)
+    if not plugs_port:
+      return
+
+    # Update the plugs we're interacting with. The PlugManager we're interacting
+    # with may have gone and come back since the last update, so we have to
+    # open a new connection each time to avoid getting an update
```
Kenadia: I didn't understand this comment, e.g. it's unclear what the problem is in “to avoid getting an update”

(I'd give a suggestion for rewording but I'm still working on understanding the bug, brb)

Kenadia: Update the plugs we're interacting with. On each state update, we open new connections to the XML-RPC server via RemotePlug.discover because the PlugManager we're interacting with may have gone and come back since the last update.

openhtf/output/web_gui/__init__.py
```diff
@@ -394,19 +394,42 @@ class MainHandler(tornado.web.RequestHandler):
...
+    def initialize(self, remote_plugs):
+      self._plugs = remote_plugs
+
+    def post(self):
+      host, port_str, test_uid, plug_name = self.request.path.split('/')[2:]
+      plug = self._plugs.get((host, int(port_str), test_uid, plug_name))
+      if plug is None:
+        self.set_status(500)
+        self.write('Plug "%s" not found in "%s"' % (
+            self.request.path, self._plugs.keys()))
```
fahhem: Done

Kenadia:
lgtm!

openhtf/output/web_gui/__init__.py
```diff
@@ -391,19 +391,43 @@ class MainHandler(tornado.web.RequestHandler):
...
       self.render('index.html', host=socket.gethostname(), port=self.port)
 
+  class PlugsHandler(tornado.web.RequestHandler):
+    """Handler for all plugs from the OpenHTF frontend app."""
+
+    def initialize(self, remote_plugs):
+      self._plugs = remote_plugs
+
+    def post(self):
+      host, port_str, test_uid, plug_name = self.request.path.split('/')[2:]
```
Kenadia: You can get path params by using capturing groups in the handler's regex:

```python
def post(self, host, port_str, test_uid, plug_name):
  ...

def __init__(...):
  ...
  handler_routes = [
    ...
    (r'/plugs/([\d\.]+)/(\d+)/(.+)/(.+)', self.PlugsHandler,
     {'remote_plugs': self.remote_plugs}),
```

fahhem: Done

PR #515, opened 2017-02-15
Frontend-aware plugs must enable remote access

fahhem:
Otherwise, what's the point of being frontend-aware? Just exposing your data, but not your API?


Kenadia:
We can remove `enable_remote = True` from `UserInput` then

PR #518, opened 2017-02-23
Introduce support for custom frontends

Kenadia:
custom_frontends

### DIFF
--- a/openhtf/output/web_gui/__init__.py
+++ b/openhtf/output/web_gui/__init__.py
@@ -49,7 +49,6 @@ import collections
 import functools
 import json
 import logging
-import os
 import signal
 import socket
 import sys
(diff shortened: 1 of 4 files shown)


openhtf/output/web_gui/__init__.py
```diff
@@ -408,7 +409,14 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
...
 
-    path = BUILD_PATH if os.path.exists(BUILD_PATH) else PREBUILT_PATH
+    if 'custom_frontend' in conf:
+      path = conf.custom_frontend
+    else:
+      path = BUILD_PATH if os.path.exists(BUILD_PATH) else PREBUILT_PATH
+    if not os.path.exists(path):
+      raise ValueError('Frontend path `%s` does not exist.' % path)
+    _LOG.info('Running frontend from path `%s`')
+
```
fahhem: it feels like this logic should actually live in __main__ and the path to use should be passed into this server. Then it could even be an argument instead of a config variable, which makes more sense to me for some reason

Kenadia: Sounds good

Kenadia:
Responded to CR.

New invocation is `python -m openhtf.output.web_gui --custom_frontend=examples/example_frontend`

PR #520, opened 2017-02-24
Make static station information actually work

fahhem:
Without this, if you do --disable_discovery, you're never going to get anything on the dashboard because we never actually attempt to watch or listen to the stations configured via the config (in `conf.stations`).

### DIFF
--- a/openhtf/output/web_gui/__init__.py
+++ b/openhtf/output/web_gui/__init__.py
@@ -223,13 +223,9 @@ class StationStore(threading.Thread):
 
   def _discover(self):
     """Discover stations through the station API."""
-    hostports = []
     for station in station_api.Station.discover():
       hostport = Hostport(station.host, station.station_api_port)
-      self.stations.setdefault(hostport, station)
-      hostports.append(hostport)
-
-    self._handle_stations(hostports)
+      self.stations[hostport] = station
 
   def _handle_stations(self, hostports):
     for hostport in hostports:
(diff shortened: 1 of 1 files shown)


openhtf/output/web_gui/__init__.py
```diff
@@ -223,26 +223,33 @@ def __getitem__(self, hostport):  # pylint:disable=invalid-name
...
+    self._handle_stations(self.stations.iterkeys())
     while not self._stop_event.is_set():
-      self._discover()
-      self._stop_event.wait(self._discovery_interval_s)
+      if self._disable_discovery:
+        _LOG.debug('Station discovery is disabled; using static station info')
+        self._handle_stations(self.stations.iterkeys())
+      else:
+        self._discover()
+        self._stop_event.wait(self._discovery_interval_s)
```
Kenadia: It seems like we should wait on `_discovery_interval_s` regardless of whether `disable_discovery` is `True`.

openhtf/output/web_gui/__init__.py
```diff
@@ -223,26 +223,33 @@ def __getitem__(self, hostport):  # pylint:disable=invalid-name
...
-    if self._disable_discovery:
-      _LOG.debug("Station discovery is disabled; StationStore won't update.")
-      return
+    self._handle_stations(self.stations.iterkeys())
     while not self._stop_event.is_set():
-      self._discover()
-      self._stop_event.wait(self._discovery_interval_s)
+      if self._disable_discovery:
+        _LOG.debug('Station discovery is disabled; using static station info')
+        self._handle_stations(self.stations.iterkeys())
```
Kenadia: Seems like it's reasonable for `_disable_discovery` to be `True` and `conf.stations` to be set as well, in which case, we want to handle the stations in `self.stations` even when `_disable_discovery` is False.

If that sounds right, then let's remove the call to `_handle_stations` from `_discover`. Then in this loop, we can call `_discover`, followed by a call to `self._handle_stations(self.stations.iterkeys())` to get all the stations.

openhtf/output/web_gui/__init__.py
fahhem: Good point!

openhtf/output/web_gui/__init__.py
fahhem: Yeah, that cleaned up a bunch, thanks!

Issue #521, opened 2017-02-24
StationApi tracks tests based on (hostport, test.uid) combo, but that's not unique enough


fahhem:
If a test is run in multiple threads in the same binary, or even multiple times in a row, this causes problems as caches/dicts get filled with the first test seen with that uid. Later, when another test is seen with the same uid, it's assumed the `TestWatcher` will watch it, but it doesn't necessarily and the frontend can freeze.

jettisonjoe:
The station API is no more.

PR #523, opened 2017-02-25
Change the test UID into an execution UID

fahhem:
This allows the frontend to track executions between loops and not
poison its cache nor history's cache.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -28,6 +28,7 @@ import socket
 import sys
 import textwrap
 import threading
+import uuid
 import weakref
 from types import LambdaType
 
@@ -107,7 +108,7 @@ class Test(object):
     self._lock = threading.Lock()
     self._executor = None
     self._test_desc = TestDescriptor(
-        self.uid, phases, test_record.CodeInfo.uncaptured(), metadata)
+        phases, test_record.CodeInfo.uncaptured(), metadata)
 
     if conf.capture_source:
       # First, we copy the phases with the real CodeInfo for them.
@@ -132,7 +133,7 @@ class Test(object):
     else:
       self.configure()
 
-    self.TEST_INSTANCES[self.uid] = self
+    self._increment_next_uid()
     # This is a noop if the server is already running, otherwise start it now
     # that we have at least one Test instance.
     station_api.start_server()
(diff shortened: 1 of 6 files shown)


fahhem:
Fixes #521

openhtf/__init__.py
```diff
@@ -151,15 +152,18 @@ def from_uid(cls, test_uid):
...
+    """Increments the next execution's UID.
 
-    Note that this identifier must be unique across Python invokations, so the
-    station_api UID is used as a prefix to guarantee that.  Test UID's need not
-    be unique across different hosts.
+    This identifier must be unique across invocations of execute() to
+    differentiate between different runs of the same test.
     """
-    return '%s:%s' % (station_api.STATION_API.UID, id(self))
+    self._uid = '%s:%s' % (station_api.STATION_API.UID, uuid.uuid4().hex)
```
Kenadia: I found this logic with `get_current_or_next_uid` and `_increment_next_uid` hard to follow, and I think it's unnecessarily complex. We're saying that UIDs now belong to executions of a test and not the test itself, but it seems like the code is only halfway there.

It appears the only reason we need `self._uid` and the idea of a “next execution” is because of the call to `test.get_current_or_next_uid()` in `list_tests` of station_api.py. Could we instead decide that a test doesn't need to be included in the station API's list of tests if the test isn't executing yet? If so then `Test` could look something like this:

Remove `self._uid`, `get_current_or_next_uid`, and `_increment_next_uid`

```
@property
def uid(self):
  """Return UID of current execution, or None if the test is not executing."""
  if self._executor:
    return self._executor.uid

def make_uid(self):
  return '%s:%s' % (station_api.STATION_API.UID, uuid.uuid4().hex)

...

def execute(...):
  ...
  self._executor = core.TestExecutor(self._test_desc, self.make_uid(), ...)
```

grybmadsci:
Notably, doing this removes the ability to differentiate between sequential invocations of the same `Test` instance versus newly created `Test` instances, which is likely to matter in some multi-up station configurations.

Instead of "fixing" the old UID, which served a different purpose, if you need the ability to differentiate sequential runs of the same `Test`, why not just add that?

openhtf/__init__.py
fahhem: I needed the next uid available for other reasons, but I think I can get by with it only available when executing. I'll just put my code as an early phase or part of test_start

fahhem:
The old UID is broken for the more common case of looping a test, versus the less common case of looping multiple tests in the same process (different processes have different station api prefixes). I changed it to be made of three parts to handle that situation, but I don't consider this a different purpose, but a bugfix. The three parts:

* StationApi's UID, which is a process-specific (pid and start time)
* TestDescriptor's UUID, which is test descriptor specific in the same process
* Test execution's UUID, which is specific to the execution of a test

All three together create a sufficiently unique string for the frontend to watch test results correctly, but also allow it to potentially show the different looping tracks as necessary.

PR #524, opened 2017-02-25
Do not pickle the TestRecord in the station API

fahhem:
The TestRecord may include things that the tester binary knows about,
such as numpy arrays (as measurement data, for instance), but that the
frontend doesn't (it may be built in a different virtualenv). In these
scenarios, the frontend cannot unpickle a TestRecord, so we first
convert to basic data types in order to avoid such problems.

### DIFF
--- a/openhtf/core/station_api.py
+++ b/openhtf/core/station_api.py
@@ -85,9 +85,9 @@ from openhtf.util import threads
 from openhtf.util import timeouts
 from openhtf.util import xmlrpcutil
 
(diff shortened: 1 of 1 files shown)


openhtf/core/station_api.py
```diff
@@ -761,7 +758,7 @@ def get_history_after(self, test_uid, start_time_millis):
     """Get a list of pickled TestRecords for test_uid from the History."""
```
Kenadia: Update this docstring

fahhem:
Done

PR #466, opened 2016-11-25
Make plug __init__ have a test record logger

fahhem:
Fixes #403

PR #525, opened 2017-02-25
Fastboot, ADB Plugs: Forward attributes to devices

arsharma1:
Rather than construct a fake BasePlug for the FastbootPlug and AdbPlug classes, construct a device member and forward all attributes there.  This allows the plugs to be used with the new logger checks enforced by the PlugManager.

### DIFF
--- a/openhtf/plugs/usb/__init__.py
+++ b/openhtf/plugs/usb/__init__.py
@@ -86,32 +86,48 @@ def _open_usb_handle(**kwargs):
 class FastbootPlug(plugs.BasePlug):
   """Plug that provides fastboot."""
 
-  def __new__(cls):
-    device = fastboot_device.FastbootDevice.connect(
+  def __init__(self):
+    self._device = fastboot_device.FastbootDevice.connect(
         _open_usb_handle(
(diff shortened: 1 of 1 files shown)


arsharma1:
@fahhem

openhtf/plugs/usb/__init__.py
```diff
@@ -86,32 +86,48 @@ def _open_usb_handle(**kwargs):
...
-    return device
+
+  def tearDown(self):
+    self._device.close()
+
+  def __getattr__(self, attr):
+    """Forward other attributes to the device."""
+    # Use the class version of logger.
+    if attr == 'logger':
+      return self.__getattribute__(attr)
```
jettisonjoe: I'm curious about the decision to use `__getattribute__` here rather than `self.logger` or simply letting the attribute lookup go through to BasePlug. Accesses only end up in `__getattr__` if those other strategies fail, right?

arsharma1: My original implementation for a similar plug required this code to pass the PlugManager check pre #466.

This is no longer necessary, so it will now forward any unknown attribute to the device class.

PR #527, opened 2017-02-28
Fixing test teardown issues

kdsudac:
* Increasing default test timeout to 30 seconds
* Fixing bug where executor checked for 'timeout_s' attribute directly on teardown_function instead of on teardown_function.options


fahhem:
I'll merge once you fix the tests (minor stuff it looks like)

PR #528, opened 2017-02-28
Allow limiting multicast discovery to localhost

fahhem:
This just does the low-level work. I'll wire this out to the `ApiServer` in a later PR

### DIFF
--- a/openhtf/util/multicast.py
+++ b/openhtf/util/multicast.py
@@ -128,6 +128,7 @@ def send(query,
          address=DEFAULT_ADDRESS,
          port=DEFAULT_PORT,
          ttl=DEFAULT_TTL,
+         local_only=False,
          timeout_s=2):
   """Sends a query to the given multicast socket and returns responses.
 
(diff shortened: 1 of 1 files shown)


openhtf/util/multicast.py
```diff
@@ -144,6 +145,16 @@ def send(query,
...
   sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
   sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL, ttl)
+  if local_only:
+    sock.setsockopt(
+        socket.IPPROTO_IP,
+        socket.IP_MULTICAST_IF,
+        # IP_MULTICAST_IF is the 8-byte group address followed by the IP
+        # assigned to the interface on which to listen.
+        struct.pack(
+            '!4sL',
```
Kenadia: Where does this address come from? I see that it's used above. Is it something that would be worth pulling out as a constant and documenting?

fahhem: 127.0.0.1 is the universal address of localhost. Or are you asking about '!4sL', which is defined by the struct module's documentation

Kenadia: Thanks—I should have checked the struct documentation!

PR #529, opened 2017-03-01
Ensure take_at_least_n_seconds never sleeps a negative amount

fahhem:
`time.sleep()` can raise an `IOError` when given a negative number

### DIFF
--- a/openhtf/util/timeouts.py
+++ b/openhtf/util/timeouts.py
@@ -102,7 +102,10 @@ class PolledTimeout(object):
   def remaining(self):
     if self.timeout_s is None:
       return None
-    return self.timeout_s - self.seconds
+    # We max() against 0 to ensure we don't return a (slightly) negative number.
+    # This reduces races between .has_expired() calls and sleeping/waiting
+    # .remaining seconds.
+    return max(0, self.timeout_s - self.seconds)
 
   @property
   def remaining_ms(self):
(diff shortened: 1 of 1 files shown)


openhtf/util/timeouts.py
```diff
@@ -369,7 +369,9 @@ def take_at_least_n_seconds(time_s):
   timeout = PolledTimeout(time_s)
   yield
   while not timeout.has_expired():
-    time.sleep(timeout.remaining)
+    # We max() against 0 to ensure we don't pass a (slightly) negative number
+    # in the race between .has_expired() being True and .remaining going negative.
+    time.sleep(max(0, timeout.remaining))
```
Kenadia: I think we have this same problem (a race condition) in `execute_after_delay` in this same file, where we also do `time.sleep(timeout.remaining)`.

Could we change PolledTimeout to guarantee returning a non-negative number? Looking at the other uses of `PolledTimeout.remaining` and `.remaining_ms` in the OpenHTF codebase, it seems that it's always used either as a timeout or as a waiting period, neither of which makes sense as a negative number. I don't think it's unintuitive for the remaining time to stop at zero.

It seems odd that we would need a two-line comment for what looks like a typical use-case of PolledTimeout.

---

Somewhat related: can we change `PolledTimeout.has_expired` to return `self.seconds >= self.timeout_s` instead of `self.seconds > self.timeout_s`?

fahhem: Done

PR #530, opened 2017-03-03
Adding name to phase_executor

kdsudac:
Trying to debug an issue with zombie threads.  Useful for debugging to have thread names.


fahhem:
(still approved) why set self.name now instead of passing it into super?

fahhem:
Ah, so you have to pass it to `__init__` :-)

kdsudac:
It seemed cleaner.  It failed on test though since PhaseExecutorActually overwrites the name attribute via @property decorator.  

I switched it back, I'm not sure if python will read the name passed through init or the @property.

PR #532, opened 2017-03-09
Proposal for dealing with replaceable components

fahhem:


fahhem:
Tests are fixed now

fahhem:
ping

jettisonjoe:
The main thing I'm having a hard time with is the naming. I find the `DependencyOptions` nomenclature a bit obscure. Something like, I dunno, `ComponentClasses` or something that evokes how the stuff you configure thus is actually going to be used?

As far as implementation goes, I'm wary of the interfaces only being specified in comment form rather than providing something like abstract base classes to enforce the interfaces.

fahhem:
The interfaces aren't great, so I'd rather spend time on cleaning those up first, then creating base classes and such.

I don't care about the name, here's `InjectedDependencies`.

openhtf/__init__.py
```diff
@@ -282,10 +283,33 @@ def trigger_phase(test):
...
+      wait_for_plug_update(name, last state, timeout_s) -> new state,
+      get_frontend_aware_plug_names() -> [names]
+  """
+
+
 class TestOptions(mutablerecords.Record('TestOptions', [], {
     'name': 'OpenHTF Test',
     'output_callbacks': list,
     'teardown_function': None,
+    'inject_dependencies': InjectedDependencies,
```
jettisonjoe: Why shorten 'injected' to 'inject' in the name?

jettisonjoe:
This PR has been approved since March, so I gather it's getting stale at this point. I'm going to close it, but feel free to re-open if necessary.

PR #534, opened 2017-03-14
Expose phases and history in the web GUI server

Kenadia:
This is a prerequisite for displaying pending phases and test history (execution history of a given Test instance) in the frontend.

In order to reliably identify pending phases, we need to be able to identify the phase descriptor that a given phase record came from. Since phases can repeat, and since phase names are not unique, this PR adds the concept of a phase descriptor ID for that purpose. The ID is the Python ID for the phase descriptor object, and it is added to `RemotePhaseDescriptor` and `PhaseRecord`.


openhtf/output/web_gui/__init__.py
```diff
@@ -387,13 +387,53 @@ def on_unsubscribe(self):
     self.subscriber_to_hostport_map.pop(self)
 
 
+class TestHandler(tornado.web.RequestHandler):
```
Kenadia: I thought it might be unnecessary to nest this within `WebGuiServer`. But I know we already nested `MainHandler` and `PlugsHandler`, so let me know what your preference is.

fahhem: It's fine. It looks like you have merged two handlers into this one, a `TestPhasesHandler` and a `TestHistoryHandler`. Can you separate them, either by creating a base class or a function to handle the common stuff in `get()`

openhtf/core/station_api.py
```diff
@@ -576,10 +576,17 @@ def get_phase_descriptors(self, test_uid):
...
+      UnrecognizedTestUidError: The test_uid is not recognized.
     """
-    phases = openhtf.Test.from_uid(test_uid).descriptor.phases
-    return [data.convert_to_base_types(phase) for phase in phases]
+    phases_as_dicts = []
+    for phase in openhtf.Test.from_uid(test_uid).descriptor.phases:
+      phase_as_dict = data.convert_to_base_types(phase)
+      phase_as_dict['id'] = id(phase)
+      phases_as_dicts.append(phase_as_dict)
+    return phases_as_dicts
```
Kenadia: Done

Kenadia:
Passed tests

Kenadia:
Passed tests, PTAL

openhtf/output/web_gui/__init__.py
```diff
@@ -387,13 +387,57 @@ def on_unsubscribe(self):
...
+      test = station.tests[test_uid]
+    except KeyError:
+      self.write('Unknown test ID %s' % test_uid)
+      self.set_status(404)
+      return
+
+    self.write_result(test)
+
+  def write_result(self, test):
+    raise NotImplmenetedError
```
fahhem: typo

openhtf/output/web_gui/__init__.py
```diff
@@ -438,6 +482,10 @@ def __init__(self, discovery_interval_s, disable_discovery, http_port,
         (r'/', self.MainHandler, {'port': http_port}),
         (r'/station/(?:\d{1,3}\.){3}\d{1,3}/(?:\d{1,5})/?',
          self.MainHandler, {'port': http_port}),
+        (r'/phases/([\d\.]+)/(\d+)/(.*)/?', PhasesHandler,
```
fahhem: shouldn't this be `/station/(host)/(port)/(test_uid)/phases`? plugs is similarly wrong to me (but we can fix that later)

PR #536, opened 2017-03-15
Race condition when cancelling test (examples only)

kdsudac:
This adds some examples to illustrate a race condition when exiting a test via ctrl-c.  Particularly susceptible when phases use threads.

Follow on branch async_fix includes fixes.  This PR does not need to be merged, only for illustrative purposes.

Steps to run example:
1) `python examples/sig_int.py`
2) Ctrl-C out of test when prompted
3) Observe logging statements to understand logic branching.
4) Repeat multiple times

Depending on where the phase is in it's loop, different branches of phase_executor.execute_phases logic can be stepped into.

Root cause:  threads.KillableThread.kill() does not cause the current_phase_thread to be killed immediately.  Delays can be up to  on order of 0.5-10 seconds in this example.  As a result phase_executor._execute_one_phase may or may not return a valid outcome.

Side effects:  
* test-level teardown function may or may not be called before the plugs are torn down.  
* in some cases test-level teardown function is called multiple times
* the test-level teardown and the plug teardown appear to be run in different threads so can actually race against each other.


fahhem:
Some context?

kdsudac:
Note: The example here actually blocks in phase_executor.stop() to show the timing.  Without this, things might be worse (perhaps better depending on your perspective) because phase_executor.stop() won't block after killing the thread.  This means the ExitStack close() will continue on to call plug_manager.tear_down_plugs

fahhem:
Fixed by #541

PR #537, opened 2017-03-15
Fix race condition when cancelling test

kdsudac:
Fixes race condition when ctrl-c to exit a test.

Root cause: threads.KillableThread.kill() does not cause the current_phase_thread to be killed immediately. Delays can be up to on order of 0.5-10 seconds in this example. As a result phase_executor._execute_one_phase may or may not return a valid outcome.

Side effects:

test-level teardown function may or may not be called before the plugs are torn down.
in some cases test-level teardown function is called multiple times
the test-level teardown and the plug teardown appear to be run in different threads so can actually race against each other.


openhtf/core/phase_executor.py
```diff
@@ -251,6 +261,13 @@ def stop(self):
...
     current_phase_thread = self._current_phase_thread
     if current_phase_thread:
-      current_phase_thread.kill()
+      kill_start = time.time()
+      current_phase_thread.stop()
+      while current_phase_thread.is_alive():
+        time.sleep(.1)
+      self.test_state.logger.debug('Took %.3f seconds to kill %s',
+          time.time() - kill_start,
+          current_phase_thread.name)
```
fahhem: The approach I would have taken is to move the whole teardown function stuff up a level to the `TestExecutor`. It's already in charge of `test_start`, and has a perfect singular place to ensure the teardown gets called (the finally clause in `_execute_test_phases`, or a new try-finally in `_thread_proc` on L144) versus the 3 places it's mentioned here.

It seems to decrease complexity to me, what do you think?

kdsudac: I definitely agree a singular place is better than the 3 places.  :)

Short term:  We should fix this race condition.   We are relying upon the current_phase_thread to actually be killed in a timely fashion and return the PhaseOutcome(ThreadTerminationError) inside of phase_executor.execute_phases for it to correctly unwind things.  Otherwise things fall apart in unpredictable fashion.

Longer term: moving it to test_executor would probably get my vote.  There is some logic (that I haven't quite groked) around the finalize that we'd have to be careful about.

fahhem:
Fixed by #541

PR #541, opened 2017-03-17
Cleanup execution separation between Test and PhaseExecutor

fahhem:
As a result, teardown_function is executed in TestExecutor in a finally
clause synchronously, rather than asynchronously in PhaseExecutor. This
should obviate the need for #537

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/541)


kdsudac:
I think this is cleaner, and I like it.  But I don't think it fixes the race condition.

The exit stack still executes in order:
phase_executor.stop()
plug_manager.tear_down_plugs()

Since phase_executor.stop() kills the thread but does not block so it will start race for the try/except/finally you added to call test teardown_func.

Other race is against plug_manager.tear_down_plugs().

fahhem:
Okay, I moved into the exit stack, which should absolutely fix the race condition

kdsudac:
Better, this should definitely make it so the test teardown function gets run first.

Can still have some odd side-effects in that the phase will still be running while the test teardown is run in parallel in separate thread.

kdsudac:
I tested this and it's worse.  test_state.running_phase_context enforces only one phase running at a time (makes sense).  So you'll need to kill the running phase first.  

On top of that, calling test teardown function in a new thread adding a new thread to the race.

fahhem:
Okay, I added tests for cancelling the test in the middle of test-start and a phase and I fixed `TestState` blocking the teardown function.

There's now a timeout (default 30s) for how long to wait for the cancelled phase to finish, since we can't assume that it ever will. `KillableThread` is both async and imperfect, as it doesn't guarantee the Python interpreter will ever see the 'Kill' exception. As a result, you have to give up waiting for it to finish or die

kdsudac:
Reviewed 1 of 6 files at r3, 2 of 3 files at r4.
Review status: 3 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[openhtf/core/phase_executor.py, line 200 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-KfrxrmAxe0Wx3XnCWTM:-KfrxrmAxe0Wx3XnCWTN:b509t37) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/phase_executor.py#L200)):*
> ```Python
>       repeat_count += 1
>       if repeat_limit is None or repeat_count < repeat_limit:
>         continue
> ```

The logic in this while True block can be greatly simplified:
outcome = self._execute_phase_once(phase)
if outcome is openhtf.PhaseResult.REPEAT and repeat_count < repeat_limit:
  continue
return outcome
  

---

*[openhtf/core/phase_executor.py, line 243 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-KfrzCqvMGxCrvq6VNVU:-KfrzCqwkRZQiL0nGImW:b-h8ysvu) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/phase_executor.py#L243)):*
> ```Python
>       time.sleep(0.1)
>     _LOG.debug('Cancelled phase did%s exit',
>                "n't" if phase_thread.is_alive() else '')
> ```

Lol, took me a bit to figure out what this was doing.

---

*[openhtf/core/test_executor.py, line 42 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfs72_DJjUs7YhVoxlY:-Kfs72_EUK7AuEiGeRiU:bf49av1) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/test_executor.py#L42)):*
> ```Python
> conf.declare('cancel_timeout_s', default_value=5,
>              description='Timeout (in seconds) when the test has been cancelled'
>              'to wait for the running phase to exit.')
> ```

Can we increase default?  Maybe to 10 seconds?

If timeout it too low, you end up with race condition (which now I think will just be the original phase racing against test teardown).

If timeout is too high, you just end up wasting a little time.

---

*[openhtf/core/test_executor.py, line 83 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfs51CeXsrdz04PXXOo:-Kfs51CeXsrdz04PXXOp:b-47bs1i) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/test_executor.py#L83)):*
> ```Python
>     """Stop this test."""
>     _LOG.info('Stopping test executor.')
>     self.kill()
> ```

self.kill() will still trigger the exit stack to close, right?  Is this a case of whichever close gets there first will be called?  

As long as its run once and only once we are okay.

---

*[test/core/exe_test.py, line 104 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfrx8irVUIpgRrUUE9H:-Kfrx8irVUIpgRrUUE9I:b-yn0etn) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/test/core/exe_test.py#L104)):*
> ```Python
>     def cancel_phase(test):
>       test.dut_id = 'DUT ID'
>       executor.stop()
> ```

You don't need to pass the executor into this phase?  How does it get access to executor?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/541)*

fahhem:
Review status: 3 of 6 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[openhtf/core/phase_executor.py, line 200 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-KfrxrmAxe0Wx3XnCWTM:-KfsTYUWGj2t92DdT75o:b-ucfknh) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/phase_executor.py#L200)):*
<details><summary><i>Previously, kdsudac wrote…</i></summary><blockquote>

The logic in this while True block can be greatly simplified:
outcome = self._execute_phase_once(phase)
if outcome is openhtf.PhaseResult.REPEAT and repeat_count < repeat_limit:
  continue
return outcome
  
</blockquote></details>

Thanks!

---

*[openhtf/core/phase_executor.py, line 243 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-KfrzCqvMGxCrvq6VNVU:-KfsTg8ImMSbBK9osIDo:b-15xhix) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/phase_executor.py#L243)):*
<details><summary><i>Previously, kdsudac wrote…</i></summary><blockquote>

Lol, took me a bit to figure out what this was doing.
</blockquote></details>

Hopefully it's better now?

---

*[openhtf/core/test_executor.py, line 42 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfs72_DJjUs7YhVoxlY:-KfsU07DCLEN0Mau553I:b-896fix) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/test_executor.py#L42)):*
<details><summary><i>Previously, kdsudac wrote…</i></summary><blockquote>

Can we increase default?  Maybe to 10 seconds?

If timeout it too low, you end up with race condition (which now I think will just be the original phase racing against test teardown).

If timeout is too high, you just end up wasting a little time.

</blockquote></details>

Done.

---

*[openhtf/core/test_executor.py, line 83 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfs51CeXsrdz04PXXOo:-KfsU2fJOg7EcsVZmabe:bo28sg4) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/openhtf/core/test_executor.py#L83)):*
<details><summary><i>Previously, kdsudac wrote…</i></summary><blockquote>

self.kill() will still trigger the exit stack to close, right?  Is this a case of whichever close gets there first will be called?  

As long as its run once and only once we are okay.
</blockquote></details>

Yeah, but `ExitStack` will only call each callback once

---

*[test/core/exe_test.py, line 104 at r4](https://reviewable.io:443/reviews/google/openhtf/541#-Kfrx8irVUIpgRrUUE9H:-KfsUl9ICDFzGfuvEbfx:bofb6km) ([raw file](https://github.com/google/openhtf/blob/6b1a5224aa601f99cfdda6902b267540e4635434/test/core/exe_test.py#L104)):*
<details><summary><i>Previously, kdsudac wrote…</i></summary><blockquote>

You don't need to pass the executor into this phase?  How does it get access to executor?
</blockquote></details>

It's because it's inside the scope where executor gets defined

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/541)*

fahhem:
The tests were previously pretty flaky, including monitoring. I've run them hundreds of times in a row and it's up to at least 99% (sometimes pytest gets stuck after they've all run?) from 90%

fahhem:
ping

fahhem:
My team has held off on reviewing/approving this, but if you're not going to respond to PRs for a week we'll need to take a different approach

jettisonjoe:
@grybmadsci  and I have been pretty slammed with our own stuff recently. My understanding was that one of the reasons we made one of your dudes a maintainer was to make sure you were unblocked in this kind of scenario. I'm not sure where all this salt is coming from. I don't have any objection to you guys reviewing and approving each other's stuff.

fahhem:
This is a larger change that affects how the core executor works, and would benefit from a review from the people who wrote the executor as-is. I wrote a very different version of all this for the previous version.

When you added 'my dude', you also made it clear that we're still second-class citizens. I have write access, but not admin access. 'My dude' was added for his frontend experience and there were comments about being sure we didn't make a bunch of changes just for my team. This salt is because at every turn in this codebase, I'm meeting race conditions, poorly designed code, false features (where it's claimed to work, but doesn't), and more. I'm an inch away from giving up on OpenHTF entirely.

grybmadsci:
Whoa there with the hostility guys.  I apologize for the delay, this is obviously a core change and warrants a review from a few of the maintainers.  I was out of town on international travel for a couple weeks and missed the traffic on this.

Notably, though, I have zero emails addressed directly to me and zero chats from anybody asking me to take a look at this.  If you want me to take a look at something, all you've got to do is ask :)

As for any "need to take a different approach", let's chat in the maintainer's meeting, which I think is a more appropriate forum - I know with my travel and the Daydream team being so slammed we haven't had a chance to meet in a while, sorry about that.

openhtf/core/test_executor.py
```diff
@@ -115,45 +124,64 @@ def _thread_proc(self):
...
 
-      # Everything is set, set status and begin test execution.  Note we don't
-      # protect this with a try: block because the PhaseExecutor handles any
-      # exceptions from test code.  Any exceptions here are caused by the
-      # framework, and we probably want them to interrupt framework state
-      # changes (like the transition to FINISHING).
+      # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
 
+  def _execute_test_start(self, phase_exec):
```
grybmadsci: This name (and even the docstring) kind of hides that this is also where plug initialization happens.  Maybe pull just that first line out and keep it above at line 137?  I think the plug init -> start phase -> test phases -> finalize flow should be visible up there.

fahhem: I disagree, the executor is just starting the test. Some plugs getting initialized is an implementation detail of how we do that, and would just clutter up the main flow.

PR #543, opened 2017-03-18
Return prompt ID from UserInput.start_prompt

fahhem:
This allows non-interactive prompts to be created:

```py
@htf.plugs(user=user_input.UserInput)
def ShowPromptForASecond(user):
  prompt_id = user.start_prompt(message='Hey Operator!')
  time.sleep(1)
  user.respond(prompt_id, 'ignored')
```

The other option is to make `remove_prompt()` public


Kenadia:
Closed in favor of #544

PR #544, opened 2017-03-18
Allow UserInput prompts to be dismissed

fahhem:
This allows non-interactive prompts to be created:

```py
@htf.plugs(user=user_input.UserInput)
def ShowPromptForASecond(user):
  prompt_id = user.start_prompt(message='Hey Operator!')
  time.sleep(1)
  user.remove_prompt()
```

This is an alternative to #543

### DIFF
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -153,8 +153,9 @@ class UserInput(plugs.FrontendAwareBasePlug):
 
     self._console_prompt.start()
     self.notify_update()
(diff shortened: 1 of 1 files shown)


Kenadia:
Making remove_prompt public seems good to me. `user.remove_prompt()` reads slightly better than `user.respond(prompt_id, 'ignored')`

Any objection to merging this instead of #543 ?

jettisonjoe:
Either way works for me.

PR #545, opened 2017-03-21
Make UserInput._remove_prompt public

kdsudac:
UserInput.prompt will raise a PromptUnansweredError.  You can catch these exceptions and react appropriately, but won't be able to fully recover because of MultiplePromptsError.


kdsudac:
Ha, I see multiple other PR's doing same thing.  Closing.

PR #546, opened 2017-03-23
Making attach_from_file handle paths

kdsudac:
if you call test.attach_from_file('/foo/bar/baz.txt') it will save attachment
Existing behavior:   attachment is named '/foo/bar/baz.txt'
After this commit:  attachment is named baz.txt

Note: this could break existing tests that rely upon file directory to be in name to ensure uniqueness.

### DIFF
--- a/openhtf/core/test_state.py
+++ b/openhtf/core/test_state.py
@@ -366,7 +366,7 @@ class PhaseState(mutablerecords.Record('PhaseState', [
     """
     with open(filename, 'rb') as f:  # pylint: disable=invalid-name
       self.attach(
-          os.path.basename(name) if name is not None else filename, f.read(),
+          name if name is not None else os.path.basename(filename), f.read(),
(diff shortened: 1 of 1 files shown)


openhtf/core/test_state.py
```diff
@@ -365,7 +366,7 @@ def attach_from_file(self, filename, name=None, mimetype=None):
     """
     with open(filename, 'rb') as f:  # pylint: disable=invalid-name
       self.attach(
-          name if name is not None else filename, f.read(),
+          os.path.basename(name) if name is not None else filename, f.read(),
```
fahhem: I think you meant to do os.path.basename(filename)

name should be used as-is, since it's a more explicit way to provide the attachment name

kdsudac: yep, os.path.basename(filename).  Fixed.

PR #550, opened 2017-03-27
Expose web_gui config info to frontend

Kenadia:
This supports use cases such as:
- Distinguishing development and production modes in the frontend Angular app
- Customizing the frontend for different deployments using command line arguments


openhtf/output/web_gui/__init__.py
```diff
@@ -445,6 +445,16 @@ def initialize(self, port):  # pylint: disable=arguments-differ
...
 
+  class ConfigHandler(tornado.web.RequestHandler):
+    """Provides configuration information.
+
+    This exposes the config file or flags used to run the web GUI.
+    Note that this is distinct from the configuration of test stations.
+    """
+
+    def get(self):
+      self.write(conf._asdict())
```
fahhem: Can you make this a dict so we can potentially add non-conf variables to it and stay wire compatible?

Kenadia:
@fahhem

Kenadia:
Clicking too fast. @fahhem I wanted to get clarification of the case you're thinking of and what you mean by wire compatible.

fahhem:
I meant that we may want to provide non-conf data, such as flags, in the same handler (since there's no benefit to proliferating handlers for more 'static' data).

Kenadia:
Updated to send the CLI args as well. PTAL. See example screenshot below:

screen shot 2017-03-27 at 4 19 35 pm

PR #551, opened 2017-03-29
Fixes for phase descriptors in web gui

Kenadia:
Follow-up to #534 to update backend so that pending phases are accessible from the frontend.


Kenadia:
Updated to respond to comments from Farz.

Main change was to change
```
return self.stations.get(hostport)
```
to
```
return self.stations[hostport]
```
in StationStore

PR #553, opened 2017-04-04
Fix util.data.convert_to_base_types

Kenadia:
Ensure `convert_to_base_types` produces types that are JSON serializable and XML marshallable.

For example, previously, if a measurement got a numpy float value, the web gui would fail with an error like the following:

```
Fault: <Fault 1: "<type 'exceptions.TypeError'>:cannot marshal <type 'numpy.float64'> objects">
```


openhtf/util/data.py
```diff
@@ -135,21 +136,27 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
...
-    obj = str(obj)
 
-  return obj
+  # Preserve booleans, None, and unicode.
+  elif obj in (True, False, None) or type(obj) is unicode:
+    return obj
+
+  # Convert numeric types (e.g. numpy ints and floats) into built-in types.
+  elif isinstance(obj, numbers.Integral):
+    return int(obj)
```
fahhem: long()

openhtf/util/data.py
```diff
@@ -135,21 +136,27 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
...
         convert_to_base_types(value, ignore_keys, tuple_type) for value in obj)
-  elif obj is not None and (
-      not isinstance(obj, numbers.Number) and not isinstance(obj, basestring)):
-    # Leave None as None to distinguish it from "None", as well as numbers and
-    # strings, converting anything unknown to strings.
-    obj = str(obj)
 
-  return obj
+  # Preserve booleans, None, and unicode.
+  elif obj in (True, False, None) or type(obj) is unicode:
```
fahhem: PASSTHROUGH_TYPES = {bool, type(None), unicode, bytes}

type(obj) in PASSTHROUGH_TYPES

test/util/data_test.py
```diff
@@ -0,0 +1,52 @@
...
+      pass
+
+    example_data = {
+      'list': [10],
+      'tuple': (10,),
+      'str': '10',
+      'unicode': u'10',
+      'int': 10,
+      'float': 10.0,
+      'long': 10000000000000000000,
```
fahhem: 2 ** 40

fahhem: Also add something for 2 ** 80, which is a long and can't be an int

openhtf/util/data.py
fahhem: maybe include int and long as well, and do this check first since it's more likely (more leaves than parents in the tree)

test/util/data_test.py
Kenadia: `10000000000000000000` is a long but I can make it `2 ** 80` for clarity

is there a reason I should also check 2 ** 40?

openhtf/util/data.py
Kenadia: I don't think `long` is preferable, consider the following:

```
>>> type(int(2 ** 64))
<type 'long'>
>>> type(int(numpy.uint64(2 ** 64 - 1)))
<type 'long'>
>>> type(long(1))
<type 'long'>
```

I can't think of an edge case in which `long` helps us—but I admit we might want to use it just because it *looks* safer.

openhtf/util/data.py
Kenadia: Done

openhtf/util/data.py
Kenadia: Done

test/util/data_test.py
Kenadia: Done

openhtf/util/data.py
```diff
@@ -31,6 +31,9 @@
 
 from enum import Enum
 
+# Used by convert_to_base_types().
+PASSTHROUGH_TYPES = {bool, bytes, int, long, type(None), unicode}
```
fahhem: should include `str`, too, otherwise we're just copying them for no reason

Kenadia:
```
>>> str is bytes
True
```

PR #554, opened 2017-04-07
make phases options available in test_record

wallacbe:


fahhem:
@wallacbe you need to update the test data used to have the `options` field

wallacbe:
Done, PTAL

openhtf/core/test_record.py
```diff
@@ -70,7 +70,7 @@ def add_outcome_details(self, code, description=''):
 
 class PhaseRecord(  # pylint: disable=no-init
     mutablerecords.Record(
-        'PhaseRecord', ['descriptor_id', 'name', 'codeinfo'],
+        'PhaseRecord', ['descriptor_id', 'name', 'codeinfo', 'options'],
```
fahhem: If you make this optional, we stay compatible with previous serialized `TestRecord` objects

wallacbe:
Observe that when made optional it results in attribute error with prior test record in unit test.   This is why I regenerated the test record.

openhtf/util/data.py:130: in convert_to_base_types
    for attr in type(obj).all_attribute_names
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <itertools.chain object at 0x7f0741be5a10>

    obj = {attr: getattr(obj, attr)
           for attr in type(obj).all_attribute_names
>          if (getattr(obj, attr) is not None or
E          AttributeError: options

openhtf/util/data.py
```diff
@@ -126,10 +126,13 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
   if hasattr(obj, '_asdict'):
     obj = obj._asdict()
   elif isinstance(obj, records.RecordClass):
-    obj = {attr: getattr(obj, attr)
-           for attr in type(obj).all_attribute_names
-           if (getattr(obj, attr) is not None or
-               attr in type(obj).required_attributes)}
+    temp = {}
+    for attr in type(obj).all_attribute_names:
+      if hasattr(obj, attr):
```
grybmadsci: i think it would be better to `getattr(obj, attr, None)` instead of doing the hasattr check

test/output/callbacks/callbacks_test.py
```diff
@@ -69,6 +69,8 @@ def setUp(cls):
   def test_json(self):
     json_output = StringIO()
     json_factory.OutputToJSON(json_output, sort_keys=True, indent=2)(self.record)
+
```
grybmadsci: no need to add blank lines (especially not two)

wallacbe:
Made updates

wallacbe:
PTAL, this is  precursor for adding updates to retry mechanism .  Thanks

test/core/measurements_test.py
```diff
@@ -34,7 +34,8 @@
 
 # Fields that are considered 'volatile' for record comparison.
 _VOLATILE_FIELDS = {'start_time_millis', 'end_time_millis', 'timestamp_millis',
-                    'lineno', 'codeinfo', 'code_info', 'descriptor_id'}
+                    'lineno', 'codeinfo', 'code_info', 'descriptor_id',
+                    'options'}
```
fahhem: options shouldn't be a volatile field...

wallacbe:
Made changes

test/core/measurements_test.py
wallacbe: Done

openhtf/util/data.py
```diff
@@ -126,10 +126,12 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
   if hasattr(obj, '_asdict'):
     obj = obj._asdict()
   elif isinstance(obj, records.RecordClass):
-    obj = {attr: getattr(obj, attr)
-           for attr in type(obj).all_attribute_names
-           if (getattr(obj, attr) is not None or
-               attr in type(obj).required_attributes)}
+    temp = {}
```
fahhem: why did this file change? I don't see how this changed behavior?

PR #555, opened 2017-04-19
Make phase result log more useful

fahhem:


fahhem:
Good point, I thought the "Finishing test execution...." wasn't getting into the output, not just as DEBUG. I'll change to that

Issue #556, opened 2017-04-19
decorating phase with "run_if=lambda: False" raises error in TestExecutorThread


wallacbe:
cloned repo at head, added the following phase to all_the_things.py

```
@htf.TestPhase(run_if=lambda: False)
def MyPhase(test):
  print 'My phase'

```
See the following:

ERROR:root:Thread raised an exception: TestExecutorThread
Traceback (most recent call last):
  File "/home/wallaceb/source/openhtf/openhtf/util/threads.py", line 60, in run
    self._thread_proc()
  File "/home/wallaceb/source/openhtf/openhtf/core/test_executor.py", line 150, in _thread_proc
    self._execute_test_phases(phase_exec)
  File "/home/wallaceb/source/openhtf/openhtf/core/test_executor.py", line 180, in _execute_test_phases
    outcome = phase_exec.execute_phase(phase)
  File "/home/wallaceb/source/openhtf/openhtf/core/phase_executor.py", line 202, in execute_phase
    if (outcome.is_repeat and repeat_count < repeat_limit):
AttributeError: 'NoneType' object has no attribute 'is_repeat'
Exception in thread TestExecutorThread:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/home/wallaceb/source/openhtf/openhtf/util/threads.py", line 60, in run
    self._thread_proc()
  File "/home/wallaceb/source/openhtf/openhtf/core/test_executor.py", line 150, in _thread_proc
    self._execute_test_phases(phase_exec)
  File "/home/wallaceb/source/openhtf/openhtf/core/test_executor.py", line 180, in _execute_test_phases
    outcome = phase_exec.execute_phase(phase)
  File "/home/wallaceb/source/openhtf/openhtf/core/phase_executor.py", line 202, in execute_phase
    if (outcome.is_repeat and repeat_count < repeat_limit):
AttributeError: 'NoneType' object has no attribute 'is_repeat'

PR #557, opened 2017-04-20
add checks for outcome=None

wallacbe:


wallacbe:
https://github.com/google/openhtf/issues/556

examples/all_the_things.py
```diff
@@ -95,12 +95,19 @@ def attachments(test):
   test.attach_from_file('example_attachment.txt')
 
 
+@htf.TestPhase(run_if=lambda: False)
```
fahhem: add a test for this please

openhtf/core/phase_executor.py
```diff
@@ -199,9 +199,10 @@ def execute_phase(self, phase):
     while not self._stopping.is_set():
       outcome = self._execute_phase_once(phase)
```
fahhem: make this return `PhaseOutcome(openhtf.PhaseResult.CONTINUE)` and undo everything else

grybmadsci: I think line 217 should return `PhaseOutcome(openhtf.PhaseResult.CONTINUE)` instead, returning None there breaks the calling convention of `_execute_phase_once()`.

examples/all_the_things.py
wallacbe: Added test in exe_test

wallacbe:
Made changes

PR #560, opened 2017-04-25
Allow measurement validators to receive phase arguments

fahhem:
This also removes the class-as-module trick in validators.py since it was causing problems (had to pass modules into the validators, causing deepcopy to try to deepcopy builtins like `NotImplementedType`)

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -483,7 +483,7 @@ class PhaseDescriptor(mutablerecords.Record(
         self,
         plugs=new_plugs.values(),
         options=self.options.format_strings(**subplugs),
-        measurements=[m.format_strings(**subplugs) for m in self.measurements])
+        measurements=[m.with_args(**subplugs) for m in self.measurements])
 
 
   def __call__(self, test_state):
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -461,7 +461,7 @@ class PhaseDescriptor(mutablerecords.Record(
     new_info = mutablerecords.CopyRecord(self)
     new_info.options = new_info.options.format_strings(**kwargs)
     new_info.extra_kwargs.update(kwargs)
-    new_info.measurements = [m.format_strings(**kwargs) for m in self.measurements]
+    new_info.measurements = [m.with_args(**kwargs) for m in self.measurements]
     return new_info
 
   def with_plugs(self, **subplugs):
--- a/openhtf/util/validators.py
+++ b/openhtf/util/validators.py
@@ -101,9 +101,6 @@ class InRange(object):
       return False
     minimum = self._type(self.minimum)
     maximum = self._type(self.maximum)
-    # Check for equal bounds first so we can use with non-numeric values.
-    if minimum == maximum and value != minimum:
-      return False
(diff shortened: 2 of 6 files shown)


openhtf/util/validators.py
```diff
@@ -50,124 +50,140 @@ def MyPhase(test):
...
+
+  def __ne__(self, other):
+    return not self == other
+
+in_range = InRange  # pylint: disable=invalid-name
+register(in_range, name='in_range')
+
+
+@register
+def equals(self, value, type=None):
```
Kenadia: I see that `type` is applied only at the time of comparison and is applied only to the “expected value” side of the Equals or InRange validation. What's the use case for this?

openhtf/util/validators.py
```diff
@@ -50,124 +50,140 @@ def MyPhase(test):
...
+  def __call__(self, value):
+    if value is None:
+      return False
+    import math
+    # Check for nan
+    if math.isnan(value):
+      return False
+    minimum = self._type(self.minimum)
+    maximum = self._type(self.maximum)
+    # Check for equal bounds first so we can use with non-numeric values.
```
Kenadia: Do you understand this comment? I'm have trouble understanding it

openhtf/util/validators.py
```diff
@@ -50,124 +50,140 @@ def MyPhase(test):
...
+    return value == self._type(self.expected)
+
+  def __str__(self):
+    return "'x' is equal to '%s'" % self.expected
+
+  def __eq__(self, other):
+    return isinstance(other, type(self)) and self.expected == other.expected
+
+
+class RegexMatcher(object):
```
grybmadsci: maybe postfix these helper classes with _, since the intention is for users to always use the lower_case validators.

openhtf/__init__.py
```diff
@@ -461,7 +461,7 @@ def with_args(self, **kwargs):
     new_info = mutablerecords.CopyRecord(self)
     new_info.options = new_info.options.format_strings(**kwargs)
     new_info.extra_kwargs.update(kwargs)
-    new_info.measurements = [m.format_strings(**kwargs) for m in self.measurements]
+    new_info.measurements = [m.with_args(**kwargs) for m in self.measurements]
```

openhtf/util/validators.py
fahhem: Ah, it was to make `in_range` support `equals` for non-integers by not doing < or > comparisons. Since we have an explicit Equals for non-integers now, this can go away

openhtf/util/validators.py
fahhem: It's to allow users to do this:

```py
@htf.measures(htf.Measurement('xyz').in_range('{voltage}', 20, type=float))
def Something(...):
  ...

Something.with_args(voltage=12)
```

That way, `Something`'s measurement 'xyz' will have a validator based on the arguments, but with the right type. We can't assume a particular type in case of precision errors

openhtf/util/validators.py
fahhem: Nope, these classes are part of the external API (for output callbacks)

fahhem:
Thanks, PTAL

PR #563, opened 2017-05-02
Fix some lint issues.

arsharma1:


arsharma1:
@fahhem

fahhem:
What created the Change-Id? Respond internally if it's an internal tool?

arsharma1:
Change-Id comes from Gerrit.  I was just about to remove that.

PR #565, opened 2017-05-03
Disable plug manager XML-RPC server when station_api_port is None

Kenadia:
~~E.g.~~
```
python examples/hello_world.py --config-value=enable_station_api=
```
~~(setting enable_station_api to the empty string because conf values can only be set to strings from the command line)~~


fahhem:
Need to fix the tests, but it looks like simple fixes

openhtf/plugs/__init__.py
```diff
@@ -417,7 +417,7 @@ def _create_or_update_rpc_server(self):
             attr_name not in plug.disable_remote_attrs):
           plug_methods.append((attr, '.'.join(('plugs', name, attr_name))))
 
-    if not plug_methods:
+    if not plug_methods or not conf.enable_station_api:
```
grybmadsci: There's already a conf option for doing this, it's setting a station_api_port of None, I think the only change you need is to check that here and leave the other stuff as-is.

Kenadia:
Rewrote in response to @grybmadsci's review. Note that I think this makes it impossible to disable station API via command-line flags alone bug I think that's okay.

PTAL @fahhem

grybmadsci:
Yeah I thought of that - the commandline flag not working now.  Commandline config values weren't meant to be used outside of debugging, anyway.  I'd be okay with adding a --disable-station-api flag that overrides the conf value as a convenience thing if someone really needs the flag, but it's a bit of a hack.

PR #566, opened 2017-05-03
Add method to get plug instance from manager

Kenadia:


openhtf/plugs/__init__.py
```diff
@@ -481,6 +481,10 @@ def initialize_plugs(self, plug_types=None):
       self.update_plug(plug_type, plug_instance)
     self._create_or_update_rpc_server()
 
+  def get_plug(self, plug_name):
```
fahhem: `get_plug_by_class_path` or something, since `plug_name` isn't 'user', it's 'openhtf.plugs.user_input.UserInput'

Also, this deserves a bit more of a docstring, with at least an example of what this is and what it's meant for

Kenadia: I can call the method name `get_plug_by_class_path`.. but note that we call this the plug's “name” everywhere else (e.g. station_api, web_gui/__init__, elsewhere in plugs/__init__)

PR #576, opened 2017-05-12
Add test functions for phase outcomes.

arsharma1:
Include fix for the setting the PlugManager logger errors are better
understood.


arsharma1:
@fahhem

PR #577, opened 2017-05-12
util/test.py: Set phase outcome

arsharma1:


arsharma1:
@fahhem

fahhem:
Does this PR do the same thing? https://github.com/google/openhtf/pull/567

arsharma1:
That does a better job, other than also needing the change to assertPhaseContinue.

fahhem:
Okay, added that to #567 so closing this one.

PR #579, opened 2017-05-13
Phase returning STOP is supposed to cause a FAIL

fahhem:
Somehow it became ABORTED, probably on accident?


arsharma1:
It was changed 11 months ago: https://github.com/google/openhtf/commit/a9de101f7426274d4cbdc05f64a529bde65860bd

fahhem:
It looks like a typo because it was FAIL, it says FAIL, it's documented as FAIL, so it should be FAIL

PR #582, opened 2017-05-24
Shorten cancel timeout for phases to 2 seconds

fahhem:
Also fixes a pickling error that makes updating the measurements' pickle record broken.


Kenadia:
What's the pickling error and where's the fix?

fahhem:
The fix is the changes in validators, it doesn't use a lambda because you can't pickle those

openhtf/util/validators.py
```diff
@@ -88,7 +90,7 @@ def __init__(self, minimum=None, maximum=None, type=None):
       raise ValueError('Minimum cannot be greater than maximum')
     self.minimum = minimum
     self.maximum = maximum
-    self._type = type or (lambda x: x)
+    self._type = type
```
grybmadsci: is there a particular reason not to do `self._type = type or _identity` here?

fahhem: pickle is the problem, it wants to pickle the lambda and complains

PR #583, opened 2017-05-24
Stop executing phases if test_start fails

fahhem:
Any terminal outcome (timeout, stop, error) will stop a test early, rather than continuing the way it does now


fahhem:
Fixed the tests, PTAL

PR #585, opened 2017-05-25
Format the error in util.test correctly

fahhem:


arsharma1:
Line 317 has trailing whitespace, but I can't comment on that from this interface.

fahhem:
Done, PTAL

PR #564, opened 2017-05-02
Repeat skip validation

arsharma1:
@fahhem

PR #587, opened 2017-06-01
Allow phases to return SKIP

arsharma1:
Allow phases to return SKIP after they have started in case they need
some additional data to determine if they should actually run.

This functions differently than run_if in that if run_if returns false,
the phase is not logged.  If the phase returns SKIP, it will be logged.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -349,7 +349,7 @@ PhaseResult = Enum('PhaseResult', [
     # repeat_limit option, this will be treated as a STOP.
     'REPEAT',
     # Causes the framework to ignore the measurement outcomes and execute the
-    # next phase.
+    # next phase.  The phase is still logged, unlike with run_if.
     'SKIP',
     # Causes the framework to stop executing, indicating a failure.
     'STOP'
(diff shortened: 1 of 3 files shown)


grybmadsci:
What's the difference between returning SKIP and returning CONTINUE (or None, since None defaults to CONTINUE)?

arsharma1:
Returning SKIP will give a phase the SKIP outcome instead of using the measurements to determine its PASS or FAIL.  This was implemented for returning REPEAT ( #564) so that the tests won't fail.  This is extending that logic so phases can just SKIP themselves.

openhtf/__init__.py
```diff
@@ -349,7 +349,7 @@ def create_arg_parser(add_help=False):
     # repeat_limit option, this will be treated as a STOP.
     'REPEAT',
     # Causes the framework to ignore the measurement outcomes and execute the
-    # next phase.
+    # next phase.  The phase is still logged, unlike with run_if.
```
grybmadsci: can you add to this comment that the phase is logged, but can not cause the test to fail?

fahhem: The first part of the comment says the framework ignores the measurement outcomes

shaperilio:
I need to run a test where the operator chooses the phases at the start. Ideally, I wouldn't have to go and change any phase code, so I can quickly change the list of phases available to the operator. Ideally, I'd somehow alter the phase list and update the `run_if` option after the operator has chosen which phase he wants to run.

Is this possible?

arsharma1:
It's possible, but returning SKIP wouldn't be part of it unless you want to modify the phase code.  Please open an issue so discussing this will reach more people.

PR #589, opened 2017-06-07
Add a WithinPercent validator

glados-verma:
Equals is useful, but often we're fine having a measurement being within a range.  The WithinPercent validator provides one way of doing such checks.

Add unit tests module for validators, with unit tests (also serving as examples) for the WithinPercent validator.


openhtf/util/validators.py
```diff
@@ -193,3 +193,33 @@ def __ne__(self, other):
...
+            self.expected == other.expected and
+            self.percent == other.percent)
+
+  def __ne__(self, other):
+    return not self == other
+
+
+@register
+def within_percent(expected, percent):
+  return WithinPercent(expected, percent)
```
fahhem: I'd fall back to a subclass of InRange that implemented `__str__`. That way mfg_inspector.py and any other UI-related code would get the minimum/maximum attributes and the string representation they want:

```py
class WithinPercent(InRange):
  def __init__(self, expected, percent):
    delta = percent / 100.0
    super(WithinPercent, self).__init__(expected - delta, expected + delta)

  def __str__(self):
    ...
```

All your unit-tests should still pass, but there's both less code here and less changes everywhere else

glados-verma: PTAL. I defined an abstract base class to represent an "in-range" type of validator, and am looking for that in mfg_inspector.
I wasn't able to have WithinPercent nicely subclass from InRange - the functionality put in place to allow with_args(...) type of usage requires overriding the entire implementation of InRange.

openhtf/util/validators.py
```diff
@@ -75,9 +76,30 @@ def create_validator(name, *args, **kwargs):
 
 _identity = lambda x: x
 
+
+class ValidatorBase(object):
```
fahhem: This implies that validators must be classes, but they just need to be callables. We use classes so there are attributes, but a validator like this is also fine: `lambda value: value == 1`

glados-verma: Added some documentation clarifying that validators do not need to be classes, but if being implemented as such, there are benefits to inheriting them from the base classes.

glados-verma:
Removed change detector tests, which has led to coverage drop. This is mainly because the pickle file I deleted from callbacks_test was generated using all_the_things (an example), an indirect dependency that's not desirable anyway.

What would you like me to do for the coverage drop? In a subsequent CL, I could check in a "representative test" that could exercise a bunch of features, leading to improved coverage. Thoughts?

openhtf/util/validators.py
fahhem: You didn't mention that it doesn't have to be a class, and the only example is of a class

glados-verma: Take a look at my edit on line 27; it segues into the non-class example (which was already there). I think that would be sufficient, right?

fahhem:
Thanks for contributing! Sorry it took so long!

PR #590, opened 2017-06-07
Add measurement-affecting .with_args to an example

fahhem:


examples/with_plugs.py
```diff
@@ -74,13 +74,17 @@ class PingDnsB(PingPlug):
...
+    'total_time_{pinger.host}_{count}',
+    htf.Measurement('retcode').equals('{expected_retcode}', type=int)
+)
+def test_ping(test, pinger, count, expected_retcode):
   """This tests that we can ping a host.
 
   The plug, pinger, is expected to be replaced at test creation time, so the
   placeholder property was used instead of the class directly.
   """
+  del expected_retcode  # Not used in the phase, only used by a measurement
```
kdsudac: I'd probably leave this line out because I think it is likely to confuse people new to htf.

You are really deleting only for linter, not really because it's a good practice.

fahhem: I'm making it clear that it has to be an input argument but doesn't have to be used in the phase

PR #595, opened 2017-06-13
Add tests for measurements module

glados-verma:
PTAL, if this looks good, I'll add similar tests to callbacks_test.py.
I think that test file should be split into separate test modules for mfg_inspector and json_factory - thoughts?

---

The fact that conf is a global required editing conf_test.py because I imported all_the_things.py in a test.

### DIFF
--- a/examples/all_the_things.py
+++ b/examples/all_the_things.py
@@ -18,16 +18,14 @@ Run with (your virtualenv must be activated first):
 python all_the_things.py
 """
 
-import json
-import os
 import time
 
 import openhtf as htf
+from openhtf import util
 from openhtf.util import units
 from openhtf.plugs import user_input
 from openhtf.output import callbacks
 from openhtf.output.callbacks import json_factory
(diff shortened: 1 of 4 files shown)


kdsudac:
It looks like changes in this CL have broken examples/all_the_things.py (unless the test is design to FAIL):

@htf.measures(
    htf.Measurement('replaced_min_only').in_range('{min}', 5, type=int),
    htf.Measurement('replaced_max_only').in_range(0, '{max}', type=int),
    htf.Measurement('replaced_min_max').in_range('{min}', '{max}', type=int),
)
def measures_with_args(test, min, max):
  test.measurements.replaced_min_only = 1
  test.measurements.replaced_max_only = 1
  test.measurements.replaced_min_max = 1

The validators do indeed change, but the measurements set are all fixed at 1.

glados-verma:
It is designed to fail on some measurements to demonstrate full functionality. I don't believe there's an expectation that the example must give a passing result.

PR #597, opened 2017-06-21
Always remove attachment data when converting to dict

Kenadia:
This PR aims to standardize the dictionary format of attachments.

In the use cases I've seen, we always remove attachment data at some point—with this PR, it will happen by default when calling data.convert_to_base_types.

### DIFF
--- a/openhtf/core/test_record.py
+++ b/openhtf/core/test_record.py
@@ -48,6 +48,12 @@ class Attachment(collections.namedtuple('Attachment', 'data mimetype')):
   def sha1(self):
     return hashlib.sha1(self.data).hexdigest()
 
+  def _asdict(self):
+    # Don't include the attachment data when converting to dict.
+    return {
+        'mimetype': self.mimetype,
+        'sha1': self.sha1,
+    }
 
 class TestRecord(  # pylint: disable=no-init
     mutablerecords.Record(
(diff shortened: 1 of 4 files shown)


openhtf/core/test_record.py
```diff
@@ -48,6 +48,12 @@ class Attachment(collections.namedtuple('Attachment', 'data mimetype')):
   def sha1(self):
     return hashlib.sha1(self.data).hexdigest()
 
+  def _asdict(self):
+    # Don't include the attachment data when converting to dict.
+    return {
```
fahhem: You'll need to update the tests to handle this not including the data field anymore

Kenadia: Looks like this isn't a problem with the test, rather I broke the`inline_attachments` feature in json_factory.

I can't find anyone that's using this feature, but I've added a commit to make it work it just in case. I tested it with all_the_things.py.

fahhem: Not all our users are on github or internal, so we'd have to communicate removing such a feature pretty loudly. But you fixed it so no communication needed!

Kenadia: @shaperilio 
Sorry for the slow reply. Before this PR, attachment data was being removed in PhaseState._asdict(). This PR makes the dictionary format of attachments more consistent. The original motivation for removing attachment data is to ensure that a test state can be rapidly serialized by the station API in order to support frontends. The entire test state might be serialized many times per second to keep a frontend up to date.

If you'd like to create a JSON test record that includes attachments you can add use the OutputToJSON output callback in your test with `inline_attachments` set to True.

shaperilio: Thanks @Kenadia

PR #599, opened 2017-06-23
Mock out station_api.ApiServer in test.TestCase

kdsudac:
Test instantiation calls station_api.start_server() that starts an ApiServer.  This causes a traceback to show in test logs in certain testing environments where opening sockets is not allowed.

In general, starting a server as a side effect of creating an openhtf.Test instance is a bad practice for unit tests.


fahhem:
I agree, I thought we stopped the api server from coming up in other ways, but I can't find them now.

PR #605, opened 2017-06-29
Removed 404 images and made them just links

CameronSamuels:
They should be added back once the images are relinked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,7 @@
 # OpenHTF
 The open-source hardware testing framework.
 
-[![Build Status](https://travis-ci.org/google/openhtf.svg?branch=master)](https://travis-ci.org/google/openhtf)
-[![Coverage Status](https://coveralls.io/repos/google/openhtf/badge.svg?branch=master&service=github)](https://coveralls.io/github/google/openhtf?branch=master)
-
-[![Issue Stats](http://issuestats.com/github/google/openhtf/badge/pr)](http://issuestats.com/github/google/openhtf)
-[![Issue Stats](http://issuestats.com/github/google/openhtf/badge/issue)](http://issuestats.com/github/google/openhtf)
-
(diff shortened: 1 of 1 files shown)


fahhem:
I see the build and coverage badges working, are they 404's for you as well?

CameronSamuels:
Yes, they are 404. I even copied the source urls and pasted into the omnibox and it said an error message like 404.

README.md
```diff
@@ -3,12 +3,7 @@
 # OpenHTF
 The open-source hardware testing framework.
 
-[![Build Status](https://travis-ci.org/google/openhtf.svg?branch=master)](https://travis-ci.org/google/openhtf)
```
fahhem: This works both in my browser (redirects to https://api.travis-ci.org/google/openhtf.svg?branch=master) and in GitHub. Same with coverage. PR stats only works in my browser, while issue stats works but says Unknown. Can we at least keep the first two if there's no obvious reason for the badges to not work in GitHub or when viewed by others?

jettisonjoe: Same for me. Only the issuestats ones are broken. I'm in favor of leaving the Travis and Coveralls ones in, and only removing the broken ones.

CameronSamuels: Did I remove non-broken ones? Because at the time, i only removed the broken ones.

Sorry for the long response. I never got a notification.

CameronSamuels:
Now the images are working. Should I close this pr?

jettisonjoe:
The issuestats images are super inconsistent, and I think you should still remove those. The build status and coverage status images have never 404'd for me, and I'd like to see those left in. In your currently pushed commit, _all_ images appear to be removed and replaced with text links.

CameronSamuels:
Oh sorry about that. I will add the non issue stats images back.

CameronSamuels:
Now in pr #607 because I deleted the fork.

PR #607, opened 2017-07-17
Removed issue stats images from readme

CameronSamuels:
These are broken images every now and then.

Originated from pr #605 but I deleted the fork.


jettisonjoe:
It might be cool to leave maybe a single link to the issuestats page even though the badges are being removed.

jettisonjoe:
That CI check failing has to do with the PyOpenSSL version I think. We need to update our setup.py to require a newer version.

CameronSamuels:
I just noticed that too.

fahhem:
I don't think this sort of change should require multiple maintainers to approve. Sorry that this one took so long and so many changes to get in

PR #608, opened 2017-07-18
Only calling _initialize_value on Measurement if measured_value isn't passed in.

kdsudac:
Currently cannot deepcopy a Measurement where the value is set due to call to _initialize_value which will raise a ValueErrror:
```
def _initialize_value(self):
  if self.measured_value and self.measured_value.is_value_set:
    raise ValueError('Cannot update a Measurement once a value is set.')
```

Overwriting deepcopy to sidestep ValueError.

### DIFF
--- a/openhtf/core/measurements.py
+++ b/openhtf/core/measurements.py
@@ -132,7 +132,7 @@ class Measurement(  # pylint: disable=no-init
 
   def _initialize_value(self):
     if self.measured_value and self.measured_value.is_value_set:
-        raise ValueError('Cannot update a Measurement once a value is set.')
+      raise ValueError('Cannot update a Measurement once a value is set.')
     if self.dimensions:
       self.measured_value = DimensionedMeasuredValue(
           self.name, len(self.dimensions))
--- a/openhtf/core/measurements.py
+++ b/openhtf/core/measurements.py
@@ -62,6 +62,7 @@ Examples:
 
 
 import collections
+import copy
 import logging
 
 from enum import Enum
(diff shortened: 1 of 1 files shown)


openhtf/core/measurements.py
```diff
@@ -139,6 +140,19 @@ def _initialize_value(self):
     else:
       self.measured_value = MeasuredValue(self.name)
 
+  def __deepcopy__(self, memo):
```
fahhem: Another option would be to make `__init__` do this, right? It would be simpler in there, and useful for other non-copying code as well

kdsudac: If we're going to side-step the ValueError there, then we might as well get rid of that check all together.  I'd be in favor of removal, but I assumed it was there for a reason.

Kenadia: The measured_value check seems like it's still useful because of how measured_value is manipulated when setting dimensions in __setattr__. @fahhem's suggestion might be an easy fix, it makes sense that initializing a Measurement with a measured_value should be allowed.

```
  def __init__(self, name, **kwargs):
    measured_value = kwargs.pop('measured_value', None)
    super(Measurement, self).__init__(name, **kwargs)
    self._initialize_value()
    self.measured_value = measured_value
```

kdsudac: How's this?  I only call _initalize_value() if measured_value is not passed in.  More intuitive to only initialize if value is not passed vs initializing while sidestepping ValueError then overriding.

Kenadia: Ah makes sense since the purpose of _initialize_value is just to set measure_valued

kdsudac:
Is it okay to merge this? or should I wait until tests run by travis CI are fixed in head/master?

fahhem:
They've been fixed in master, please rebase and push to have travis retry

PR #610, opened 2017-07-19
Add ordering to measurements

Kenadia:
Use OrderedDict instead of dict for measurements on PhaseState and PhaseRecord. This allows the test author to fix an ordering for measurements in console output.

The measurements will keep the order in which they were declared, e.g.

```
@htf.measures(
    htf.Measurement('first'),
    htf.Measurement('second'),
    htf.Measurement('third'),
)
```

(Also, fix `callbacks/console_summary.py`)

### DIFF
--- a/openhtf/core/test_state.py
+++ b/openhtf/core/test_state.py
@@ -414,16 +414,15 @@ class PhaseState(mutablerecords.Record(
     Any UNSET measurements will cause the Phase to FAIL unless
     conf.allow_unset_measurements is set True.
     """
-    # Clear notification callbacks for later serialization.
-    for meas in self.measurements.values():
-      meas.set_notification_callback(None)
-
-    # Validate multi-dimensional measurements now that we have all values.
     for measurement in self.measurements.itervalues():
-      measurement.validate()
+      # Clear notification callbacks for later serialization.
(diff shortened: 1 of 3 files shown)


openhtf/core/test_state.py
```diff
@@ -416,22 +418,12 @@ def _finalize_measurements(self):
...
-        name: measurement.validate()
-        for name, measurement in self.measurements.iteritems()
-        if measurement.outcome is measurements.Outcome.PARTIALLY_SET
-    })
+    for measurement in self.measurements.itervalues():
+      measurement.validate()
 
     # Fill out final values for the PhaseRecord.
-    self.phase_record.measurements = validated_measurements
+    self.phase_record.measurements = self.measurements.copy()
```
fahhem: Why do we need to copy this? This is in `_finalize_measurements` which implies not touching these anymore

fahhem:
If you rebase onto master Travis will get past the openssl error

openhtf/core/test_state.py
Kenadia: Should be fine, as long as there's no code assuming phase_state.measurements and phase_record.measurements are separate objects

PR #614, opened 2017-07-25
Capture plug exceptions into the test record

fahhem:
(Still needs a test)

### DIFF
--- a/openhtf/core/test_executor.py
+++ b/openhtf/core/test_executor.py
@@ -146,6 +146,12 @@ class TestExecutor(threads.KillableThread):
 
       # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
+      
+  def _thread_exception(self, exc_type, exc_val, exc_tb):
+    """When the framework throws an exception, turn it into a failure code."""
+    if self.test_state and not self.test_state.is_finalized:
+      self.test_state.test_record.add_outcome_details(exc_type.__name__, exc_val.message)
+    return True
 
   def _execute_test_start(self, phase_exec):
     """Run the start trigger phase, and check that the DUT ID is set after.


openhtf/core/test_executor.py
```diff
@@ -146,6 +146,12 @@ def _thread_proc(self):
 
       # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
+      
+  def _thread_exception(self, exc_type, exc_val, exc_tb):
+    """When the framework throws an exception, turn it into a failure code."""
+    if self.test_state and not self.test_state.is_finalized:
+      self.test_state.test_record.add_outcome_details(exc_type.__name__, exc_val.message)
+    return True
```
jettisonjoe: This is going to be a departure from the previous logic, which suppressed only in the case of ThreadTerminationError. Is that intentional?

test/core/exe_test.py
```diff
@@ -224,3 +249,4 @@ def test_execute_run_if_false(self):
   def test_execute_phase_return_skip(self):
     result = self.phase_executor.execute_phase(phase_return_skip)
     self.assertEqual(PhaseResult.SKIP, result.phase_result)
+
```
Kenadia: Extra newline

openhtf/core/test_executor.py
```diff
@@ -146,6 +146,12 @@ def _thread_proc(self):
 
       # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
+      
+  def _thread_exception(self, exc_type, exc_val, exc_tb):
+    """When the framework throws an exception, turn it into a failure code."""
+    if self.test_state and not self.test_state.is_finalized:
+      self.test_state.test_record.add_outcome_details(exc_type.__name__, exc_val.message)
```
Kenadia: Line length > 80

openhtf/core/test_executor.py
```diff
@@ -146,6 +146,12 @@ def _thread_proc(self):
 
       # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
+      
+  def _thread_exception(self, exc_type, exc_val, exc_tb):
+    """When the framework throws an exception, turn it into a failure code."""
```
Kenadia: I think we should clarify `When the framework throws an exception` a little more. Looks like this is called when an exception is raised inside a test but outside of all phases?

openhtf/core/test_executor.py
Kenadia: Agree we should look into this more.

- Looking at `util/threads.py` it seems like this is mainly a matter of logging (if the exception is suppressed, it is not printed or logged, otherwise, it is sent to both logging.exception and stderr) or am I missing another side effect that can occur as a result of the thread raising an exception?
- It seems like returning True here risks silencing some errors completely (except for the outcome detail added to the test record). Errors in plug initialization are logged in PlugManager.initialize_plugs but other errors might be silenced.
- I noticed PhaseExecutorThread has sort of the opposite behavior—it never suppresses exceptions. It also makes its own call to logger.exception, so in the console you get “Phase <...> raised an exception” followed by “Thread raised an exception” followed by “Exception in thread” (three printouts of the same exception). We should see if the logging behavior can be more consistent (and not unnecessarily repetitive?).

wallacbe:
I'm also not seeing the logger in the plug show up in the test record, example self.logger.info('blah').  This this address that also?  Thanks

fahhem:
I don't think so. That should be in the test record already...

wallacbe:
Ah I see it in test record when run within a phase, I don't see it when called in tearDown

jettisonjoe:
@Kenadia is this PR still needed? Looks like it's getting stale sitting here.

wallacbe:
Having plug exceptions into the test record would still be useful

PR #616, opened 2017-07-26
Add FAIL_AND_CONTINUE to just fail a phase.

arsharma1:
Add FAIL_AND_CONTINUE to PhaseResult to indicate to the executor that
the phase should just be marked as a failure while continuing onto the
next phase.


antonysigma:
@fahhem A follow-up question: if I enable the web frontend, will it show a green color (indicating a PASS) or a red color?

I can confirm that `OutputToJSON` returns the `FAIL_AND_CONTINUE` state, but not sure how the web frontend will behave.

fahhem:
I'm not working on operhtf anymore, I recommend opening a new issue or something, sorry!

PR #618, opened 2017-07-26
Replace gflags with argv.

arsharma1:
Make the usb module more portable by removing the gflags library.

### DIFF
--- a/openhtf/util/argv.py
+++ b/openhtf/util/argv.py
@@ -3,7 +3,7 @@
 StoreInModule:
   Enables emulating a gflags-esque API (flag affects global value), but one
   doesn't necessarily need to use flags to set values.
-  
+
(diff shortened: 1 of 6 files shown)


openhtf/util/argv.py
```diff
@@ -41,3 +41,35 @@ def __call__(self, parser, namespace, values, option_string=None):
     else:
       module = __import__(self._tgt_mod)
     setattr(module, self._tgt_attr, values)
+
+
+class _StoreValueInModule(argparse.Action):
```
fahhem: merge with storeinmodule?

arsharma1: Merged as much as possible, following argparse's _StoreConstAction, _StoreTrueAction, and _StoreFalseAction.

Issue #559, opened 2017-04-20
append_record() in history.py adding test time


wallacbe:
history call back is causing lag.  Pinpointed issue in append_record() 

Added prints in append_record():
1492674275.94 before line 135
1492674275.94 before line 136
1492674279.13 before line 137:
1492674282.73 after line 137

Options proposed in email discussion
drop all_tests_history,
Update total_size() memory size estimates
serialize to disk

glados-verma:
Stale; I couldn't find a `history.py` or `append_record`.

PR #621, opened 2017-07-27
issue 559: comment out history

wallacbe:
https://github.com/google/openhtf/issues/559


wallacbe:
Removed from __init__.py.  PTAL

Issue #623, opened 2017-07-28
Upstream measurement/attachment retrieval into TestApi


kdsudac:

DanLipsitt:
I don't want to add too much noise to PR #624, so I'll ask here: 

@fahhem [said](https://github.com/google/openhtf/pull/624#issuecomment-319463318)
>Phase names don't have to be unique either, and phase uuid's change until the test starts, and you can reuse the same phase twice in a row so there's no possible way to uniquely identify a phase. Instead of dealing with all of that, we pushed the burden back and said if you need to retrieve a specific measurement, name it uniquely. If you reuse a phase, use .with_args to give each an index or name.

How does `with_args` help here?

fahhem:
If you have a phase repeating, here's what you can do:

@htf.measures(
    htf.Measurement('voltage_attempt{attempt}'),
)
def Phase(test, attempt):
  test.measurements['voltage_attempt%s' % attempt] = 9


htf.Test(Phase.with_args(attempt=1), Phase.with_args(attempt=2))

Or:

VOLTAGE_MEAS = htf.Measurement('voltage_attempt{attempt}')

@htf.measures(VOLTAGE_MEAS)
def Phase(test, attempt):
  test.measurements[VOLTAGE_MEAS.name] = 9

htf.Test(Phase.with_args(attempt=1), Phase.with_args(attempt=2))

fahhem:
If you have a phase repeating, here's what you can do:

```py
@htf.measures(
    htf.Measurement('voltage_attempt{attempt}'),
)
def Phase(test, attempt):
  test.measurements['voltage_attempt%s' % attempt] = 9


htf.Test(Phase.with_args(attempt=1), Phase.with_args(attempt=2))

Or:

VOLTAGE_MEAS = htf.Measurement('voltage_attempt{attempt}')

@htf.measures(VOLTAGE_MEAS)
def Phase(test, attempt):
  test.measurements[VOLTAGE_MEAS.name] = 9

htf.Test(Phase.with_args(attempt=1), Phase.with_args(attempt=2))
```

DanLipsitt:
Even though phases can repeat, they do help narrow down the scope when you're looking for a measurement a lot. You could have a ton of phases with a measurement called `voltage` and as long as the phases had different names you'd be fine. Here's what I just cobbled together:

```python
def get_measurement(test, phase_name, measurement_name):
    """Look up a measurement from a previous phase by name.

    Args:
        phase_name(str): The requested phase.
[...]

PR #624, opened 2017-07-28
Adding methods to TestApi to get attachments and measurements.

kdsudac:


openhtf/core/test_state.py
```diff
@@ -67,6 +67,22 @@ class BlankDutIdError(Exception):
...
+                           'name, value, units, dimensions, outcome')):
+  """Immutable copy of a measurement."""
+
+  @classmethod
+  def FromMeasurement(cls, measurement):
+    measured_value = measurement.measured_value
+    value = measured_value.value if measured_value.is_value_set else None
+    return cls(
+        measurement.name,
+        copy.deepcopy(value),
```
fahhem: This works with multi-dim measurements?

openhtf/core/test_state.py
```diff
@@ -130,7 +146,52 @@ def test_api(self):
...
+  def get_measurement(self, measurement_name):
+    """Get a copy of a measurement value from current or previous phase.
+
+    Args:
+      measurement_name: str of the measurement name
+    Returns:
+      an ImmutableMeasurement
+    """
+    # The framework ignores measurements from SKIP and REPEAT phases
+    ignore_outcomes = (test_record.PhaseOutcome.SKIP,)
```
fahhem: make this a set: `{test_record.PhaseOutcome.SKIP}` or just check it directly since it's only 1

openhtf/core/test_state.py
```diff
@@ -130,7 +146,52 @@ def test_api(self):
...
+    # Iterate through phases in reversed order for LIFO.
+    for phase_record in reversed(self.test_record.phases):
+      if (measurement_name in phase_record.measurements and
+          phase_record.result not in ignore_outcomes):
```
fahhem: I'd check the outcome first, since it's computationally cheaper than checking strings in a list

openhtf/core/test_state.py
```diff
@@ -130,7 +146,52 @@ def test_api(self):
...
+        measurement = phase_record.measurements[measurement_name]
+        return ImmutableMeasurement.FromMeasurement(measurement)
+
+    # Check current running phase state
+    if self.running_phase_state:
```
fahhem: I'd check the running phase first, since you also reverse the phase order. That way the checked phase order is "current->first" instead of "previous->first then current"

openhtf/core/test_state.py
```diff
@@ -130,7 +146,52 @@ def test_api(self):
...
+      if attachment_name in phase_record.attachments:
+        attachment = phase_record.attachments[attachment_name]
+        return copy.deepcopy(attachment.data)
+
+    # Check current running phase state
+    if self.running_phase_state:
+      if attachment_name in self.running_phase_state.phase_record.attachments:
+        attachment = self.running_phase_state.phase_record.attachments.get(
+            attachment_name)
+        return copy.deepcopy(attachment.data)
```
fahhem: The data is already a (immutable in Python) string, so copying it is a waste of resources.

Also, what about the mimetype?

openhtf/core/test_state.py
kdsudac: Yep, added an example to examples/all_the_things.py

openhtf/core/test_state.py
kdsudac: Done.

openhtf/core/test_state.py
kdsudac: Done.

openhtf/core/test_state.py
kdsudac: Done.

openhtf/core/test_state.py
kdsudac: attachment here is test_record.Attachment which is just a named tuple.  So attachment.data at this point hasn't been serialized so can be a mutable data type.

i.e. someone could do:

```
my_data = {'foo': 1}
test.attach('my_data.txt', my_data)
test.get_attachment('my_data.txt').update({'bar': 2})
```
Which would mutate my_data and the attachment.

kdsudac: Do people care about mimetype?

kdsudac:
Added unit tests.

openhtf/core/test_state.py
```diff
@@ -159,20 +159,21 @@ def get_measurement(self, measurement_name):
...
     """
     # The framework ignores measurements from SKIP and REPEAT phases
-    ignore_outcomes = (test_record.PhaseOutcome.SKIP,)
-    # Iterate through phases in reversed order for LIFO.
-    for phase_record in reversed(self.test_record.phases):
-      if (measurement_name in phase_record.measurements and
-          phase_record.result not in ignore_outcomes):
-        measurement = phase_record.measurements[measurement_name]
-        return ImmutableMeasurement.FromMeasurement(measurement)
+    ignore_outcomes = set([test_record.PhaseOutcome.SKIP,])
```
fahhem: Why not use `{test_record.PhaseOutcome.SKIP}` syntax?

openhtf/core/test_state.py
```diff
@@ -159,20 +159,21 @@ def get_measurement(self, measurement_name):
...
 
     # Check current running phase state
     if self.running_phase_state:
       if measurement_name in self.running_phase_state.measurements:
         return ImmutableMeasurement.FromMeasurement(
             self.running_phase_state.measurements[measurement_name])
 
+    # Iterate through phases in reversed order for LIFO.
```
fahhem: nit: 'for LIFO' seems a bit off? Just end the sentence after 'order'

test/test_state_test.py
```diff
@@ -0,0 +1,76 @@
...
+    attachment_name = 'attachment.txt'
+    input_contents = 'This is some attachment text!'
+    self.test_api.attach(attachment_name, input_contents)
+
+    output_contents = self.test_api.get_attachment(attachment_name)
+    self.assertEqual(input_contents, output_contents)
+
+  def test_get_attachment_immutable(self):
+    attachment_name = 'attachment.txt'
+    input_contents = {'foo': 'This is some attachment text in a dict!'}
```
fahhem: I'm not sure we want to support non-bytes attachments? It seems fine at first, but then everything dealing with attachments in any way (serialization, visualization, etc) has to check their Python type as well as mimetype. I think requiring it to be bytes/str and using the mimetype to interpret it is much better.

So really I'm just saying to not add this use of `test.attach(..., <non-string>)` into the openhtf codebase to avoid letting on that it technically works

openhtf/core/test_state.py
fahhem: I can see cases where you would want to be able to take the attachment, mutate it, then attach a new version: I grab an image and attach the raw image. In a later phase, I mark it up using CV to highlight edges, draw rectangles, etc, and attach that. In the latter phase, I'd want to know the original mimetype (was it a PNG, JPG, etc?) for feeding into my CV library, and potentially for serializing it back into the attachments.

Also, as in the comment in the unit-tests, while technically `attachment.data` can be anything, allowing non-strings doesn't make sense alongside mimetype, and adds complexity everywhere that deals with attachments

DanLipsitt:
It doesn't seem like measurement names are required to be unique. So doesn't there need to be a way to specify which phase to pull from?

fahhem:
We've thought about this a lot actually. Phase names don't have to be unique either, and phase uuid's change until the test starts, and you can reuse the same phase twice in a row so there's no possible way to uniquely identify a phase. Instead of dealing with all of that, we pushed the burden back and said if you need to retrieve a specific measurement, name it uniquely. If you reuse a phase, use `.with_args` to give each an index or name. That way, you can retrieve the correct measurement and be on your way. If you don't do that, you'll receive the later phase's measurement.

DanLipsitt:
So this will always give the most recent measurement with the given name? Is that guaranteed? Also, regardless of whether it's the current phase or a previous one?

fahhem:
Yes, that's the current design and what we've been using (only for ~2
months) internally

openhtf/core/test_state.py
```diff
@@ -130,7 +146,53 @@ def test_api(self):
...
+
+    # Iterate through phases in reversed order for LIFO.
+    for phase_record in reversed(self.test_record.phases):
+      if (phase_record.result not in ignore_outcomes and
+          measurement_name in phase_record.measurements):
+        measurement = phase_record.measurements[measurement_name]
+        return ImmutableMeasurement.FromMeasurement(measurement)
+
+    test.logger.warning('Could not find measurement: %s', measurement_name)
+    return None
```
DanLipsitt: Since measurements can be `None`, I don't think it's a good return value to use to mean "not found." In my [mockup](https://github.com/google/openhtf/issues/623#issuecomment-319538471) of this functionality I've used `LookupError` to signal that nothing valid is found.

kdsudac: This returns an ImmutableMeasurement which.  Even if a measurement value is set at None, it will show up as as the value attribute.  

```
test.measurements.my_measure = None
my_measure = test.get_measurement('my_measure')

my_measure is None # False
my_measure.value is None # True

other_measurement = test.get_measurement('other_measurement')
other_measurement is None # True

File: openhtf/core/test_state.py (hunk 5 above)
Comment (kdsudac) [reply openhtf/core/test_state.py]: Done.  Modified to have get_attachment return a copy of an Attachment, instead of just the contents.

File: test/test_state_test.py (hunk 8 above)
Comment (kdsudac) [reply test/test_state_test.py]: Agreed. Dropped this unit test case.

File: openhtf/core/test_state.py (hunk 5 above)
Comment (kdsudac) [reply openhtf/core/test_state.py]: Agreed.

**Comment by kdsudac** [comment]
Dusted off this PR.  I believe I address all concerns.

File: openhtf/core/test_state.py [hunk 10]
```diff
@@ -130,7 +146,53 @@ def test_api(self):
...
+  def get_measurement(self, measurement_name):
+    """Get a copy of a measurement value from current or previous phase.
+
+    Args:
+      measurement_name: str of the measurement name
+    Returns:
+      an ImmutableMeasurement
+    """
+    # The framework ignores measurements from SKIP and REPEAT phases
+    ignore_outcomes = set([test_record.PhaseOutcome.SKIP,])
```
grybmadsci: `{test_record.PhaseOutcome.SKIP}` i think works?  also maybe make this IGNORED_OUTCOMES, though not sure you need a set() since it only has one thing in it now?

openhtf/core/test_state.py
```diff
@@ -130,7 +146,53 @@ def test_api(self):
...
+                self.get_attachment,
+                self.notify_update,))
+
+  def get_measurement(self, measurement_name):
+    """Get a copy of a measurement value from current or previous phase.
+
+    Args:
+      measurement_name: str of the measurement name
+    Returns:
+      an ImmutableMeasurement
```
grybmadsci: update returns docstring to indicate the None return on missing measurement

openhtf/core/test_state.py
```diff
@@ -130,7 +146,53 @@ def test_api(self):
...
+      if (phase_record.result not in ignore_outcomes and
+          measurement_name in phase_record.measurements):
+        measurement = phase_record.measurements[measurement_name]
+        return ImmutableMeasurement.FromMeasurement(measurement)
+
+    test.logger.warning('Could not find measurement: %s', measurement_name)
+    return None
+
+  def get_attachment(self, attachment_name):
+    """Get a copy of an attachment from current or previous phase."""
```
grybmadsci: is get_attachment() necessary, there's already access to a .attachments attribute?   if it's necessary, can you add some better documentation here? and it should function more closely to get_measurement(), ie, it should iterate over phases in reverse order, not the opposite order of get_measurement().

openhtf/core/test_state.py
```diff
@@ -130,7 +146,53 @@ def test_api(self):
...
+    for phase_record in self.test_record.phases:
+      if attachment_name in phase_record.attachments:
+        attachment = phase_record.attachments[attachment_name]
+        return copy.deepcopy(attachment)
```
grybmadsci: attachments are by-design potentially quite large, i'm not sure the implicit deepcopy in what is otherwise seemingly a simple getter method is particularly user-friendly.  if there isn't a concrete need for get_attachment() i'd remove it.

openhtf/core/test_state.py
```diff
@@ -130,7 +146,53 @@ def test_api(self):
                 running_phase_state.attachments,
                 running_phase_state.attach,
                 running_phase_state.attach_from_file,
-                self.notify_update))
+                self.get_measurement,
+                self.get_attachment,
+                self.notify_update,))
+
+  def get_measurement(self, measurement_name):
+    """Get a copy of a measurement value from current or previous phase.
```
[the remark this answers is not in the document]
kdsudac: Done

openhtf/core/test_state.py
kdsudac: Done

openhtf/core/test_state.py
kdsudac: Done.  Would rather keep variable local since it only make sense in this context and not at the level.  i.e test_state.IGNORED_OUTCOMES would imply that they are ignored at test_state level.

openhtf/core/test_state.py
kdsudac: From documentation on TestApi.measurements: "Only attachments that have been attached in the current phase show up here, and this attribute should not be modified directly;"

Added better documentation on this method.

No need to iterate through in reversed order, because attachment names are unique.  I did swap order so the running_phase_state it checked first (as it is in get_measurement).  I think of get_measurement as the special case and clarified comment to explain reversed.

Also swapped order of get_attachment and get_measurement.

openhtf/core/test_state.py
kdsudac: So had a long discussion with Farz about this in this PR conversation.  Technically you can attach a mutable data structure (like a dict) and I'm trying to prevent users from mutating the attachment.

i.e. a test author shouldn't do something like this:

```
# in PhaseOne
test.attach('my_data', {data: [1, 2, 3] })

# in PhaseTwo
extra_data = [4, 5, 6]
attachment.get_attachment('my_data')
attachment.contents.update({'extra_data': extra_data})

```

kdsudac:
@grybmadsci made requested changes, PTAL.

P.S.  Not sure with github's review system if there's a way to signal that all changes have been made.

kdsudac:
@grybmadsci PTAL, made requested changes.

grybmadsci:
LGTM, sorry for the delay, I seem to have borked my github settings and no longer get an email notification when someone @'s me...

PR #629, opened 2017-07-28
Consume CLI flags when parsing them.

jettisonjoe:
Proposed solution to enable using 3rd party libraries that blow up when encountering unknown command line arguments.

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -328,9 +328,10 @@ def create_arg_parser(add_help=False):
           'My args title', parents=[openhtf.create_arg_parser()])
   >>> parser.parse_args()
   """
-  return argparse.ArgumentParser('OpenHTF-based testing', parents=[
+  return util.argv.SysArgvConsumingParser('OpenHTF-based testing', parents=[
       conf.ARG_PARSER, phase_executor.ARG_PARSER, logs.ARG_PARSER],
-      add_help=add_help)
+      add_help=add_help,
+      consume=True)
 
 
 # Result of a phase.
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -319,7 +319,7 @@ class TestDescriptor(collections.namedtuple(
     return {plug.cls for phase in self.phases for plug in phase.plugs}
 
 
-def create_arg_parser(add_help=False):
+def create_arg_parser(add_help=False, consume=True):
   """Creates an argparse.ArgumentParser for parsing command line flags.
 
   If you want to add arguments, create your own with this as a parent:
@@ -331,7 +331,7 @@ def create_arg_parser(add_help=False):
   return util.argv.SysArgvConsumingParser('OpenHTF-based testing', parents=[
       conf.ARG_PARSER, phase_executor.ARG_PARSER, logs.ARG_PARSER],
       add_help=add_help,
-      consume=True)
+      consume=consume)
 
 
 # Result of a phase.
--- a/openhtf/util/argv.py
+++ b/openhtf/util/argv.py
@@ -3,7 +3,7 @@
 StoreInModule:
   Enables emulating a gflags-esque API (flag affects global value), but one
   doesn't necessarily need to use flags to set values.
-  
+
   Example usage:
     DEFAULT_VALUE = 0
     ARG_PARSER = argv.ModuleParser()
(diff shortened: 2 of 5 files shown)


fahhem:
Is this to deal with gflags? The way I'd recommend with that it is to pass the arguments you want to `gflags.Flags()`. Instead of `Flags(sys.argv)`, pass the `unknown` variable you get from `parse_known_args()`

grybmadsci:
I like the idea, but it may be better to make this an optional thing rather than the default.  for example, the way this broke the CI unit test is fairly likely to happen, i think...

jettisonjoe:
@fahhem, this is to deal with the general case of coexisting with other frameworks and libraries that also take command-line args. In my case it's Mobly. We have a parallel initiative in progress to patch Mobly so it doesn't blow up on unrecognized flags, but this would enable OpenHTF to hide its flags from other badly-behaving consumers of command line args.

@grybmadsci, Fair point, I'll work up an optional version and see how that looks.

jettisonjoe:
Updated to make consuming happen conditionally on an argument to the consuming parser's constructor, but had to do additional work on the unit test to make that pass. Turns out the sys.argv mods that test was doing weren't working right, as other tests getting imported first would also import the conf module, and so the conf import after the sys.argv mod in conf_test.py was a no-op. PTAL.

openhtf/__init__.py
```diff
@@ -328,9 +328,10 @@ def create_arg_parser(add_help=False):
           'My args title', parents=[openhtf.create_arg_parser()])
   >>> parser.parse_args()
   """
-  return argparse.ArgumentParser('OpenHTF-based testing', parents=[
+  return util.argv.SysArgvConsumingParser('OpenHTF-based testing', parents=[
       conf.ARG_PARSER, phase_executor.ARG_PARSER, logs.ARG_PARSER],
-      add_help=add_help)
+      add_help=add_help,
+      consume=True)
```
grybmadsci: maybe take consume to create_arg_parser() and default it to True there?  this still isn't especially optional if the optionality isn't propagated to the external API.

grybmadsci:
just the one comment, otherwise LGTM

jettisonjoe:
Broke it out further to create_arg_parser. Would maybe be nice to have it _actually_ externally settable via `configure()` but there's not a great way to do that since it's necessarily not part of the TestOptions interface. We could make it an additional magic kwarg that `configure()` takes, or just leave it where I have it for now.

jettisonjoe:
PTAL

openhtf/__init__.py
```diff
@@ -319,7 +319,7 @@ def plug_types(self):
     return {plug.cls for phase in self.phases for plug in phase.plugs}
 
 
-def create_arg_parser(add_help=False):
+def create_arg_parser(add_help=False, consume=True):
```
fahhem: I thought we agreed it wouldn't be the default?

openhtf/util/argv.py
```diff
@@ -17,10 +17,39 @@
...
+  """ArgumentParser variant that can consume known arguments from sys.argv.
+
+  Args (in addition to the normal argparse.ArgumentParser args):
+    consume: If True, consumes known args from sys.argv upon parsing.
+  """
+  def __init__(self, *args, **kwargs):
+    self._consume = kwargs.pop('consume')
+    super(SysArgvConsumingParser, self).__init__(*args, **kwargs)
+
+  def parse_args(self):
```
fahhem: why are you overriding this? Also, it doesn't return anything, so you're just breaking the API

openhtf/__init__.py
```diff
@@ -328,9 +328,10 @@ def create_arg_parser(add_help=False):
           'My args title', parents=[openhtf.create_arg_parser()])
   >>> parser.parse_args()
   """
-  return argparse.ArgumentParser('OpenHTF-based testing', parents=[
+  return util.argv.SysArgvConsumingParser('OpenHTF-based testing', parents=[
```
fahhem: why not choose the class based on whether consume is true?

setup.py
```diff
@@ -160,7 +160,7 @@ def run_tests(self):
 
 setup(
     name='openhtf',
-    version='1.1.0',
+    version='1.1.1',
```
fahhem: are we supposed to update this in every commit?

jettisonjoe:
The offending other library accepted our fix to their arg parsing code, so this is no longer an issue. Retiring this PR for now.

PR #626, opened 2017-07-28
Add missing color for ABORTED outcome in ConsoleSummary.

DanLipsitt:

PR #637, opened 2017-08-08
Populate outcome_details when test_start fails.

### LINKED ISSUE #635: outcome_details not set for aborted phase run from test_start
Here's some simplified code that aborts inside a phase run from `test_start`:

```python
import openhtf as htf
from openhtf.output.callbacks import console_summary


@htf.PhaseOptions(repeat_limit=1)
def repeat(test):
    test.test_record.dut_id = '1'
    return htf.PhaseResult.REPEAT

if __name__ == '__main__':
    test = htf.Test()
    test.add_output_callbacks(console_summary.ConsoleSummary())
    test.execute(test_start=repeat)
```

In real life, this would show up if we are trying to get a DUT ID from user input and repeat if it's invalid.

The code triggers a bug when it gets to `ConsoleSummary` (assuming you're running with a version of `ConsoleSummary` that can handle `ABORT`—see #626):

```
2017-08-07 15:51:04,454 - ERROR - Phase returned REPEAT, exceeding repeat_limit.
Finishing test with outcome ABORTED.
Finishing test execution early due to test abortion, outcome ABORTED.
2017-08-07 15:51:04,457 - ERROR - Output callback <openhtf.output.callbacks.console_summary.ConsoleSummary instance at 0x1026d0a28> raised; continuing anyway
  File "/Users/dan/src/shaper/openhtf/openhtf/__init__.py", line 271, in execute
    output_cb(final_state.test_record)
  File "/Users/dan/src/shaper/openhtf/openhtf/output/callbacks/console_summary.py", line 70, in __call__
    (self.indent, record.outcome_details[0].code))
IndexError: list index out of range
```

I have a patch for `ConsoleSummary` that prevents this error, but before I submit it I want to ask whether it makes more sense to fix the underlying issue, which is that `record.outcome_details` isn't being populated.

DanLipsitt:
Make sure `outcome_details` gets populated from a `test_start` phase, even if it aborts.

I don't know how to unit test this, but here's an end-to-end test:
```python
import openhtf as htf

@htf.PhaseOptions()
def phase(test):
    test.test_record.dut_id = '1'
    return htf.PhaseResult.STOP

def check_outcome(record):
    assert(record.outcome_details)

if __name__ == '__main__':
    test = htf.Test()
    test.add_output_callbacks(check_outcome)
    test.execute(test_start=phase)
```

Depends on #626. Fixes #635.

### DIFF
--- a/openhtf/core/test_executor.py
+++ b/openhtf/core/test_executor.py
@@ -161,6 +161,8 @@ class TestExecutor(threads.KillableThread):
     self.test_state.plug_manager.initialize_plugs(
         (phase_plug.cls for phase_plug in self._test_start.plugs))
     outcome = phase_exec.execute_phase(self._test_start)
+    if outcome.is_terminal:
+      self.test_state.finalize_from_phase_outcome(outcome)
 
     if self.test_state.test_record.dut_id is None:
       _LOG.warning('Start trigger did not set DUT ID. A later phase will need'


DanLipsitt:
PTAL.

PR #641, opened 2017-08-21
Add autogenerated api docs.

DanLipsitt:
This is a first crack at generating api docs from docstrings using Sphinx. I've decided to use [PBR](https://docs.openstack.org/pbr/latest/) for this. I've seen it used in various other projects recently, and it seems to work well. The alternative is to use `sphinx-autogen`, but it isn't as automatic and the output is a bit messy. There are other useful things it can do as well, but I haven't configured that.

Things that this PR doesn't address:
- [ ]  Prose docs about how to use OpenHTF


DanLipsitt:
This is currently interfering with `setup.py build_proto`. Not clear why yet.

fahhem:
This should be compatible with readthedocs, but the openhtf project is
owned by maddychan so I can't point it at this branch temporarily to see it
working.

jettisonjoe:
@fahhem, Maddy just added us as maintainers to the OpenHTF readthedocs project.

PR #645, opened 2017-08-24
Contribute plug_dependencies.py

arsharma1:
plug_dependencies.py provides dependencies between plugs.  This uses
descriptors that are initialized by a special phase that must scan over
all the existing phases and must be run before the instantiated plugs
are actually used.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/645)

### DIFF
--- a/contrib/plugs/plug_dependencies.py
+++ b/contrib/plugs/plug_dependencies.py
@@ -1,0 +1,137 @@
+"""Extension to OpenHTF to provide dependency declaration for Plugs.
+
+To declare a dependency, add an attribute to the dependent class with the
+PlugDependency descriptor:
+  class NeededPlug(plugs.BasePlug):
+    pass
+
+  class AnotherPlug(plugs.BasePlug):
+    needed = PlugDependency(NeededPlug)
+
+The depended upon plug is then available after the initialization takes place,
+which is done by the phase created by create_dependency_init_phase.
+
+The dependent plugs must not expect their dependencies to work during their
+__init__ calls.
+
+This makes no guarantee about when plugs are torn down; therefore, dependent
+plugs must not use their dependencies during their own tear down.
+"""
+
+import inspect
+
+import openhtf as htf
+from openhtf import plugs
+
+
+class PlugDependency(object):
+  """Descriptor that declares a dependency to another plug."""
+
+  def __init__(self, plug_type):
+    assert issubclass(plug_type, plugs.BasePlug), (
+        '%s must be a subclass of plugs.BasePlug' % plug_type)
+    self.plug_type = plug_type
+    self._attr = '__dependent_plug.%s' % self.plug_type.__name__
+
+  def __get__(self, instance, owner):
+    # If called on the class, return myself for enumeration.
+    if instance is None:
+      return self
+    return getattr(instance, self._attr, None)
+
+  def __set__(self, instance, value):
+    setattr(instance, self._attr, value)
+
+
+def _enumerate_dependencies(plug_type):
+  for attr_name, attr in inspect.getmembers(plug_type,
+                                            inspect.isdatadescriptor):
+    if isinstance(attr, PlugDependency):
+      yield attr_name, attr.plug_type
+
+
+class _PlugDependencyData(object):
+  """Handler object for evaluating plug dependencies."""
+
+  def __init__(self, plug_types):
+    self.plugs_with_dependencies = set()
+    self.dependent_plugs = set()
+    self.all_plugs = self._find_all_dependent_plugs(plug_types)
+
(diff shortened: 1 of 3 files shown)


pylint_plugins/validate_plugin.py
```diff
@@ -0,0 +1,18 @@
...
+from pylint import lint
+
+
+def __init__(self):
+  pass
+
+
+def validate_transform(cls):
+  if cls.name == 'openhtf.util.validators':
+    # Put all the attrs in Validators into the openhtf.util.validators node.
```
fahhem: This isn't necessary as I've removed this behavior of the validators module

contrib/plugs/plug_dependencies.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def _enumerate_dependencies(plug_type):
+  for attr_name, attr in inspect.getmembers(plug_type,
+                                            inspect.isdatadescriptor):
+    if isinstance(attr, PlugDependency):
+      yield attr_name, attr.plug_type
+
+
+class _PlugDependencyData(object):
```
fahhem: Hmm, the only changes I'd make before putting it into openhtf proper is:

1) Replace the descriptors concept with a single attribute, maybe `dependencies` or `requires`?
2) Use C3 linearization (which is actually how Python's MRO is calculated): https://en.wikipedia.org/wiki/C3_linearization
3) Obviously have the `PlugManager` handle it

But those aren't necessary for this to get into contrib/

contrib/plugs/plug_dependencies.py
```diff
@@ -0,0 +1,137 @@
...
+    needed = PlugDependency(NeededPlug)
+
+The depended upon plug is then available after the initialization takes place,
+which is done by the phase created by create_dependency_init_phase.
+
+The dependent plugs must not expect their dependencies to work during their
+__init__ calls.
+
+This makes no guarantee about when plugs are torn down; therefore, dependent
+plugs must not use their dependencies during their own tear down.
```
kdsudac: Not sure if it's just me, but I find the use of depend* in naming and docstring to be really confusing. i.e. "depended upon", "dependent plugs", "their dependencies".. I get lost in the grammar and have to think too much about the relationship.  
Would it make sense to change naming to use a more concrete relationship like parent/child?

arsharma1: I'm leaning towards Farz's suggestion of a dependency attribute, which would then set attributes at initialization time.

contrib/plugs/plug_dependencies.py
arsharma1: (2) This seems to try to get an order to run the initialization in, but here we need to worry about that.  Instead, we just want to get new plugs that are only referenced as plug dependencies.

Unless you mean that each plug should actually be initialized in the specified order.  That could could provide a way to set the attributes before __init__ like the logger attribute.

contrib/plugs/plug_dependencies.py
```diff
@@ -0,0 +1,137 @@
...
+      return self
+    return getattr(instance, self._attr, None)
+
+  def __set__(self, instance, value):
+    setattr(instance, self._attr, value)
+
+
+def _enumerate_dependencies(plug_type):
+  for attr_name, attr in inspect.getmembers(plug_type,
+                                            inspect.isdatadescriptor):
```
jettisonjoe: I think I would prefer to have a plugs deps listed in something like a plug metadata attribute with a known name rather than relying on introspection to pick through and find the stuff that happen to be PlugDependency instances.

contrib/plugs/plug_dependencies.py
jettisonjoe: I would add beefing up the docstrings to the list of requests for non-contrib inclusion.

test/contrib/plugs/plug_dependencies_test.py
```diff
@@ -0,0 +1,77 @@
+"""Tests for contrib.plugs.plug_dependencies."""
```
jettisonjoe: I'm torn on whether to have tests for contrib stuff live in the tests directory (one of several problems I have with contrib's existence overall). The contrib.* import path is only valid if you're sitting in the top of the repo source, which granted you probably are if you're running unit tests, but it's misleading since contrib itself isn't actually a package.

contrib/plugs/plug_dependencies.py
arsharma1: That's the plan!

test/contrib/plugs/plug_dependencies_test.py
arsharma1: I'm not going to commit this CL as is.  The refactor will be implemented directly into the Plug Manager with tests in the appropriate location.

Issue #647, opened 2017-09-05
Test record logging problems


DanLipsitt:
We want to be able to 
* log to the console through a handler of our choice ([colorlog](https://github.com/borntyping/python-colorlog))
* log to `log_records` in our test output
* set test record loglevels to INFO or DEBUG as needed
* not have duplicate logging to the console

I'm willing to hack up `openhtf.logs.util` to get it done, but I can't for the life of me seem to get all of these things to happen at once. Console logging and record output logging seem to be tied together, but I can't tell where. 

I've seen the note that says "Test record logs are by default output to stdout at a debug level.  There is no way to change this." Why?

DanLipsitt:
Some things that make logging configuration hard:
* `propagate` is set to `False` in the main logger: https://github.com/google/openhtf/blob/6e05dbedb28f8a2739fd565407457f606de91741/openhtf/util/logs.py#L203
* The `record_logger` deletes existing handlers: https://github.com/google/openhtf/blob/6e05dbedb28f8a2739fd565407457f606de91741/openhtf/util/logs.py#L120-L121

Note that it's probably necessary to be able to add handlers before `Test`s are created or `Plug` initializers (for example) won't be able to log correctly.

jettisonjoe:
Now all handlers are on the top-level 'openhtf' logger, so you should be able to go in and change them as long as you do it after you call `configure_cli_logging()`.

PR #648, opened 2017-09-05
Don't force DEBUG loglevel in test_record loggers.

DanLipsitt:
Allow users to set log level for `test_record` and its children.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/648)

### DIFF
--- a/openhtf/util/logs.py
+++ b/openhtf/util/logs.py
@@ -116,7 +116,6 @@ def initialize_record_logger(test_uid, test_record, notify_update):
   # All record loggers have a shared parent that's separately configured, so
   # we want to propagate to that logger.
   logger.propagate = True
-  logger.setLevel(logging.DEBUG)
   # Just in case, make sure we don't have any extra handlers hanging around.
   logger.handlers = [RecordHandler(test_record, notify_update)]
   return logger


DanLipsitt:
Helps with #647.

DanLipsitt:
PTAL.

DanLipsitt:
@fahhem any thoughts?

jettisonjoe:
I would rather approach this by separating the console logging from the test record logging, and still default test record logging to DEBUG. I can't think of a compelling reason why I _wouldn't_ want DEBUG level logs in my test record; we usually filter those logs in whatever tool we're using to display test records.

DanLipsitt:
@jettisonjoe one reason is the [16MB limit for MongoDB records](https://docs.mongodb.com/manual/reference/limits/#BSON-Document-Size). Although 16MB of logs sounds like a lot, there's no guarantee that users won't exceed it.

jettisonjoe:
Hmm. That's a fair point. We should at least wire this out to Test.configure() or similarly allow test script authors to control it, though.

dbhatman:
logs.py was refactored in https://github.com/google/openhtf/pull/799 so these changes are no longer relevant but even as is we should make this more clearly configurable so they are not breaking changes for users.

PR #651, opened 2017-09-12
Fix unicode logs

DanLipsitt:
Tested against the following code:

```python
# coding=utf-8
import openhtf
from openhtf.output.callbacks.json_factory import OutputToJSON
import sys
from openhtf.core.measurements import Measurement


@openhtf.measures(Measurement('val'))
def utf8(test):
    val = u'☹'
    test.logger.error('uh oh')
    test.logger.error(val)
    test.measurements.val = val
    raise Exception(val)

if __name__ == '__main__':
    test = openhtf.Test(utf8)
    test.add_output_callbacks(
        OutputToJSON('log_demo.json')
    )
    test.execute(test_start=lambda: 'fake_dut')
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/651)

### DIFF
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -23,6 +23,7 @@ prompt state should use the openhtf.prompts pseudomodule.
 
 import collections
 import functools
+import locale
 import logging
 import os
 import platform
@@ -92,7 +93,10 @@ class ConsolePrompt(threading.Thread):
         if sys.stdin.isatty():
           termios.tcflush(sys.stdin, termios.TCIFLUSH)
 
-        line = ''
+        # Although this isn't threadsafe with do_setlocale=True, it doesn't work without it.
+        encoding = locale.getpreferredencoding(do_setlocale=True)
+
+        line = u''
         while not self._stopped:
           inputs, _, _ = select.select([sys.stdin], [], [], 0.001)
           for stream in inputs:
@@ -111,7 +115,7 @@ class ConsolePrompt(threading.Thread):
                   # want to actually quit.
                   print "Hit ^C (Ctrl+c) to exit."
                   break
-              line += new
+              line += new.decode(encoding)
               if '\n' in line:
                 response = line[:line.find('\n')]
                 self._answered = True
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -143,7 +143,7 @@ class UserInput(plugs.FrontendAwareBasePlug):
   def _create_prompt(self, message, text_input):
     """Sets the prompt."""
     prompt_id = uuid.uuid4()
-    _LOG.debug('Displaying prompt (%s): "%s"%s', prompt_id, message,
+    _LOG.debug(u'Displaying prompt (%s): "%s"%s', prompt_id, message,
                ', Expects text' if text_input else '')
 
     self._response = None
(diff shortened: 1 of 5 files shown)


DanLipsitt:
One test failure is because I converted a complex object into unicode, which should be OK I think, and the test should be changed. The other is a weird timing thing that I think might be a fluke.

fahhem:
Then please change the test to make it pass

openhtf/util/data.py
```diff
@@ -158,7 +158,7 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
     return float(obj)
 
   # Convert all other types to strings.
-  return str(obj)
+  return unicode(obj)
```
fahhem: Do we need to convert to unicode? There are cases where the data is actually bytes (serial numbers, measurement values, etc) or doing this potentially expensive conversion is unwanted. Can we split this into a 3-way switch? If it's a str, return it, if it's unicode, return that, and then actually return unicode(obj)?

DanLipsitt: I don't think I understand the code well enough right now to answer that, but what you say makes sense.

fahhem: Did you mean to change this later then?

DanLipsitt: If I get the time, yes.

DanLipsitt: It turns out unicode strings are already skipped because they're listed in `PASSTHROUGH_TYPES`. I've rolled this part back.

DanLipsitt:
PTAL.
This PR has few changes but now contains a bunch of commits, so I think it's a good candidate for squash merging. Let me know if you want me to do something else to clean it up.

DanLipsitt:
Oh hold on, looks like this doesn't fix everything yet.

DanLipsitt:
OK, fixed another bug in UserInput logging. I also had to add charset decoding for console input. That doesn't match the the original title of the PR but it's necessary for fixing the unicode exceptions, so I can change the PR title if you want 😉.

Getting the system locale [isn't thread safe](https://docs.python.org/2/library/locale.html#locale.getpreferredencoding) but I don't think there's much risk of a race condition. I'm actually not clear on whether multiple ConsoleInput threads can run at one time anyway so I'm interested to hear your thoughts.

openhtf/plugs/user_input.py
```diff
@@ -92,7 +93,10 @@ def run(self):
         if sys.stdin.isatty():
           termios.tcflush(sys.stdin, termios.TCIFLUSH)
 
-        line = ''
+        # Although this isn't threadsafe with do_setlocale=True, it doesn't work without it.
+        encoding = locale.getpreferredencoding(do_setlocale=True)
```
jettisonjoe: Since this isn't threadsafe, and ConsolePrompt is a thread, maybe we can do it once a the module level during import and store the result in an `ENCODING` module global (locale is conceptually global state anyway)? That is probably safer than running it here in a thread.

dbhatman:
Joe had a suggestion based on the race condition but I think more relevantly -- now that we are python3 only I suspect these issues are self resolved (ie python3's str supports unicode directly so no more mangling of types is necessary)! Going to close this out. If you still see this as an issue please consider implementing Joe's ideas and updating the PR to be based at the current master.

PR #652, opened 2017-09-18
Give frontends access to plug inheritance info

Kenadia:
This PR adds the concept of a plug descriptor. Currently this contains one item of info, the MRO, which here is a list of class names, like `['openhtf.plugs.user_input.UserInput', 'my_module.advanced_user_input.AdvancedUserInput']`.

This allows frontends to make use of base class information to render subclassed plugs appropriately. For example, a frontend might use the same frontend component to render both the standard `openhtf.plugs.user_input.UserInput` and a custom `AdvancedUserInput`, leaving the test writer with the option of using either plug for operator input.

### DIFF
--- a/openhtf/plugs/__init__.py
+++ b/openhtf/plugs/__init__.py
@@ -426,9 +426,10 @@ class PlugManager(object):
     self._plug_types.add(plug_type)
     if plug_type in self._plugs_by_type:
       self._plugs_by_type[plug_type].tearDown()
+    plug_name = self.get_plug_name(plug_type)
     self._plugs_by_type[plug_type] = plug_value
-    self._plugs_by_name[self.get_plug_name(plug_type)] = plug_value
-    self._plug_descriptors[plug_type] = self._make_plug_descriptor(plug_type)
+    self._plugs_by_name[plug_name] = plug_value
+    self._plug_descriptors[plug_name] = self._make_plug_descriptor(plug_type)
 
   def provide_plugs(self, plug_name_map):
     """Provide the requested plugs [(name, type),] as {name: plug instance}."""
--- a/openhtf/plugs/__init__.py
+++ b/openhtf/plugs/__init__.py
@@ -338,10 +338,12 @@ class PlugManager(object):
         ['openhtf.plugs.user_input.UserInput',
          'my_module.advanced_user_input.AdvancedUserInput']
     """
-    return ','.join([
+    ignored_classes = (BasePlug, FrontendAwareBasePlug)
+    return [
         self.get_plug_name(base_class) for base_class in plug_type.mro()
-        if issubclass(base_class, BasePlug) and base_class is not BasePlug
-    ])
+        if (issubclass(base_class, BasePlug) and
+            base_class not in ignored_classes)
+    ]
 
   def get_plug_name(self, plug_type):
     """Returns the plug's name, which is the class name and module.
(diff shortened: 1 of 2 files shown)


Kenadia:
Responded to feedback from @fahhem. Instead of putting MRO info in the “plug name” we now put it in the “plug descriptor”

openhtf/plugs/__init__.py
```diff
@@ -116,6 +116,15 @@ def __init__(self, my_config_key)
...
 _LOG = logging.getLogger(__name__)
 
+PlugDescriptor = collections.namedtuple('PlugDescriptor', ['mro'])  # pylint: disable=invalid-name
+
+# Placeholder for a specific plug to be provided before test execution.
+#
+# Use the with_plugs() method to provide the plug before test execution. The
+# with_plugs() method checks to make sure the substitute plug is a subclass of
+# the PlugPlaceholder's base_class.
+PlugPlaceholder = collections.namedtuple('PlugPlaceholder', ['base_class'])  # pylint: disable-invalid-name
```
fahhem: why did you need to move the docstring into a comment?

openhtf/plugs/__init__.py
```diff
@@ -151,24 +150,53 @@ class BasePlug(object):
...
   logger = _LOG
 
   @classproperty
   def placeholder(cls):
-    """Returns a PlugPlaceholder for calling class."""
+    """Returns a PlugPlaceholder for the calling class."""
     return PlugPlaceholder(cls)
 
+  @classmethod
+  def get_descriptor(cls):
```
fahhem: not sure I want this exposed to plug implementors? We limit the attributes we set here to ones that make sense for implementors to set/override (except for `placeholder` which has to be here). But getting the name, mro, and descriptor should be up to the `PlugManager` since we don't want to have hard-to-debug failures of the frontend because someone decided to implement `get_descriptor()` or `get_name()` as normal plug functions

Kenadia: Done

openhtf/plugs/__init__.py
Kenadia: Only since it seemed weird to have two different styles in use:
```python
PlugDescriptor = collections.namedtuple(...)
```
and
```python
class PlugPlaceholder(collections.namedtuple(...)):
   ...
```
And since the former seems to be the dominant style, except when something else is added to the class.

The class-style was nice for the long docstring so I could see this going either way, but I opted for what I thought was more consistent. Happy to revert this if you have an opinion about it.

openhtf/plugs/__init__.py
```diff
@@ -273,10 +272,14 @@ def __init__(self, plug_types=None, logger=None):
     self._xmlrpc_server = None
 
   def _asdict(self):
-    return {'plug_states': {name: plug._asdict()
-                            for name, plug in self._plugs_by_name.iteritems()},
-            'xmlrpc_port': self._xmlrpc_server and
-                           self._xmlrpc_server.socket.getsockname()[1]}
+    return {
+        'plug_descriptors': {name: self.get_plug_descriptor(plug)
```
fahhem: getting the plug descriptor can be expensive (even now, mro traversal isn't cached), do we want to cache it in a self._plug_descriptors?

Kenadia: Hm, if plug descriptors don't change, then we can calculate plug_descriptors in `initialize_plugs()`.

PR #658, opened 2017-10-13
timeouts: Add a loop_until_true_or_raise utility.

jettisonjoe:


wallacbe:
Other methods have timeout_s as required parameter and sleep as an optional parameter as default value of 1 second.  Any the values in the new method arbitrary?

jettisonjoe:
@wallacbe I updated the arg names and defaults to more closely match the other similar functions, and the name to avoid ambiguous senses of "or".

PR #659, opened 2017-10-18
Neg value handling for within_percent validator.

rcfie:
The WithinPercent validator did not sufficiently handle the case of negative values (as might be encountered when measuring voltages).  This change uses the absolute value of expected and tested values to make accommodation for those cases.

### DIFF
--- a/openhtf/util/validators.py
+++ b/openhtf/util/validators.py
@@ -239,12 +239,16 @@ class WithinPercent(RangeValidatorBase):
     self.percent = percent
 
   @property
+  def _applied_percent(self):
+    return abs(self.expected * self.percent / 100.0)
+
+  @property
   def minimum(self):
-    return (1.0 - self.percent / 100.0) * self.expected
+    return self.expected - self._applied_percent
 
   @property
   def maximum(self):
-    return (1.0 + self.percent / 100.0) * self.expected
+    return self.expected + self._applied_percent
 
   def __call__(self, value):
     return self.minimum <= value <= self.maximum


openhtf/util/validators.py
```diff
@@ -235,7 +235,7 @@ class WithinPercent(RangeValidatorBase):
   def __init__(self, expected, percent):
     if percent < 0:
       raise ValueError('percent argument is {}, must be >0'.format(percent))
-    self.expected = expected
+    self.expected = abs(expected)
```
arsharma1: Don't use the absolute value of the expected like this.  It would accept the wrongly signed value.

Instead, just absolute value to the applied percent and then add/subtract that as indicated.

`@property`
`def _applied_percent(self):`
`  return abs(self.expected * self.percent / 100.0)`
` `
`@property`
`def minimum(self):`
`  return self.expected - self._applied_percent`
` `
`@property`
`def maximum(self):`
`  return self.expected + self._applied_percent`

rcfie: ahh, you're right.

fahhem:
Can you provide an example input where it didn't work before and it works now?

rcfie:
Let's say we have
Expected = -10
Percent = 10
Measurement = -9.5
 then the final equality would evaluate to:
-9 <= -9.5 <= -11

This fails because -9 > -9.5

The problem was in how min and max were defined.

fahhem:
Ah, I see, the min/max got swapped essentially when expected is negative. Good catch!

Issue #660, opened 2017-10-19
Linux contributing instruction issues


wallacbe:
git clone https://github.com/google/openhtf.git
instead of
git clone git@github.com:google/openhtf.git

Remove space after backslash to allow copy/paste into CLI
```
sudo apt-get install python-pip swig libssl-dev python-dev libffi-dev \
protobuf-compiler libprotobuf-dev

```
Issue with installing cryptography  (also see issue in pypi)
Installed /tmp/easy_install-q0NaH_/cryptography-2.1.1/pycparser-2.18-py2.7.egg
error: Setup script exited with error in cryptography setup command: Invalid environment marker: python_version < '3'

Appears to be issue with cryptography 2.1.1, cryptography 1.9 is fine

wallacbe:
Running "pip install setuptools --upgrade" after activating virtualenv fixes cryptography issue

PR #661, opened 2017-10-24
CONTRIBUTING.md updates

wallacbe:
https://github.com/google/openhtf/issues/660

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -210,10 +210,10 @@ accordingly.
 
 ```bash
 # Clone into the repo.
-git clone git@github.com:google/openhtf.git
+git clone https://github.com/google/openhtf.git
 
 # Install system-level third-party dependencies.
-sudo apt-get install python-pip swig libssl-dev python-dev libffi-dev \ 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -210,10 +210,10 @@ accordingly.
 
 ```bash
 # Clone into the repo.
-git clone git@github.com:google/openhtf.git
+git clone https://github.com/google/openhtf.git
```
jettisonjoe: Curious what the motivation is for changing the recommended way from ssh to https?

wallacbe: I get an error when using ssh:
Permission denied (publickey).
fatal: Could not read from remote repository.

Which is probably fixable.  

The github page specifies  "Clone with HTTPS" as the default option which works, so I figured this would be the method to get someone up and running quicker.

jettisonjoe:
Fair enough. The ssh method requires you to set up your ssh keys, but enables you to avoid having to put in your credentials all the time once you've set your keys up. GitHub's docs recommend https, so I guess it does make sense to have that as our default in the instructions.

Issue #663, opened 2017-10-25
Add current_phase attribute to test object


wallacbe:
Would be nice to be able access current phase name from test object.  I find myself referencing this to the test logger quite a bit

kdsudac:
Technically you can do this now if you apply a @htf.PhaseOptions(requires_state) to your phase.

Then inside your phase: `test.running_phase_state.name`

I think I'd support adding the running phase name to the TestApi.

PR #664, opened 2017-10-26
add phase name property to TestApi

wallacbe:
https://github.com/google/openhtf/issues/663


openhtf/__init__.py
```diff
@@ -605,7 +607,10 @@ def dut_id(self, dut_id):
                           self.test_record.dut_id, dut_id)
     self.test_record.dut_id = dut_id
     self.notify_update()
-
+  
+  @property
+  def phase(self):
+    return inspect.stack()[1][3]
```
fahhem: definitely do not do this, it will break over time in weird and confusing ways

PR #666, opened 2017-10-30
Typecast CLI Conf Values

maddychan:
- Typecasted CLI conf values so check not needed in tests.
	- Added unit test to test this.

### DIFF
--- a/test/util/conf_test.py
+++ b/test/util/conf_test.py
@@ -21,7 +21,7 @@ sys.argv.extend([
     # You can specify arbitrary keys, but they'll get ignored if they aren't
     # actually declared anywhere (included here to make sure of that).
     '--config_value=undeclared_flag=who_cares',
-    '--config-value', 'true_value=true',
+    '--config-value=true_value=true',
     '--config-value', 'num_value=100',
 ])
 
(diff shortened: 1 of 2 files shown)


test/util/conf_test.py
```diff
@@ -21,6 +21,8 @@
     # You can specify arbitrary keys, but they'll get ignored if they aren't
     # actually declared anywhere (included here to make sure of that).
     '--config_value=undeclared_flag=who_cares',
+    '--config-value', 'true_value=true',
```
fahhem: can you make one of them '--config-value=true_value=true'?

maddychan:
Sorry had to force push, but did the change!

test/util/conf_test.py
maddychan: yups!

fahhem:
You need to fix the conf_test, then you can submit this.

BTW, @wallacbe, you've wanted this for a while and @grybmadsci was pushing back a long time ago. I'm assuming @jettisonjoe is onboard so I think that discussion has been settled in your favor.

test/util/conf_test.py
```diff
@@ -89,10 +93,19 @@ def test_flag_values(self):
...
+    self.assertEquals(True, conf.true_value)
+
   def test_as_dict(self):
     conf.load(station_id='station_id')
     self.assertEquals({
         'flag_key': 'flag_value',
+        'true_value': True,
+        'num_value': 100,
         'cancel_timeout_s': 2,
         'enable_station_discovery': True,
```
maddychan: This key isn't showing up in conf._asdict() and I'm not sure why, and that makes this unit test fail. This fails on master too and I'm not sure what causes it; should I just take it out?

fahhem: looks like you fixed it, since travis is happy?

PR #673, opened 2017-11-21
Add a Frontline bluetooth sniffer plug.

jettisonjoe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/673)


jettisonjoe:
@grybmadsci @Kenadia PTAL

In addition to adding the sniffer plug, this PR shows a proposed vendor-based plug directory layout, and can serve as an example of how to use DeviceWrappingPlug.

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+import logging
+import os
+import socket
+import time
+
+from openhtf.util import files
+
+DEFAULT_AUTOMATION_PORT = 22901
+DEFAULT_HTTP_PORT = 55455
+REMOTE_BIN_PATH = (r'C:\Program Files (x86)\Frontline Test System II'
```
grybmadsci: some docs on these magic values would be good

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    self._socket.sendall(full_command.encode() + b'\n')
+
+    for line in self._socket.makefile():
+      response = line.split(';')
+      _LOG.debug('Sniffer response: %s', line.strip())
+      if command in response and 'SUCCEEDED' in response:
+        return
+      raise SnifferError(full_command, line.strip())
+
+  def _export_tab(self, tab):
```
grybmadsci: docstring

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    self._socket.settimeout(16)
+    try:
+      self._sniffer_command('Stop FTS')
+    except (SnifferError, socket.timeout):
+      pass
+
+    self._socket.settimeout(None)
+
+  def _sniffer_command(self, command, *args):
+    """Send command to the sniffer's automation server and wait for response."""
```
grybmadsci: maybe a little more description about the 'SUCCEEDED' response check/mention SnifferError in the docstring

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Classes for controlling Frontline bluetooth sniffers.
+
+This module provides control over a variety of Frontline (now owned by Teledyne
+Lecroy) bluetooth sniffers. For full details on Frontline's sniffer control
+protocol, which this module partially implements, see the reference document:
```
grybmadsci: doc link is good but a quick summary of the bit implemented here would be pretty handy in the docstring i think.

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    except SnifferError as e:
+      _LOG.warning('Failed to export tab %s: %s', tab, e.response)
+    else:
+      self._files.append(self.export_prefix + ending)
+
+  @property
+  def files(self):
+    """Iterator over the names of log files created by the sniffer."""
+    for path in self._files:
+      yield os.path.basename(path)
```
grybmadsci: i think it's a bit more readable to return a generator expression here:
`return (os.path.basename(path) for path in self._files)`

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+  def export_prefix(self):
+    """Generate a filename prefix specific to a single sniffer run."""
+    return os.path.join(self._log_path, 'sniff_{}_{}'.format(
+        self.capture_start_time.strftime('%Y%m%d-%H%M%S'), self.capture_id))
+
+  def start_sniffing(self,
+                     bluetooth_mac,
+                     capture_id,
+                     binary_path=REMOTE_BIN_PATH):
+    """Configure the sniffer and start sniffing."""
```
grybmadsci: args section?  this is a pretty important public api so probably worth being a bit more thorough on this one docstring

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    self._files.append(self.export_prefix + '.fsc')
+
+    self._export_tab('LE:LE LL')
+    self._export_tab('LE:LE DATA')
+
+    self._sniffer_command('Stop FTS')
+
+    _LOG.info('BLE data exported with prefix: %s', self.export_prefix)
+
+    self.reset()
```
grybmadsci: maybe put the reset() call in a finally in case any of the commands fail

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    self.capture_id = None
+    self.capture_start_time = None
+
+  @contextlib.contextmanager
+  def sniff(self, bluetooth_mac, capture_id):
+    """Start and stop the sniffer through the context manager interface."""
+    self.start_sniffing(bluetooth_mac, capture_id)
+
+    try:
+      yield
```
grybmadsci: maybe yield self?  not sure it would actually be used at all, but it's a common paradigm for a contextmanager like this

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+def _run_bluetooth_sniffer(host, port, local_log_path, download_port, address,
+                           verbose, capture_id, output, seconds):
+  """Run the bluetooth sniffer; for use when module is invoked as runnable.
+
+  Args: All arguments correspond to the matching args defined directly above in
+      the _parse_args function. See help strings there for details.
+  """
+  _init_logging(_LOG, logging.DEBUG if verbose else logging.INFO)
+
+  if not os.path.exists(output):
```
grybmadsci: maybe `os.path.isdir()` instead of exists?

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+    os.makedirs(output)
+
+  sniffer = BleSniffer(host, port)
+  try:
+    sniffer.start_sniffing(address, capture_id)
+    if seconds is None:
```
grybmadsci: since 0 for seconds is a useless value, might be more readable to do `if seconds:` and invert the blocks.  up to you though if you'd prefer to keep the stricter check.

openhtf/plugs/frontline/bluetooth_sniffer.py
```diff
@@ -0,0 +1,277 @@
...
+      _LOG.info('Press ENTER to stop bluetooth sniffing.')
+      input()
+    else:
+      _LOG.info('Sniffing bluetooth for %f seconds.', seconds)
+      time.sleep(seconds)
+  finally:
+    sniffer.stop_sniffing()
```
grybmadsci: right now stop_sniffing() can raise, which would result in the rest of this not happening, which is probably not good?

openhtf/plugs/frontline/bluetooth_sniffer_plug.py
```diff
@@ -0,0 +1,82 @@
...
+  def __init__(self,
+               bluetooth_sniffer_enable,
+               bluetooth_sniffer_host,
+               bluetooth_sniffer_port):
+    # Default to no device is sniffer support is disabled via conf.
+    device = None
+    if bluetooth_sniffer_enable:
+      device = bluetooth_sniffer.Sniffer(
+          bluetooth_sniffer_host, bluetooth_sniffer_port)
+    super(BluetoothSnifferPlug, self).__init__(device)
```
grybmadsci: hm, i actually like ternary expressions in this case, but it's a bit of a personal preference, up to you:
```
super(BluetoothSnifferPlug, self).__init__(
    bluetooth_sniffer.Sniffer(bluetooth_sniffer_host, bluetooth_sniffer_port)
    if bluetooth_sniffer_enable else None)
```

openhtf/plugs/frontline/bluetooth_sniffer_plug.py
```diff
@@ -0,0 +1,82 @@
...
+
+  def is_enabled(self):
+    """True if sniffer support is enabled and sniffer control is set up."""
```
grybmadsci: it's unclear what 'sniffer control is set up' means here

openhtf/util/files.py
```diff
@@ -0,0 +1,71 @@
...
+
+def download(url, fd):
+  """Downloads a URL with HTTP to a file descriptor."""
+  try:
+    request = urllib.urlopen(url)  # Python 2 does not support with blocks
+    shutil.copyfileobj(request, fd)
+  finally:
+    try:
+      request.close()
+    except NameError:
```
grybmadsci: I think you can pull the request = assignment out of the try block and then you don't need to handle the NameError, as it won't hit the finally: block unless request has been successfully assigned to.

openhtf/util/files.py
```diff
@@ -0,0 +1,71 @@
...
+
+def download_file(url, directory):
+  with open(os.path.join(directory, url.rsplit('/', 1)[-1]), 'wb') as fd:
+    download(url, fd)
+    return fd.name
+
+
+def download_files(remote_paths, host, port, directory, logger=_LOG):
+  """Downloads files from http://host:port/path; returns list of local paths."""
+  if not os.path.exists(directory):
```
grybmadsci: prefer `os.path.isdir()` for dirs

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Beefed up the docstring with better description of what's here plus examples.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Done.

openhtf/plugs/frontline/bluetooth_sniffer.py
jettisonjoe: Moved the stop_sniffing() call to inside the try block since exceptions earlier in that block almost guarantee that stop_sniffing() will actually blow up as well.

openhtf/plugs/frontline/bluetooth_sniffer_plug.py
jettisonjoe: I do too, but no matter how much I fiddled with the line breaks the expression always seemed harder to visually parse that way. Part of the problem is that the arg names have to be somewhat long to avoid collisions since they're actually config values injected `@inject_positional_args`.

The version you have there looks nicer, but has a line that ends up being 81 characters. :confounded:

PR #674, opened 2017-11-28
python3 WIP

charlesnicholson:
All of the Python3 migration changes I've made so far. Tests pass but Brandon discovered lots of things that don't work. This PR is just to get my repo changes into the mainline so we can develop publicly there instead of in my github.

(might also need to roll back the pathing changes I made in setup.py around the protobufs)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: python
 python:
-- '2.7'
+  # - '2.7'
+  - '3.6'
 addons:
   apt:
     packages:
+    - python3
+    - python3-pip
     - swig
     - libusb-1.0-0-dev
     - libprotobuf-dev
@@ -14,7 +17,8 @@ cache:
   directories:
   - $HOME/.cache/pip
 env:
-- TOXENV=py27
+  #- TOXENV=py27
+- TOXENV=py36
 install:
 - pip install tox coveralls
 - pip install -e .
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ addons:
     - swig
     - libusb-1.0-0-dev
     - libprotobuf-dev
-    - protobuf-compiler
+      #    - protobuf-compiler
 cache:
   pip: true
   apt: true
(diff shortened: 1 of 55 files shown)


fahhem:
Update the travis yaml to use python 3 for the tests

wallacbe:
Add python3-dev to packages in CONTRIBUTING.md

wallacbe:
Is the a reason why Travis would show and error vs setting up a local virtualenv with Python3.6 and having it pass running:
python setup.py test --pytest-cov=term-missing

charlesnicholson:
Yeah, I noticed that too, I'm having trouble recreating the failure. I'll take a closer look sometime soon.

In the meantime, I'm seeing what looks like some nondeterminism in the tests, Re-running "python3 setup.py test" a few times gave me this failure in test_cancel_start:

>     self.assertLessEqual(record.end_time_millis, record.start_time_millis)
E     AssertionError: 1511967289989 not less than or equal to 1511967289988

Those look like real values, not fake ones? Anyway, I'll look into that as well.

jettisonjoe:
Most of the times I've seen that I've had uncommitted changes on my local repo. You can try clearing Travis's cache from the web GUI, though.

charlesnicholson:
Thanks, seems worth a try. I don't see any obvious configuration that I have access too, am I just being dense or do you need to be a project maintainer to clear the cache?

wallacbe:
I have access to Travis.  I deleted the cache for that PR and started a build (which failed).  What protoc version does Travis use?  I don't see that in the logs

jettisonjoe:
That's a good question; I don't know. Our travis.yml file only specifies the package libprotobuf-dev not a specific version. Now that you mention it, I did have issues compiling the protos recently on a Debian-unstable machine and I think it was due to the new protoc breaking the way we use our protos.

jettisonjoe:
Try changing this line in `setup.py` to require protobuf version `3.2.0` instead.
```
'protobuf>=2.6.1,<4.0'
```

wallacbe:
I searched through Trusty raw logs, looks like protoc 2.5.0 ?
Get:27 http://archive.ubuntu.com/ubuntu trusty/main amd64 protobuf-compiler amd64 2.5.0-9ubuntu1 [19.8 kB]

Looks like Python3 is supported in protoc versions 3+

My local machine is:
protoc --version
libprotoc 3.5.0

Travis has a blurb in their documentation about creating a script and using wget to install protoc
https://docs.travis-ci.com/user/installing-dependencies/

jettisonjoe:
We should consider adding a protoc version check and download to setup.py as an alternative to doing it in a travis script.

charlesnicholson:
Cool, the travis fixes from @wallacbe worked, though they're not what @jettisonjoe recommended. I'm in favor of this PR going in to origin/python3 as-is so that Brandon (and maybe others?) can independently contribute to it rather than open fake PR's for me to copy-paste into this one :)

wallacbe:
LGTM

PR #675, opened 2017-11-28
Fixes to reconnect to the phone and specify a serial number

arsharma1:
Add fixes that allow reconnecting to the underlying ADB device via the plug and allow specifying a serial number with subclasses in order to support multiple devices at one time.

### DIFF
--- a/openhtf/plugs/usb/__init__.py
+++ b/openhtf/plugs/usb/__init__.py
@@ -119,9 +119,24 @@ class AdbPlug(plugs.BasePlug):
   """Plug that provides ADB."""
 
   def __init__(self):
-    kwargs = {}
     if conf.libusb_rsa_key:
-      kwargs['rsa_keys'] = [adb_device.M2CryptoSigner(conf.libusb_rsa_key)]
+      self._rsa_keys = [adb_device.M2CryptoSigner(conf.libusb_rsa_key)]
+    else:
+      self._rsa_keys = None
+    self._device = None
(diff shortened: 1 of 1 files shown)


openhtf/plugs/usb/__init__.py
```diff
@@ -118,21 +118,41 @@ def __getattr__(self, attr):
 class AdbPlug(plugs.BasePlug):
   """Plug that provides ADB."""
 
+  serial_number = None
```
Kenadia: Optional - might be worth adding a bit of documentation (e.g. “Attributes” in the AdbPlug docstring) describing serial_number and mentioning that it is incompatible with conf.remote_usb/conf.ethersync

jettisonjoe:
I'm going to merge this cause it LGTM, but I wanted to mention that in Daydream we're making heavy use of the Mobly AndroidDevice controller as an alternative to a separate "pure" Python ADB implementation. While the approach of shelling out to the CLI ADB may feel kludgey, Mobly's AndroidDevice is actively maintained and seems to have wide usership (through Mobly) at least internally. I would like to upstream our AndroidDevice-based plug (as an extra) and maybe even retire our ADB implementation in favor of it.

One possible objection I see is that this will introduce Mobly as another dependency for folks who want to use ADB, but I believe it's probably worth it in order to have our ADB plug built on an actively used and maintained foundation rather than our own fringe implementation. Thoughs?

PR #677, opened 2017-12-01
python3: updates to allow examples to run in python27 and unit test to pass

wallacbe:


wallacbe:
PTAL

openhtf/util/xmlrpcutil.py
```diff
@@ -16,13 +16,20 @@
...
+import os
 import socketserver
+import sys
 import threading
 import xmlrpc.client
 import collections
 
 DEFAULT_PROXY_TIMEOUT_S = 3
 
+if os.name == 'posix' and sys.version_info[0] < 3:
```
charlesnicholson: `future` takes care of this without having to detect which Python version:

http://python-future.org/compatible_idioms.html#xmlrpc-module

setup.py
```diff
@@ -53,8 +53,12 @@ class BuildProtoCommand(Command):
   def initialize_options(self):
     self.skip_proto = False
     try:
-      prefix = subprocess.getoutput(
-          'pkg-config --variable prefix protobuf').strip()
+      if os.name == 'posix' and sys.version_info[0] < 3:
```
charlesnicholson: Also handled by future. In general, try future before switching on sys.version_info for 3 => 2 compatibility, that's what it's for and it's guaranteed to keep working as Python3 evolves.

http://python-future.org/compatible_idioms.html#commands-subprocess-modules

openhtf/util/xmlrpcutil.py
charlesnicholson: Also I'm not sure what the os.name == 'posix' clause guards against, do these packages not work on windows or something?

wallacbe:
Added travis back in, seeing tests fail similar to running local

wallacbe:
PTAL

test/util/data_test.py
```diff
@@ -45,7 +46,10 @@ class FloatSubclass(float):
     self.assertIs(type(converted['unicode']), str)
     self.assertIs(type(converted['int']), int)
     self.assertIs(type(converted['float']), float)
-    self.assertIs(type(converted['long']), int)
+    if sys.version_info[0] < 3:
```
charlesnicholson: There are a couple of idioms that are preferable to testing the Python version explicitly. See http://python-future.org/compatible_idioms.html#long-integers

pseudocode:
```
from builtins import int
isInstance(converted['long'], int)
```

test/util/validators_test.py
```diff
@@ -10,10 +11,16 @@
 class TestInRange(unittest.TestCase):
 
   def test_raises_if_invalid_arguments(self):
-    with self.assertRaisesRegex(ValueError, 'Must specify minimum'):
-      validators.InRange()
-    with self.assertRaisesRegex(ValueError, 'Minimum cannot be greater'):
-      validators.InRange(minimum=10, maximum=0)
+    if sys.version_info[0] < 3:
```
charlesnicholson: This should use six instead:
https://pythonhosted.org/six/#six.assertRaisesRegex

test/util/validators_test.py
```diff
@@ -108,8 +115,12 @@ class MyType(object):
 class TestWithinPercent(unittest.TestCase):
 
   def test_raises_for_negative_percentage(self):
-    with self.assertRaisesRegex(ValueError, 'percent argument is'):
-      validators.WithinPercent(expected=100, percent=-1)
+    if sys.version_info[0] < 3:
```
charlesnicholson: Ditto here, six fixes this: https://pythonhosted.org/six/#six.assertRaisesRegex

openhtf/util/xmlrpcutil.py
```diff
@@ -16,13 +16,20 @@
...
+import os
 import socketserver
+import sys
 import threading
 import xmlrpc.client
 import collections
 
 DEFAULT_PROXY_TIMEOUT_S = 3
 
+if sys.version_info[0] < 3:
```
charlesnicholson: http://python-future.org/compatible_idioms.html#xmlrpc-module makes it look like we shouldn't need to switch on 2 vs 3 here. Does that not work?

setup.py
```diff
@@ -53,8 +53,12 @@ class BuildProtoCommand(Command):
   def initialize_options(self):
     self.skip_proto = False
     try:
-      prefix = subprocess.getoutput(
-          'pkg-config --variable prefix protobuf').strip()
+      if sys.version_info[0] < 3:
```
charlesnicholson: Does http://python-future.org/compatible_idioms.html#subprocess-check-output not let us avoid the test for 2 vs 3?

wallacbe:
Made updates, I left xmlrpc alone.  Following futures instructions still gave me error.

# Python 2 and 3 (after ``pip install future``):
import xmlrpc.client

Appear to be some issues on github for this:
https://github.com/PythonCharmers/python-future/issues/280

PR #679, opened 2017-12-03
Merge python3 into master

wallacbe:


wallacbe:
PTAL

openhtf/core/station_api.py
```diff
@@ -86,7 +86,7 @@
 
 # Fix for xmlrpclib to use <i8> for longs and ints instead of <int>, because our
 # timestamps are in millis, which are too big for 4-byte ints.
-xmlrpclib.Marshaller.dispatch[long] = xmlrpclib.Marshaller.dispatch[int] = (
+xmlrpc.client.Marshaller.dispatch[int] = xmlrpc.client.Marshaller.dispatch[int] = (
```
charlesnicholson: The same expression is used twice here, and I'm not sure what the backwards compatibility story here is since int and long are separate types on Python2. We have builtins for compatibility and the `int(123)`.

I wonder if the right thing to do here instead is:

`xmlrpc.client.Marshaller.dispatch[builtins.int] = xmlrpc.client.Marshaller.dispatch[int] = (`

That's redundant on Python3 since `builtins.int` is the same as `int` but maybe we don't care?

openhtf/core/station_api.py
```diff
@@ -453,7 +453,7 @@ def history(self):
     """
     last_start_time = self._cached_history.last_start_time(self.test_uid)
     new_history = self.shared_proxy.get_history_after(
-        self.test_uid, long(last_start_time))
+        self.test_uid, int(last_start_time))
```
charlesnicholson: I think 2to3 did this, but to preserve the intent of the original code explicitly requesting wider-than-integer size, I think we need `from builtins import int` at the top of this file.

Alternately we can revert these long => int changes and add `from past.builtins import long`

http://python-future.org/compatible_idioms.html#long-integers

openhtf/core/station_api.py
```diff
@@ -680,10 +680,10 @@ def list_tests(self):
     retval = [{
         'test_uid': test.uid,
         'test_name': test.get_option('name'),
-        'created_time_millis': long(test.created_time_millis),
+        'created_time_millis': int(test.created_time_millis),
```
charlesnicholson: ditto int / long, see above conversation

openhtf/core/station_api.py
charlesnicholson: Alternately maybe
```
from past.builtins import long
...
xmlrpc.client.Marshaller.dispatch[long] = ...[int] = (
```

openhtf/core/station_api.py
```diff
@@ -680,10 +680,10 @@ def list_tests(self):
...
         'last_run_time_millis':
-            test.last_run_time_millis and long(test.last_run_time_millis),
-    } for test in openhtf.Test.TEST_INSTANCES.values() if test.uid is not None]
+            test.last_run_time_millis and int(test.last_run_time_millis),
```
charlesnicholson: ditto int / long

openhtf/core/test_executor.py
```diff
@@ -109,7 +109,7 @@ def finalize(self):
   def wait(self):
     """Waits until death."""
     try:
-      self.join(sys.float_info.max)  # Timeout needed for SIGINT handling.
+      self.join(5)  # Timeout needed for SIGINT handling.
```
charlesnicholson: This is a question for the maintainers. In Python3, self.join raises a ValueError since sys.float_info.max is too large of a value to wait on.

This is only test code, and the timeout doesn't seem to be intended to be tested, so I put 5 seconds in as a non-float-max placeholder. Is this reasonable?

openhtf/core/test_state.py
```diff
@@ -426,7 +426,7 @@ def _asdict(self):
         'name': self.name,
         'codeinfo': self.phase_record.codeinfo,
         'descriptor_id': self.phase_record.descriptor_id,
-        'start_time_millis': long(self.phase_record.start_time_millis),
+        'start_time_millis': int(self.phase_record.start_time_millis),
```
charlesnicholson: This code will behave subtly different on Python2 if the value is > 32 bits.
I think adding `from past.builtins import long` at the top and reverting to long is the right thing here. (2to3 did this)

openhtf/output/callbacks/__init__.py
```diff
@@ -22,13 +22,14 @@
 
 import base64
 import contextlib
-import cPickle as pickle
+import pickle as pickle
```
charlesnicholson: This may cause a large slowdown on Python2.

Maybe
```
try:
   import cPickle as pickle
except:
   import pickle
```
from https://pymotw.com/2/pickle/
?

openhtf/output/callbacks/json_factory.py
```diff
@@ -39,7 +39,7 @@ def convert_to_dict(self, test_record):
     as_dict = data.convert_to_base_types(test_record)
     if self.inline_attachments:
       for phase, original_phase in zip(as_dict['phases'], test_record.phases):
-        for name, attachment in phase['attachments'].iteritems():
+        for name, attachment in phase['attachments'].items():
           original_data = original_phase.attachments[name].data
-          attachment['data'] = base64.standard_b64encode(original_data)
+          attachment['data'] = base64.standard_b64encode(original_data).decode('utf-8')
```
charlesnicholson: Maintainers, please look at this, it was subtle. My understanding of the code was that the attachments were meant to be held as strings that held b64 encodings. Because of that, I chose to decode the byte string into a regular Python3 unicode string via utf-8. WDYT?

openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -121,7 +121,7 @@ def _populate_header(record, testrun):
     attachment = testrun.info_parameters.add()
     attachment.name = 'config'
     attachment.value_binary = json.dumps(
-        record.metadata['config'], sort_keys=True, indent=4)
+        record.metadata['config'], sort_keys=True, indent=4).encode('utf-8')
```
charlesnicholson: Maintainers, PTAL at this too, do you want these json dumps to be unicode or byte strings?

openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -155,7 +155,7 @@ def _extract_attachments(phase, testrun, used_parameter_names):
     name = _ensure_unique_parameter_name(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
-    if isinstance(attachment_data, unicode):
+    if isinstance(attachment_data, str):
```
charlesnicholson: Does this work as intended on Python2 now? 2to3 did this, with no concern for backwards compatibility.

openhtf/output/proto/test_runs.proto
```diff
@@ -21,7 +21,7 @@ syntax = "proto2";
 package openhtf;
 
 // import "assembly.proto";
-import "units.proto";
+import "openhtf/output/proto/units.proto";
```
charlesnicholson: This was an artifact of an older version of the proto runtime + compiler; the handling of relative imports changed. Because this proto includes its peer relative to the project root, I also altered the proto compiler invocation to run from the root of openhtf.

openhtf/output/web_gui/__main__.py
```diff
@@ -16,7 +16,7 @@
 """Entry point to run OpenHTF's built-in web gui server."""
 
 
-from __future__ import print_function
+
```
charlesnicholson: I think we probably want to add this line back in; 2to3 deleted it.

openhtf/plugs/usb/adb_device.py
```diff
@@ -30,7 +30,7 @@
 class, subclass, and protocol.
 """
 
-import cStringIO
+import io
```
charlesnicholson: Maintainers, maybe someone can comment on whether cStringIO is strictly required on Python2, vs regular io?

openhtf/plugs/usb/filesync_service.py
```diff
@@ -223,7 +223,7 @@ def _check_for_fail_message(self, transport, exc_info, timeout):  # pylint: disa
       if sys.exc_info()[0] is usb_exceptions.AdbRemoteError:
         raise
     # Otherwise reraise the original exception.
-    raise exc_info[0], exc_info[1], exc_info[2]
+    raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
```
charlesnicholson: I think this isn't Python2 compatible:
http://python-future.org/compatible_idioms.html#raising-exceptions

I think we might need `raise_with_traceback` here instead.

openhtf/plugs/user_input.py
```diff
@@ -73,20 +73,20 @@ def Stop(self):
...
+      print("Nevermind; prompt was answered from elsewhere.")
 
   def run(self):
     """Main logic for this thread to execute."""
     try:
       if platform.system() == 'Windows':
         # Windows doesn't support file-like objects for select(), so fall back
         # to raw_input().
-        response = raw_input(self._message + '\n\r')
+        response = input(self._message + '\n\r')
```
charlesnicholson: This 2to3 change means we need `from builtins import input` at the top for Python2.
http://python-future.org/compatible_idioms.html#raw-input

openhtf/util/data.py
```diff
@@ -32,7 +32,7 @@
 from enum import Enum
 
 # Used by convert_to_base_types().
-PASSTHROUGH_TYPES = {bool, bytes, int, long, type(None), unicode}
+PASSTHROUGH_TYPES = {bool, bytes, int, int, type(None), str}
```
charlesnicholson: More 2to3 fun. I think we want to put `long` back, and `from past.builtins import long` at the top.

openhtf/util/data.py
```diff
@@ -84,7 +84,7 @@ def assert_records_equal_nonvolatile(first, second, volatile_fields, indent=0):
     assert_records_equal_nonvolatile(first._asdict(), second._asdict(),
                                      volatile_fields, indent)
   elif hasattr(first, '__iter__') and hasattr(second, '__iter__'):
-    for idx, (fir, sec) in enumerate(itertools.izip(first, second)):
+    for idx, (fir, sec) in enumerate(zip(first, second)):
```
charlesnicholson: I don't understand how Python2 handles this, but
http://python-future.org/compatible_idioms.html#zip-izip
makes it seem like we should change `zip` to `izip` and add `from builtins import izip` to the top.

openhtf/util/data.py
```diff
@@ -153,7 +153,7 @@ def convert_to_base_types(obj, ignore_keys=tuple(), tuple_type=tuple):
 
   # Convert numeric types (e.g. numpy ints and floats) into built-in types.
   elif isinstance(obj, numbers.Integral):
-    return long(obj)
+    return int(obj)
```
charlesnicholson: `from past.builtins import long` at the top, then `return long(obj)` here

openhtf/util/data.py
```diff
@@ -181,9 +181,9 @@ def _sizeof(current_obj):
 
     if isinstance(current_obj, dict):
       size += sum(map(sizeof, itertools.chain.from_iterable(
-          current_obj.iteritems())))
+          iter(current_obj.items()))))
```
charlesnicholson: Unsure what this should look like, since http://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items suggests a few strategies. Does this currently have test coverage?

openhtf/util/exceptions.py
```diff
@@ -71,4 +71,4 @@ def main():
   line_msg = 'line %s: ' % last_lineno
   if message:
     line_msg += str(message)
-  raise exc_type(line_msg, *args, **kwargs), None, sys.exc_info()[2]
+  raise exc_type(line_msg, *args, **kwargs).with_traceback(sys.exc_info()[2])
```
charlesnicholson: `raise_with_traceback`
http://python-future.org/compatible_idioms.html#raising-exceptions

openhtf/util/logs.py
```diff
@@ -149,7 +149,7 @@ def filter(self, record):
       record.msg = self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, record.msg)
       record.args = tuple([
           self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, str(arg))
-          if isinstance(arg, basestring)
+          if isinstance(arg, str)
```
charlesnicholson: http://python-future.org/compatible_idioms.html#basestring

openhtf/util/test.py
```diff
@@ -285,8 +298,8 @@ def test_wrapper(test_func):
     # Make MagicMock instances for the plugs.
     plug_kwargs = {}  # kwargs to pass to test func.
     plug_typemap = {}  # typemap for PlugManager, maps type to instance.
-    for plug_arg_name, plug_fullname in mock_plugs.iteritems():
-      if isinstance(plug_fullname, basestring):
+    for plug_arg_name, plug_fullname in mock_plugs.items():
+      if isinstance(plug_fullname, str):
```
charlesnicholson: http://python-future.org/compatible_idioms.html#basestring

openhtf/util/test.py
```diff
@@ -355,7 +368,7 @@ def assertion_wrapper(self, phase_or_test_record, *args):
             exc_info = sys.exc_info()
         else:
           if exc_info:
-            raise exc_info[0], exc_info[1], exc_info[2]
+            raise exc_info[0](exc_info[1]).with_traceback(exc_info[2])
```
charlesnicholson: http://python-future.org/compatible_idioms.html#raising-exceptions

openhtf/util/validators.py
```diff
@@ -175,8 +172,8 @@ def __ne__(self, other):
 def equals(value, type=None):
   if isinstance(value, numbers.Number):
     return InRange(minimum=value, maximum=value, type=type)
-  elif isinstance(value, basestring):
-    assert type is None or issubclass(type, basestring), (
+  elif isinstance(value, str):
```
charlesnicholson: http://python-future.org/compatible_idioms.html#basestring

openhtf/util/validators.py
```diff
@@ -175,8 +172,8 @@ def __ne__(self, other):
...
+    assert type is None or issubclass(type, str), (
```
charlesnicholson: http://python-future.org/compatible_idioms.html#basestring

openhtf/util/xmlrpcutil.py
```diff
@@ -14,32 +14,40 @@
...
+import socketserver
+import sys
 import threading
-import xmlrpclib
+import xmlrpc.client
+import collections
 
 DEFAULT_PROXY_TIMEOUT_S = 3
 
+if sys.version_info[0] < 3:
```
charlesnicholson: Maybe a comment here about why this is necessary and the bugs in six that prevent us from using it?

setup.py
```diff
@@ -63,26 +63,26 @@ def initialize_options(self):
...
     pass
 
   def run(self):
     if self.skip_proto:
-      print 'Skipping building protocol buffers.'
+      print('Skipping building protocol buffers.')
       return
     # Build regular proto files.
-    protos = glob.glob(os.path.join(self.indir, '*.proto'))
+    protos = glob.glob(os.path.join(self.indir, 'openhtf', 'output', 'proto', '*.proto'))
```
charlesnicholson: Maintainers: this is what I mentioned above. Since proto imports are relative to the project root and not their specific location, it seemed correct to invoke the proto compiler from the root as well.

Note that this doesn't change the strategy or location of the pb2.py files, it just aligns the new import path requirement with the compilation paths.

test/output/callbacks/callbacks_test.py
```diff
@@ -49,15 +50,18 @@ def _save_result(test_record):
   def test_json(self, user_mock):
     user_mock.prompt.return_value = 'SomeWidget'
     record = yield self._test
-    json_output = StringIO()
+    if sys.version_info[0] < 3:
```
charlesnicholson: This is suspicious and makes me think I screwed something up when migrating the attachment strings. Maintainers, any insights / opinions?

test/util/data_test.py
```diff
@@ -14,7 +14,9 @@
 
 import unittest
 import mock
+import sys
```
charlesnicholson: unnecessary?

test/util/data_test.py
```diff
@@ -39,13 +41,14 @@ class FloatSubclass(float):
...
     self.assertIs(type(converted['list']), list)
     self.assertIs(type(converted['tuple']), tuple)
     self.assertIs(type(converted['str']), str)
-    self.assertIs(type(converted['unicode']), unicode)
-    self.assertIs(type(converted['int']), int)
+    self.assertIs(type(converted['unicode']), str)
+    assert isinstance(converted['int'], int)
     self.assertIs(type(converted['float']), float)
-    self.assertIs(type(converted['long']), long)
+    assert isinstance(converted['long'], int)
```
charlesnicholson: This smells like 2to3, it might be better to use `past.builtins.long`

test/util/validators_test.py
```diff
@@ -84,7 +85,7 @@ def test_comparable_with_equivalent_equals_validator(self):
 class TestEqualsFactory(unittest.TestCase):
 
   def test_with_numbers(self):
-    for expected in [1, 1.0, decimal.Decimal(1), 1L]:
+    for expected in [1, 1.0, decimal.Decimal(1), 1]:
```
charlesnicholson: This last 1 should be `int(1)` with `int` coming from `builtins`, that way it gets a long.
http://python-future.org/compatible_idioms.html#long-integers

wallacbe:
Manually testing today uncovered issue when adding:
from openhtf.output.callbacks import mfg_inspector

serialized_pb=_b('\n!openhtf/output/proto/guzzle.proto\x12\x07openhtf\"N\n\x0fTestRunEnvelope\x12*\n\x0cpayload_type\x18\x02 \x01(\x0e\x32\x14.openhtf.PayloadType\x12\x0f\n\x07payload\x18\x03 \x01(\x0c*S\n\x0bPayloadType\x12\x17\n\x13\x43OMPRESSED_TEST_RUN\x10\x01\x12\x13\n\x0f\x43OMPRESSED_VARZ\x10\x02\x12\x16\n\x12\x43OMPRESSED_LOGFILE\x10\x03')
TypeError: __init__() got an unexpected keyword argument 'syntax'

fahhem:
That's a protobuf-protoc version discrepancy

openhtf/output/callbacks/json_factory.py
grybmadsci: SGTM, anything base64 encoded should make sense as utf-8 - your understanding is correct, that's the intention.

openhtf/output/callbacks/mfg_inspector.py
grybmadsci: config should generally be YAML text from a file so utf-8 makes sense as an encoding here also.

openhtf/plugs/usb/adb_device.py
grybmadsci: I don't know of any reason why cStringIO is strictly required, I believe it was potentially used for performance reasons?  i don't believe this adb_device plug is actually used by anyone, ADB interfaces seem to be everyone's favorite thing to implement themselves :)

setup.py
grybmadsci: SGTM

openhtf/util/data.py
wallacbe: from builtins import izip  didn't appear to work

wallacbe:
Made some updates based on comments

contrib/poll_stations.py
```diff
@@ -32,7 +32,7 @@
 
 import logging
 import os
-import Queue
+import queue
```
fahhem: 3 specific

examples/all_the_things.py
```diff
@@ -91,7 +91,7 @@ def set_measurements(test):
 def dimensions(test):
   for dim in range(5):
     test.measurements.dimensions[dim] = 1 << dim
-  for x, y, z in zip(range(1, 5), range(21, 25), range (101, 105)):
+  for x, y, z in zip(list(range(1, 5)), list(range(21, 25)), list(range(101, 105))):
```
fahhem: not necessary, you can zip iterators in PY3, and range is an iterator in PY3. This was an unnecessary 2to3 change

examples/repeat.py
```diff
@@ -40,7 +40,7 @@ def __init__(self):
   def run(self):
     """Increments counter and raises an exception for first two runs."""
     self.count += 1
-    print 'FailTwicePlug: Run number %s' % (self.count)
+    print('FailTwicePlug: Run number %s' % (self.count))
```
fahhem: missing: `from __future__ import print_function`

openhtf/__init__.py
```diff
@@ -192,14 +192,14 @@ def configure(self, **kwargs):
...
-    for key, value in kwargs.iteritems():
+    for key, value in kwargs.items():
       setattr(self._test_options, key, value)
 
   @classmethod
   def handle_sig_int(cls, *_):
     if cls.TEST_INSTANCES:
       _LOG.error('Received SIGINT, stopping all tests.')
-      for test in cls.TEST_INSTANCES.values():
+      for test in list(cls.TEST_INSTANCES.values()):
```
fahhem: don't need to list() something in for, since it accepts iterators too. Another unnecessary 2to3 change

examples/with_plugs.py
```diff
@@ -52,7 +52,7 @@ def _get_command(self, count):
 
   def run(self, count):
     command = self._get_command(count)
-    print "running: %s" % ' '.join(command)
+    print("running: %s" % ' '.join(command))
```
fahhem: another missing `__future__` import

openhtf/core/measurements.py
```diff
@@ -198,7 +198,7 @@ def with_dimensions(self, *dimensions):
 
   def with_validator(self, validator):
     """Add a validator callback to this Measurement, chainable."""
-    if not callable(validator):
+    if not isinstance(validator, collections.Callable):
```
fahhem: not necessary? callable is in PY3.5 (as of PY3.2). PY3.3 is actually End-of-Life'd already (as of September 2017): https://www.python.org/dev/peps/pep-0398/#lifespan

openhtf/core/phase_executor.py
```diff
@@ -91,12 +91,13 @@ class PhaseExecutionOutcome(collections.namedtuple(
   other value will raise an InvalidPhaseResultError.
   """
 
-  def __init__(self, phase_result):
+  def __new__(cls, phase_result):
```
fahhem: why is this changed? We used `__init__` because we're not modifying anything, just checking that the arguments are valid

openhtf/core/station_api.py
```diff
@@ -683,7 +684,7 @@ def list_tests(self):
         'created_time_millis': long(test.created_time_millis),
         'last_run_time_millis':
             test.last_run_time_millis and long(test.last_run_time_millis),
-    } for test in openhtf.Test.TEST_INSTANCES.values() if test.uid is not None]
+    } for test in list(openhtf.Test.TEST_INSTANCES.values()) if test.uid is not None]
```
fahhem: not necessary `list()`

openhtf/core/test_executor.py
```diff
@@ -109,7 +109,7 @@ def finalize(self):
   def wait(self):
     """Waits until death."""
     try:
-      self.join(sys.float_info.max)  # Timeout needed for SIGINT handling.
+      self.join(315400000)  # Timeout needed for SIGINT handling.
```
fahhem: Why was this changed? PY3.5 has this still

openhtf/output/callbacks/__init__.py
```diff
@@ -81,7 +87,7 @@ def open_output_file(self, test_record):
     record_dict = data.convert_to_base_types(
         test_record, ignore_keys=('code_info', 'phases', 'log_records'))
     pattern = self.filename_pattern
-    if isinstance(pattern, basestring) or callable(pattern):
+    if isinstance(pattern, str) or isinstance(pattern, collections.Callable):
```
fahhem: another unnecessary callable change

openhtf/output/callbacks/__init__.py
```diff
@@ -22,13 +22,17 @@
...
+   import cPickle as pickle
+except:
+   import pickle
 import os
 import shutil
 import tempfile
 
 from openhtf import util
 from openhtf.util import data
+import collections
```
fahhem: unnecessary if you remove the `isinstance()` belowe

openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -173,8 +173,8 @@ def _mangle_measurement(name, measured_value, measurement, mangled_parameters,
   We generate these by doing some name mangling, using some sane limits for
   very large multidimensional measurements.
   """
-  for coord, val in measured_value.value_dict.items(
-      )[:MAX_PARAMS_PER_MEASUREMENT]:
+  for coord, val in list(measured_value.value_dict.items(
```
fahhem: unnecessary

openhtf/output/web_gui/__init__.py
```diff
@@ -239,12 +239,12 @@ def run(self):
     if self._disable_discovery:
       _LOG.debug('Station discovery is disabled; only using static stations.')
 
-    self._handle_stations(self.stations.iterkeys())
+    self._handle_stations(iter(self.stations.keys()))
```
fahhem: unnecessary `iter()` call

openhtf/plugs/usb/__init__.py
```diff
@@ -28,7 +28,7 @@ def MyPhase(test, adb):
     adb.Shell('ls')
 """
 import argparse
-import commands
+import subprocess
```
fahhem: just remove the line

openhtf/plugs/usb/adb_device.py
fahhem: it was definitely for performance compared to StringIO and due to ignorance of the io module

openhtf/plugs/usb/adb_protocol.py
```diff
@@ -79,7 +79,7 @@
 import collections
 import itertools
 import logging
-import Queue
+import queue
```
fahhem: not PY2 compatible

either version check or do a try-except to `import Queue as queue`

openhtf/plugs/usb/fastboot_device.py
```diff
@@ -85,7 +86,7 @@ def __getattr__(self, attr):  # pylint: disable=invalid-name
       raise usb_exceptions.HandleClosedError()
 
     val = getattr(self._protocol, attr)
-    if callable(val):
+    if isinstance(val, collections.Callable):
```
fahhem: unnecessary change, this whole file can be reverted

openhtf/plugs/user_input.py
```diff
@@ -73,20 +74,20 @@ def Stop(self):
...
         self._answered = True
         self._callback(response)
       else:
         # First, display the prompt to the console.
-        print self._message
+        print(self._message)
```
fahhem: `from __future__ import print_function`

openhtf/util/__init__.py
```diff
@@ -143,9 +144,9 @@ def format_string(target, kwargs):
   """
   if not target:
     return target
-  if callable(target):
+  if isinstance(target, collections.Callable):
```
fahhem: unnecessary

openhtf/util/data.py
fahhem: this should be fine as-is

openhtf/core/test_executor.py
wallacbe: I get this: OverflowError: timestamp too large to convert to C _PyTime_t

openhtf/output/callbacks/mfg_inspector.py
wallacbe: I see this:
      for coord, val in measured_value.value_dict.items(
>         )[:MAX_PARAMS_PER_MEASUREMENT]:
E         TypeError: 'odict_items' object is not subscriptable

openhtf/core/phase_executor.py
wallacbe: I see this:
    self = super(PhaseExecutionOutcome, cls).__new__(cls, phase_result)
  File "<string>", line 14, in __new__
TypeError: tuple.__new__(X): X is not a type object (PhaseExecutionOutcome)

charlesnicholson: https://stackoverflow.com/questions/42385916/inheriting-from-a-namedtuple-base-class-python

openhtf/core/test_executor.py
charlesnicholson: sys.float_info.max isn't a very reasonable value to pass to self.join, even as a placeholder. I'd suggest replacing them with something like "60" and leaving a comment saying that the timeout is not actually used.

wallacbe: Made update

contrib/poll_stations.py
```diff
@@ -29,10 +29,9 @@
 Outputs something like, which updates regularly:
 
 """
-
+from queue import Queue
```
fahhem: ohh, future provides `queue`... never mind, go back to `import queue`, sorry!

contrib/poll_stations.py
```diff
@@ -182,7 +181,7 @@ def print_station(self, station):
           self.update_threads[remote_test] = update_thread
     except socket.error as e:
       print(' |-- Connection Error: %s' % e)
-    print
+    print()
```
fahhem: `from __future__ import print_function`

Can you please test these files in PY2?

openhtf/core/test_executor.py
```diff
@@ -109,7 +109,8 @@ def finalize(self):
   def wait(self):
     """Waits until death."""
     try:
-      self.join(sys.float_info.max)  # Timeout needed for SIGINT handling.
+      # Timeout needed for SIGINT handling. Timeout is not actualy used.
```
fahhem: typo: actually

openhtf/core/test_record.py
```diff
@@ -125,7 +125,7 @@ def _get_source_safely(obj):
     return ''
 
 
-class CodeInfo(mutablerecords.Record(
+class CodeInfo(mutablerecords.HashableRecord(
```
fahhem: not sure why this was changed?

openhtf/output/callbacks/__init__.py
```diff
@@ -81,7 +86,7 @@ def open_output_file(self, test_record):
     record_dict = data.convert_to_base_types(
         test_record, ignore_keys=('code_info', 'phases', 'log_records'))
     pattern = self.filename_pattern
-    if isinstance(pattern, basestring) or callable(pattern):
+    if isinstance(pattern, str):
```
fahhem: why did you remove `callable(pattern)`?

fahhem: Why did you remove `callable(pattern)`?

openhtf/output/callbacks/mfg_inspector.py
fahhem: Ah, I missed the indexing, sorry

openhtf/plugs/usb/usb_handle.py
```diff
@@ -58,7 +58,7 @@ def wrapper_requiring_open_handle(self, *args, **kwargs):
   return wrapper_requiring_open_handle
 
 
-class UsbHandle(object):
+class UsbHandle(object, metaclass=abc.ABCMeta):
```
fahhem: PY3 only, please use `with_metaclass`

openhtf/util/data.py
fahhem: @wallacbe, missed this?

openhtf/util/data.py
fahhem: This should. We can remove the `iter()` call here since `.items()` is an iterable in PY2 and 3

openhtf/plugs/usb/usb_handle.py
wallacbe: So
from future.utils import with_metaclass
class UsbHandle(with_metaclass(object)):
?

contrib/poll_stations.py
wallacbe: Done

contrib/poll_stations.py
wallacbe: Added from __future__ import print_function. I see "Waiting for stations..." when running from PY2 and PY3 virtualenv

openhtf/core/test_executor.py
wallacbe: Done.  Fixed spelling error

openhtf/core/test_record.py
wallacbe: Changed to use Record.

openhtf/output/callbacks/__init__.py
wallacbe: Added callable(pattern)? back in.

wallacbe: Added callable(pattern)? back in.

openhtf/plugs/usb/usb_handle.py
wallacbe: Failing unit test.  I may have incorrectly implemented.

openhtf/core/test_record.py
wallacbe: If I don't use HashableRecord on PY3 I see:

cls = <Record: PhaseDescriptor>

    def __hash__(cls):
        return hash(
            (cls.required_attributes,
>            frozenset(cls.optional_attributes.items())))
E       TypeError: unhashable type: 'CodeInfo'

wallacbe:
PTAL, addressed most comments

Kenadia:
@wallacbe, @jettisonjoe, and I chatted, and we think this is ready to go. I think we can handle dict iteration more consistently (see PEP 469) but we can follow up on that separately.

PR #681, opened 2017-12-05
Implement plug placeholders with defaults.

arsharma1:
Implement a second type of placeholders that allow defaulting to the specified type if with_plugs is not used.


arsharma1:
Added https://github.com/google/openhtf/pull/683 to replace this CL with one that addresses (2).

PR #683, opened 2017-12-11
Add auto_placeholder for plugs.

arsharma1:
Plugs that set the auto_placeholder attribute can be overridden with a
phase's with_plugs without needing to use a .placeholder declaration.

### DIFF
--- a/test/phase_descriptor_test.py
+++ b/test/phase_descriptor_test.py
@@ -40,10 +40,29 @@ class ExtraPlug(plugs.BasePlug):
   def echo(self, phrase):
     return '%s says %s' % (self.name, phrase)
 
+
 @openhtf.PhaseOptions(name='extra_plug_func[{plug.name}][{phrase}]')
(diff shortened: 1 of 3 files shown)


test/phase_descriptor_test.py
```diff
@@ -96,3 +115,18 @@ def test_with_plugs(self):
...
+      self.assertEqual(1, len(phase.plugs))
+
+  def test_with_plugs_subclass_auto_placeholder_error(self):
+      with self.assertRaises(plugs.InvalidPlugError):
+          _ = sub_placeholder_using_plug.with_plugs(
+              subplaced=SubPlaceholderCapablePlug)
+
+  def test_with_plugs_auto_placeholder_non_subclass_error(self):
+      with self.assertRaises(plugs.InvalidPlugError):
+          _ = placeholder_using_plug.with_plugs(placed=ExtraPlug)
```
Kenadia: nit unnecessary assignment

PR #691, opened 2018-01-07
Attempt to run 'protoc' without a prefix if one isn't found.

charlesnicholson:
If protoc isn't found via pkg-config or at the hard-coded locations, attempt to just invoke it without any prefix path. This allows for protobuf to be sandboxed (in a git repository, for example).

This change preserves the existing behavior by explicitly preferring '/usr/bin/protoc' on Linux and '/usr/local/bin/protoc' on OSX, but also allows control of the specific protoc by manipulating the path beforehand.

Fix a Python3 bug, convert subprocess.check_output to a string and append standard (non-byte) strings to it. This is redundant but functional in Python2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ class BuildProtoCommand(Command):
     self.skip_proto = False
     try:
       prefix = subprocess.check_output(
-          'pkg-config --variable prefix protobuf'.split()).strip()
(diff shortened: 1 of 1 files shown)


charlesnicholson:
@wallacbe FYI

wallacbe:
This OK to merge?

Kenadia:
LGTM too. Also tried it and it works for me (mac)

Issue #690, opened 2018-01-06
DeprecationWarning: inspect.getargspec() is deprecated, 


wallacbe:
Observed a deprecation warning when running python3 from pypi

python3.6/site-packages/openhtf/__init__.py:528: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  arg_info = inspect.getargspec(self.func)

PR #701, opened 2018-01-15
add getfullargspec() for python3

wallacbe:
https://github.com/google/openhtf/issues/690

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -525,10 +525,16 @@ class PhaseDescriptor(mutablerecords.Record(
     kwargs = dict(self.extra_kwargs)
     kwargs.update(test_state.plug_manager.provide_plugs(
         (plug.name, plug.cls) for plug in self.plugs if plug.update_kwargs))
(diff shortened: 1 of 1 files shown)


wallacbe:
PTAL

openhtf/__init__.py
```diff
@@ -525,10 +525,15 @@ def __call__(self, test_state):
     kwargs = dict(self.extra_kwargs)
     kwargs.update(test_state.plug_manager.provide_plugs(
         (plug.name, plug.cls) for plug in self.plugs if plug.update_kwargs))
-    arg_info = inspect.getargspec(self.func)
+    try:
+      arg_info = inspect.getfullargspec(self.func)
```
jettisonjoe: Maybe put this behind a hasattr check instead of catching the AttributeError.

Kenadia: Or maybe a version check?
`if sys.version_info[0] < 3:`
That make it more clear why this is needed.

wallacbe:
Made changes

PR #703, opened 2018-01-19
Fix setup.py for systems with older setuptools.

### LINKED ISSUE #693: pypi release 1.2.0 is missing built proto files in openhtf/output/proto
pypi release 1.2.0 is missing built proto files in openhtf/output/proto.  They are present in release 1.1.0.   This is observed by downloading the tarball from pypi.  Can this be corrected?  Thanks

jettisonjoe:
Change includes a comment explaining why we might want to use the old school approach. I've seen this cause silent install failures on two lab machines where setup bailed out before building protocol buffers. In those cases the failure didn't become apparent until import time, when import of the unbuilt protos failed.

Fixes #693

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/703)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -178,7 +178,7 @@ class PyTestCommand(test):
 
 setup(
     name='openhtf',
-    version='1.2.0',
+    version='1.2.1',
     description='OpenHTF, the open hardware testing framework.',
     author='John Hawley',
     author_email='madsci@google.com',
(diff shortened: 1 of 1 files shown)


jettisonjoe:
I logged into one of the failing systems and tried running the build_proto command by hand in a fresh virtualenv. There was a vague error message about the format of install_requires argument and that line was the last one to change.

PR #705, opened 2018-01-22
Include unbuilt protos in the package not just the source repo.

jettisonjoe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/705)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -178,14 +178,15 @@ class PyTestCommand(test):
 
 setup(
     name='openhtf',
-    version='1.2.1',
+    version='1.2.2',
     description='OpenHTF, the open hardware testing framework.',
     author='John Hawley',
     author_email='madsci@google.com',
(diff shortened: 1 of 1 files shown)


fahhem:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/705)*

jettisonjoe:
Removed version bump so we can do that in a separate PR. I have a theory as to when this broke. Looking into it now.

jettisonjoe:
My theory was that we were doing wheel-based distribution before and switched to source-based distribution at some point (multi-platform support?), but I can't find where we would have had to change that.

jettisonjoe:
Fixes #693

PR #706, opened 2018-01-22
Fix compatibility regression in inspector output module.

jettisonjoe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/706)

### DIFF
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -31,7 +31,6 @@ import json
 import logging
 import numbers
 import os
-import sys
 import threading
 import zlib
 try:
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -31,6 +31,7 @@ import json
 import logging
 import numbers
 import os
+import sys
 import threading
 import zlib
 
@@ -155,8 +156,9 @@ def _extract_attachments(phase, testrun, used_parameter_names):
     name = _ensure_unique_parameter_name(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
-    if isinstance(attachment_data, str):
-      attachment_data = attachment_data.encode('utf8')
+    if (isinstance(attachment_data, unicode) or
+        sys.version_info[0] >= 3 and isinstance(attachment_data, str)):
+      attachment_data = attachment_data.encode('utf-8')
     testrun_param.value_binary = attachment_data
     if mimetype in MIMETYPE_MAP:
       testrun_param.type = MIMETYPE_MAP[mimetype]
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -34,6 +34,10 @@ import os
 import sys
 import threading
 import zlib
+try:
+  from past.types import unicode
+except ImportError:
+  pass
 
 import httplib2
 import oauth2client.client
@@ -156,8 +160,7 @@ def _extract_attachments(phase, testrun, used_parameter_names):
     name = _ensure_unique_parameter_name(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
-    if (sys.version_info[0] < 3 and isinstance(attachment_data, unicode) or
-        sys.version_info[0] >= 3 and isinstance(attachment_data, str)):
+    if isinstance(attachment_data, unicode):
       attachment_data = attachment_data.encode('utf-8')
     testrun_param.value_binary = attachment_data
     if mimetype in MIMETYPE_MAP:
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -156,7 +156,7 @@ def _extract_attachments(phase, testrun, used_parameter_names):
     name = _ensure_unique_parameter_name(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
-    if (isinstance(attachment_data, unicode) or
+    if (sys.version_info[0] < 3 and isinstance(attachment_data, unicode) or
         sys.version_info[0] >= 3 and isinstance(attachment_data, str)):
       attachment_data = attachment_data.encode('utf-8')
     testrun_param.value_binary = attachment_data


openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -155,8 +156,9 @@ def _extract_attachments(phase, testrun, used_parameter_names):
     name = _ensure_unique_parameter_name(name, used_parameter_names)
     testrun_param = testrun.info_parameters.add()
     testrun_param.name = name
-    if isinstance(attachment_data, str):
-      attachment_data = attachment_data.encode('utf8')
+    if (isinstance(attachment_data, unicode) or
```
Kenadia: I think this will break with `'unicode' is not defined` in python 3

fahhem:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 160 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UPdEHKl2CPXKc3Tfy:-L3UPdEIct2XzqaO_EWk:bv6yutj) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L160)):*
> ```Python
>     testrun_param.name = name
>     if (isinstance(attachment_data, unicode) or
>         sys.version_info[0] >= 3 and isinstance(attachment_data, str)):
> ```

Hmm, `str == unicode` (http://python-future.org/reference.html?highlight=unicode#past.builtins.unicode)? If you replace `str` with `unicode`, then this becomes redundant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

jettisonjoe:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 160 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UPdEHKl2CPXKc3Tfy:-L3UX6XpXlmADaSVIVns:bcuzsxb) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L160)):*
<details><summary><i>Previously, fahhem (Fahrzin Hemmati) wrote…</i></summary><blockquote>

Hmm, `str == unicode` (http://python-future.org/reference.html?highlight=unicode#past.builtins.unicode)? If you replace `str` with `unicode`, then this becomes redundant.
</blockquote></details>

Interesting, because `unicode` is what we had before, so I assumed it was changed consciously for Python3 support:

https://github.com/google/openhtf/blame/109771fcad5f255da175b230a28f67e8afd40568/openhtf/output/callbacks/mfg_inspector.py#L158

I'm great with changing it back to just `unicode` if that gets the same result.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

jettisonjoe:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 159 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UOZQGOnUj9zy-vxuL-r1-159:-L3UXZqSD6UW3Z7YRtxX:byzwruj) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L159)):*
<details><summary><i>Previously, Kenadia (Kenneth Schiller) wrote…</i></summary><blockquote>

I think this will break with `'unicode' is not defined` in python 3
</blockquote></details>

I pushed a more explicit check, but actually see Farz's reply below. Maybe we can revert to just `unicode`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

jettisonjoe:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 160 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UPdEHKl2CPXKc3Tfy:-L3V9KAIhXOvQlEQ1ENo:b-llau9i) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L160)):*
<details><summary><i>Previously, jettisonjoe (Joe Ethier) wrote…</i></summary><blockquote>

Interesting, because `unicode` is what we had before, so I assumed it was changed consciously for Python3 support:

https://github.com/google/openhtf/blame/109771fcad5f255da175b230a28f67e8afd40568/openhtf/output/callbacks/mfg_inspector.py#L158

I'm great with changing it back to just `unicode` if that gets the same result.
</blockquote></details>

Kenneth is right; there is no more name `unicode` in Python3. My latest patch with the more explicit check seems to work, and I don't think we need to get too fancy here. PTAL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

Kenadia:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 159 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UOZQGOnUj9zy-vxuL-r1-159:-L3ZCeNM2c8mOt3NcH8J:bwvekfa) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L159)):*
<details><summary><i>Previously, jettisonjoe (Joe Ethier) wrote…</i></summary><blockquote>

I pushed a more explicit check, but actually see Farz's reply below. Maybe we can revert to just `unicode`?
</blockquote></details>

```python
from past.types import unicode
...
if isinstance(attachment_data, unicode):
  ...
```

This is the best solution since it makes the intention the most clear

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

jettisonjoe:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[openhtf/output/callbacks/mfg_inspector.py, line 159 at r1](https://reviewable.io:443/reviews/google/openhtf/706#-L3UOZQGOnUj9zy-vxuL-r1-159:-L3ZY_EgDAo3ZBSnLsi6:b-gdnvj5) ([raw file](https://github.com/google/openhtf/blob/467487d9d2c274341c19094031fef83339bfd13f/openhtf/output/callbacks/mfg_inspector.py#L159)):*
<details><summary><i>Previously, Kenadia (Kenneth Schiller) wrote…</i></summary><blockquote>

```python
from past.types import unicode
...
if isinstance(attachment_data, unicode):
  ...
```

This is the best solution since it makes the intention the most clear
</blockquote></details>

Done. I had to use  a try/except around the import, though, since `past` isn't a thing in 2.7.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706)*

Kenadia:
LGTM except for extraneous `import sys`

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706#-:-L3ZZNJutrxY5qNqs-DK:b16sy9)*

jettisonjoe:
Removed extra import.

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/google/openhtf/706#-:-L3Z_OL5T0x27dxaw9oN:b-77i1f0)*

PR #713, opened 2018-01-27
adding __eq__ and __str__ to Dimension to make unit testing easier

kdsudac:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/713)

### DIFF
--- a/openhtf/core/measurements.py
+++ b/openhtf/core/measurements.py
@@ -348,6 +348,12 @@ class Dimension(object):
     self.description = description
     self.unit = unit
 
+  def __eq__(self, other):
+    return (self.description == other.description and self.unit == other.unit)
+
+  def __str__(self):
+    return str(self._asdict())
+
   @classmethod
   def from_unit_descriptor(cls, unit_desc):
     return cls(unit=unit_desc)
--- a/openhtf/core/measurements.py
+++ b/openhtf/core/measurements.py
@@ -351,8 +351,11 @@ class Dimension(object):
   def __eq__(self, other):
     return (self.description == other.description and self.unit == other.unit)
 
-  def __str__(self):
-    return str(self._asdict())
+  def __ne__(self, other):
+    return not self == other
+
+  def __repr__(self):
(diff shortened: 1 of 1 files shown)


openhtf/core/measurements.py
```diff
@@ -348,6 +348,12 @@ def __init__(self, description='', unit=units.NO_DIMENSION):
     self.description = description
     self.unit = unit
 
+  def __eq__(self, other):
```
Kenadia: To handle Python 2, when implementing \_\_eq__, implement \_\_ne__ as well

```python
def __ne__(self, other):
  return not self == other
```

discussion: https://stackoverflow.com/questions/4352244/python-should-i-implement-ne-operator-based-on-eq

openhtf/core/measurements.py
```diff
@@ -348,6 +348,12 @@ def __init__(self, description='', unit=units.NO_DIMENSION):
...
+    return (self.description == other.description and self.unit == other.unit)
+
+  def __str__(self):
+    return str(self._asdict())
```
Kenadia: Let's use the pattern commonly used in OpenHTF of including the class name.

Also I think this should be \_\_repr__ not \_\_str__. And \_\_str__ will then default to \_\_repr__.

```python
def __repr__(self):
  return '<%s: %s>' % (type(self).__name__, self._asdict())
```

openhtf/core/measurements.py
Kenadia: (meant to “request changes)

openhtf/core/measurements.py
Kenadia: (meant to “request changes)

openhtf/core/measurements.py
kdsudac: Done.

openhtf/core/measurements.py
kdsudac: Done.

kdsudac:
Thanks for the review and suggestions.

PR #714, opened 2018-01-27
Logging CLI Changes

maddychan:
- Replaced --verbosity and --quiet with --framework-verbosity and --testrecord-verbosity
  - Added option to set level to off

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/714)

### DIFF
--- a/openhtf/util/logs.py
+++ b/openhtf/util/logs.py
@@ -77,20 +77,24 @@ from openhtf.util import argv
 from openhtf.util import functions
 
 
-DEFAULT_LEVEL = 'warning'
+DEFAULT_FRAMEWORK_LEVEL = 'warning'
 DEFAULT_LOGFILE_LEVEL = 'warning'
+DEFAULT_TESTRECORD_LEVEL = 'info'
 QUIET = False
 LOGFILE = None
 
-LEVEL_CHOICES = ['debug', 'info', 'warning', 'error', 'critical']
+LEVEL_CHOICES = ['debug', 'info', 'warning', 'error', 'critical', 'off']
 ARG_PARSER = argv.ModuleParser()
 ARG_PARSER.add_argument(
-    '--verbosity', default=DEFAULT_LEVEL, choices=LEVEL_CHOICES,
-    action=argv.StoreInModule, target='%s.DEFAULT_LEVEL' % __name__,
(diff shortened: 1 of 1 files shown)


openhtf/util/logs.py
```diff
@@ -77,20 +76,24 @@ def MyPhase(test, helper):
 from openhtf.util import functions
 
 
-DEFAULT_LEVEL = 'warning'
+DEFAULT_FRAMEWORK_LEVEL = 'warning'
 DEFAULT_LOGFILE_LEVEL = 'warning'
+DEFAULT_TESTRECORD_LEVEL = 'info'
 QUIET = False
```
Kenadia: Remove `QUIET`

openhtf/util/logs.py
```diff
@@ -116,10 +119,11 @@ def initialize_record_logger(test_uid, test_record, notify_update):
   logger = get_record_logger_for(test_uid)
   # All record loggers have a shared parent that's separately configured, so
   # we want to propagate to that logger.
-  logger.propagate = True
-  logger.setLevel(logging.DEBUG)
-  # Just in case, make sure we don't have any extra handlers hanging around.
-  logger.handlers = [RecordHandler(test_record, notify_update)]
+  if DEFAULT_TESTRECORD_LEVEL != 'off':
+    logger.propagate = True
+    logger.setLevel(DEFAULT_TESTRECORD_LEVEL.upper())
```
Kenadia: I'm supposing that the intention with the logging flags is to filter what appears in the console, but not to filter what ends up in the test record? If that's right then I think the level here needs to be `DEBUG` to ensure all logs are emitted to `test_record.log_records`. We can then filter what goes to `stdout` using the top-level `RECORD_LOGGER`. I got the behavior I wanted by setting the level on the handler instead of the logger itself:

```python
record_handler = logging.StreamHandler(stream=sys.stdout)
record_handler.setLevel(DEFAULT_TESTRECORD_LEVEL.upper())
record_logger = logging.getLogger(RECORD_LOGGER)
record_logger.propagate = False
record_logger.addHandler(record_handler)
```

jettisonjoe:
Closing this PR for now since we'll have a new one to replace it sometime soon.

PR #717, opened 2018-02-02
Add a 'config-help' CLI flag.

jettisonjoe:
If enabled, this flag will cause the framework to print all config
keys and their descriptions and exit rather than executing the test.

Example output:
```
  ble_sniffer_download_port
  -------------------------
  Port for HTTP server for downloading sniffer logs.

  ble_sniffer_enable
  ------------------
  (no description found)

  ble_sniffer_host
  ----------------
  Host running BLE sniffer software.

  ble_sniffer_local_path
  ----------------------
  Local path to access BLE Sniffer logs.

  ble_sniffer_port
  ----------------
  Port the BLE sniffer is running on.

  ble_sniffer_usb_hub_channel
  ---------------------------
  If the acroname_usb_hub_serial exists and can be commanded, this config
  key specifies the channel that the BLE Sniffer is connected to.
```

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/717)

### DIFF
--- a/openhtf/util/conf.py
+++ b/openhtf/util/conf.py
@@ -467,7 +467,6 @@ class Configuration(object):  # pylint: disable=too-many-instance-attributes
       result.append('')
     return '\n'.join(result)
 
-
   def save_and_restore(self, _func=None, **config_values):
     """Decorator for saving conf state and restoring it after a function.
 
--- a/openhtf/util/conf.py
+++ b/openhtf/util/conf.py
@@ -446,6 +446,22 @@ class Configuration(object):  # pylint: disable=too-many-instance-attributes
         retval[key] = value
     return retval
 
+  @property
+  def help_text(self):
+    """Return a string with all config keys and their descriptions."""
+    result = []
+    for name in sorted(self._declarations.keys()):
+      result.append(name)
+      result.append('-' * len(name))
+      if self._declarations[name].description:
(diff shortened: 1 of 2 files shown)


openhtf/util/conf.py
```diff
@@ -446,6 +446,22 @@ def _asdict(self):
...
+  def help_text(self):
+    """Return a string with all config keys and their descriptions."""
+    result = []
+    for name in sorted(self._declarations.keys()):
+      result.append(name)
+      result.append('-' * len(name))
+      if self._declarations[name].description:
+        result.append(self._declarations[name].description.strip())
+      else:
+        result.append('(no description found)')
```
Kenadia: I think it's worth adding a line for `default_value` when present

Maybe:
```
plug_teardown_timeout_s
-----------------------
Timeout (in seconds) for each plug tearDown function if > 0; otherwise, will wait an unlimited time.
Default value: 0


station_api_bind_address
------------------------
(no description found)
Default value: '0.0.0.0'
```
(With added quotes to identify string values)

wallacbe:
This is an awesome feature, thanks

openhtf/util/conf.py
grybmadsci: +1

jettisonjoe:

PR #719, opened 2018-02-08
open test run as binary

wallacbe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/719)


fahhem:
I believe this only affects Windows, right? If that's the point, then LGTM

wallacbe:
observed on linux + python3

PR #722, opened 2018-02-09
CLI Output Tools

jettisonjoe:
This PR adds a handful of tools to help facilitate a CLI interface revamp. It's still a WIP, so input/feedback is welcome. In particular, the sys.excepthook override doesn't work as written since nearly all exceptions are caught and logged by the framework. My plan for that is to stash the exception's message string somewhere and have the executor print it last (in red) in the event of a test error so you don't have to scroll back past teardown phase output (we have a spammy teardown) in order to see it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/722)


openhtf/util/console_output.py
```diff
@@ -0,0 +1,253 @@
...
+import logging
+import math
+import sys
+import time
+
+import colorama
+import contextlib2 as contextlib
+
+# Colorama module has to be initialized before use.
+colorama.init()
```
wallacbe: I'm using colorama in custom output module on Linux and OSX and didn't need to do init

openhtf/util/console_output.py
```diff
@@ -0,0 +1,253 @@
...
+  This function is intended as a 'tee' mechanism to enable the CLI interface as
+  a first-class citizen, while ensuring that everything the operator sees also
+  has an analogous logging entry in the test record for later inspection.
+
+  Args:
+    msg: The message to print/log.
+    color: Optional colorama color string to be applied to the message. You can
+        concatenate colorama color strings together in order to get any set of
+        effects you want.
+    file: A file object to which the baracketed text will be written. Intended
```
melodylu: bracketed

(I miss Obama too)

openhtf/util/console_output.py
jettisonjoe: The colorama docs specify that it's a no-op on Linux and Mac but not on Windows, so we should just call it to make sure Windows is supported.

openhtf/util/console_output.py
jettisonjoe: Done. Now excuse me while I quietly sob in the corner for a while.

fahhem: Now Joe's crying! Thanks Obama!

openhtf/util/console_output.py
```diff
@@ -0,0 +1,253 @@
...
+  Example:
+
+    >>> banner_print('Foo Bar Baz')
+    ======================== Foo Bar Baz =======================
+  """
+  lpad = int(math.ceil((width - len(msg) - 2) / 2.0)) * '='
+  rpad = int(math.floor((width - len(msg) - 2) / 2.0)) * '='
+  file.write('{color}{lpad} {msg} {rpad}\n'.format(
+      color=color, lpad=lpad, msg=msg, rpad=rpad))
+  file.write(colorama.Style.RESET_ALL)
```
grybmadsci: probably doesn't matter, but I think it would look a bit better if this was {reset} in the format string above and was written before the \n.  if you had a particular reason for doing it after the \n then it's fine as-is.

openhtf/util/console_output.py
```diff
@@ -0,0 +1,253 @@
...
+    msg: The message to print/log.
+    color: Optional colorama color string to be applied to the message. You can
+        concatenate colorama color strings together in order to get any set of
+        effects you want.
+    file: A file object to which the baracketed text will be written. Intended
+        for use with CLI output file objects like sys.stdout.
+    logger: logger: Logger to which to send analogous output for each message.
+        Intended for use with test record loggers.
+  """
+  file.write('{color}{msg}{reset}'.format(
```
grybmadsci: no \n on this one? also would be good to be consistent on whether the reset is a separate .write() call or baked into the format string - i kind of like it in the format string, myself, but your call.

openhtf/util/console_output.py
jettisonjoe: Done.

openhtf/util/console_output.py
jettisonjoe: Done.

PR #723, opened 2018-02-15
Modify setup.py to build protos on Windows

daiweili:
Also change yaml library to one that builds on Windows. I've test python setup.py install on Windows 10 and Linux. Let me know if there's anything else you'd like me to test. Thanks!


setup.py
```diff
@@ -109,6 +114,9 @@ def run(self):
                   '"protobuf-compiler" and "libprotobuf-dev" packages.')
           elif sys.platform == 'darwin':
             print('On Mac, protobuf is often installed via homebrew.')
+          else:
+            print('On Windows, protoc should be installed and added '
+                  'to the path.')
```
jettisonjoe: Maybe you could leave more of a bread crumb here for how the user should go about getting protoc for windows, even if it's just to point them at the protobuf website. The Linux and Mac strings at least mention which method is recommended to get protoc.

setup.py
```diff
@@ -63,19 +64,23 @@ def initialize_options(self):
...
         self.protoc = maybe_protoc
-    else:
+      else:
         print('Warning: protoc not found at %s' % maybe_protoc)
         print('setup will attempt to run protoc with no prefix.')
-        self.protoc = 'protoc'
+    if not self.protoc:
+      self.protoc = find_executable('protoc')
+      pc_path = os.path.dirname(find_executable('protoc'))
+      self.protodir = os.path.abspath(os.path.dirname(pc_path) + '/../lib')
```
jettisonjoe: Prefer using os.path.join instead of hard-coding path separators.

setup.py
daiweili: Added a link to https://github.com/google/protobuf/releases

setup.py
daiweili: Done

daiweili:
PTAL at the latest change. Sorry, I force pushed and overwrote the first commit (haven't worked with pull requests before).

setup.py
```diff
@@ -63,30 +63,29 @@ def initialize_options(self):
...
+      else:
         print('Warning: protoc not found at %s' % maybe_protoc)
         print('setup will attempt to run protoc with no prefix.')
-        self.protoc = 'protoc'
-
-    self.protodir = os.path.join(prefix, 'include')
+    if not self.protoc:
+      self.protoc = find_executable('protoc')
+      pc_path = os.path.dirname(find_executable('protoc'))
+      self.protodir = os.path.abspath(os.path.join(os.path.dirname(pc_path), '../lib'))
```
jettisonjoe: '../' should probably be os.path.pardir instead.

daiweili: Done.

daiweili:
PTAL

jettisonjoe:
Try re-running that unit test after rebasing on master. I made a change recently that should have fixed its flakiness.

setup.py
```diff
@@ -63,30 +63,29 @@ def initialize_options(self):
...
+      else:
         print('Warning: protoc not found at %s' % maybe_protoc)
         print('setup will attempt to run protoc with no prefix.')
-        self.protoc = 'protoc'
-
-    self.protodir = os.path.join(prefix, 'include')
+    if not self.protoc:
+      self.protoc = find_executable('protoc')
+      pc_path = os.path.dirname(find_executable('protoc'))
+      self.protodir = os.path.abspath(os.path.join(os.path.dirname(pc_path), os.path.pardir, 'lib'))
```
Kenadia: I'm not familiar with protos on Windows, but this looks a bit suspicious, in that you're going up two directory levels and assuming that `lib/` is there.

If you expect `lib/` in a certain place on Windows, consider hardcoding that path. This is similar to what we do for Mac/Linux where we check `/usr/local/include/` on Mac and `/usr/include/` on Linux. If needed, you can get the drive path using `os.path.splitdrive(self.protoc)[0]`.

If you actually want to go up two directories then this should be cleaned up as:

```python
self.protoc = find_executable('protoc')
self.protodir = os.path.join(
    os.path.dirname(os.path.dirname(os.path.dirname(self.protoc))), 'lib')
```

PR #725, opened 2018-02-21
Allow users running the test to specify the log level of test records.

melodylu:
Users can run with --test-record-verbosity=info to have a quieter but
still informative way to watch a test run.

We run a lot of tests from the console that have important information for test fixture operators in our test.logger.info level, but we also hoard a lot of diagnostic information from instruments that we'd rather not show the user on the console. This just adds a quick flag solution so that we don't lose data or overwhelm the user.  Test records still output by default to stdout unless the user intervenes, so hopefully this won't mess up anyone else's use of the stdout.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/725)

### DIFF
--- a/openhtf/util/logs.py
+++ b/openhtf/util/logs.py
@@ -59,8 +59,8 @@ can be additionally logged to a file (with a different level) via the
 all framework log output to stderr.  The --verbosity flag may be used to set
 the log level threshold for framework logs output to stderr.
 
-Test record logs are by default output to stdout at a debug level.  There is
-no way to change this, if you don't like it redirect stdout to /dev/null.
+Test record logs are by default output to stdout at a debug level.  Use the
+--test-record-verbosity flag to select a different level of chatter.
 """
 
 from past.builtins import basestring
@@ -78,6 +78,7 @@ from openhtf.util import functions
 
 
 DEFAULT_LEVEL = 'warning'
+DEFAULT_RECORD_VERBOSITY = 'debug'
 DEFAULT_LOGFILE_LEVEL = 'warning'
 QUIET = False
 LOGFILE = None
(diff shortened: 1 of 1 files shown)


melodylu:
🖊 📃 ✅

melodylu:
@arsharma1 👓

openhtf/util/logs.py
```diff
@@ -88,6 +89,10 @@ def MyPhase(test, helper):
     '--verbosity', default=DEFAULT_LEVEL, choices=LEVEL_CHOICES,
     action=argv.StoreInModule, target='%s.DEFAULT_LEVEL' % __name__,
     help='Console log verbosity level (stderr).')
+ARG_PARSER.add_argument(
+    '--test-record-verbosity', default=DEFAULT_RECORD_VERBOSITY, choices=LEVEL_CHOICES,
+    action=argv.StoreInModule, target='%s.DEFAULT_RECORD_VERBOSITY' % __name__,
+    help='Console log verbosity level for test record (not framework logger) (stderr).')
```
jettisonjoe: The twin parentheticals at the end of the help string might be a bit confusing.

melodylu: Fair point, will edit.

melodylu:
@jettisonjoe Totally; and no worries if you steamroll and end up changing the way logging works. I saw your [PR](https://github.com/google/openhtf/pull/722/files), looks awesome! But while the proposed console.py has a great set of tools in it, I didn't see the change in logs.py that we could really use right now to handle test logs going straight to stdout. Our contract manufacturers monitor running tests at the console (not the web GUI, not yet), and it would really help to be able to show them just enough of what's happening while still keeping our own detailed logs. 

*Also applause to @DanLipsitt 's [PR](https://github.com/google/openhtf/pull/648) for hilarious terseness; we were actually doing something like that as a stopgap while developing our tests, but you're right, in the end we need to actually keep debug level log files for analysis later.

jettisonjoe:
Cool, yeah, I split the logging refactor out into a separate branch from the cli tools. I'll send it for review soon!

PR #726, opened 2018-02-21
Enable specifying output directory

maddychan:
- Enable specifying an output directory during test execution for
  output file callback.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/726)


fahhem:
Why can't it be passed into the constructor already prepended with the directory?

maddychan:
whoops I found a way around it, closing this now

PR #720, opened 2018-02-08
[WIP] Add new web GUI

Kenadia:
## How to try it out

To run an OpenHTF test with the frontend, simply run:

```
python frontend_example.py
```

To run the dashboard, run

```
python -m openhtf.output.servers.dashboard_server
```

If you run this while `python frontend_example.py` is running in another tab, the dashboard should pick up the running test.

You can run the development server (Webpack) as follows:

```
cd openhtf/output/web_gui
npm install
npm run start
```

During development, the app will connect to the dashboard server if it is running on the default port.

--

Todos before merge:
- see all instances of “DO NOT MERGE”
- station server should optionally run automatically with every test
- I've already written some tests, just need to add these

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/720)

PR #727, opened 2018-02-27
Add new web GUI

Kenadia:
See comments on the old PR at #720

---

### How to try it out

To run an OpenHTF test with the frontend, simply run:

```
python frontend_example.py
```

To run the dashboard, run

```
python -m openhtf.output.servers.dashboard_server
```

If you run this while `python frontend_example.py` is running in another tab, the dashboard should pick up the running test.

You can run the development server (Webpack) as follows:

```
cd openhtf/output/web_gui
npm install
npm run start
```

During development, the app will connect to the dashboard server if it is running on the default port.

--

Some todos:
- Station server should optionally run automatically with every test
- I've written some tests, add those
- Implement the history from disk feature. This may be a little tricky since the implementation depends on the format used for test records on disk.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/727)


jettisonjoe:
Frontend and code are looking good overall. My main request is to go through and beef up docstrings throughout.

setup.py
```diff
@@ -187,8 +187,11 @@ def run_tests(self):
     maintainer_email='jethier@google.com',
     packages=find_packages(exclude='examples'),
     package_data={'openhtf': ['output/proto/*.proto',
-                              'output/web_gui/prebuilt/**/*.*',
-                              'output/web_gui/prebuilt/*.*']},
```
jettisonjoe: Previously we explicitly differentiated between the prebuilt frontend and the build in dist. We made frontend developers run a separate command to update the pre-built to ensure it was done intentionally. Just want to make sure the intention is to move away from that workflow.

wallacbe: Will this PR deprecate old front end?

Kenadia:
Rebased with `master`

Kenadia:
Responded to questions on the old PR #720.

In that PR, Brandon asked about the `enable_station_discovery` and `station_api_port` config values. Those were part of station_api.py and are no longer used. In 2c2ec5a I've removed station_api.py and references to it.

setup.py
Kenadia: @jettisonjoe As discussed off GitHub, I recommend that we check in `dist/` instead of `prebuilt/`. The recommended developer workflow is to use the webpack development server to develop off of `src/`, and to update `dist/` with `npm run build` when you're done making changes.

@wallacbe As discussed off GitHub, this will remove the old front end and Station API entirely. The new station server should provide at least as much functionality and be at least as performant as the StationAPI.

PR #732, opened 2018-03-05
minor improvements for DeviceWrappingPlug

kdsudac:
* Making DeviceWrappingPlug compatible with @test.patch_plugs.  We can't use a strict mock spec, because a DeviceWrappingPlug instance proxies calls to getattr to the underlying device.
* Adding option to disable logging_wrapper on subclasses of DeviceWrappingPlug.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/732)

### DIFF
--- a/openhtf/plugs/device_wrapping.py
+++ b/openhtf/plugs/device_wrapping.py
@@ -68,6 +68,9 @@ class DeviceWrappingPlug(openhtf.plugs.BasePlug):
     openhtf.plugs.InvalidPlugError: The _device attribute has the value None
         when attribute access is attempted.
   """
+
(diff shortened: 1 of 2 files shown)


openhtf/plugs/device_wrapping.py
```diff
@@ -79,7 +82,7 @@ def __getattr__(self, attr):
 
     attribute = getattr(self._device, attr)
 
-    if not callable(attribute):
+    if not verbose or not callable(attribute):
```
fahhem: self.verbose

kdsudac: Good catch: https://github.com/google/openhtf/pull/735

PR #737, opened 2018-03-07
Convert futures to six

wallacbe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/737)

### DIFF
--- a/bin/units_from_xls.py
+++ b/bin/units_from_xls.py
@@ -47,6 +47,7 @@ import re
 import sys
 from tempfile import mkstemp
 
+import six
 import xlrd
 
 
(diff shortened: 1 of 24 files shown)


wallacbe:
in unit tests encountering:

 AttributeError: 'NoneType' object has no attribute 'iteritems'

wallacbe:
When running unit test with python 3 six is imported as <class 'module'>, python 2 six (and other modules) import as <type 'NoneType'> .  Unclear yet why.

wallacbe:
PTAL

wallacbe:
PTAL

PR #739, opened 2018-03-20
Additional Python2/3 compatibility fixes.

arsharma1:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/739)

### DIFF
--- a/openhtf/plugs/usb/fastboot_protocol.py
+++ b/openhtf/plugs/usb/fastboot_protocol.py
@@ -17,7 +17,6 @@
 
 import binascii
 import collections
-import io
 import logging
 import os
 import struct
@@ -71,7 +70,7 @@ class FastbootProtocol(object):
     """
     if arg is not None:
       command = '%s:%s' % (command, arg)
-    self._write(io.StringIO(command), len(command))
+    self._write(six.StringIO(command), len(command))
 
   def handle_simple_responses(
       self, timeout_ms=None, info_cb=DEFAULT_MESSAGE_CALLBACK):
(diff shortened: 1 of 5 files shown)


openhtf/plugs/usb/shell_service.py
```diff
@@ -68,7 +68,6 @@
   # output.getvalue() now contains the output of the arecord command.
 """
 
-import io
```
kdsudac: unused?

arsharma1: It was already converted to six.StringIO.

openhtf/plugs/usb/fastboot_protocol.py
```diff
@@ -71,7 +70,7 @@ def send_command(self, command, arg=None):
     """
     if arg is not None:
       command = '%s:%s' % (command, arg)
-    self._write(io.StringIO(command), len(command))
+    self._write(six.StringIO(command), len(command))
```
Kenadia: (I was curious about the difference--so using six.StringIO, the value under Python 3 must be unicode, but under Python 2, it doesn't have to be. Whereas when using io.StringIO, the value must always be unicode.)

arsharma1: (It's also that the StringIO class was moved to the new module io in 3.x and the original modules were removed.)

Kenadia: Right, but since `io` is available in 2.6+ I was just thinking in another use case we might have decided to use `io.StringIO` in hybrid 2/3 code rather than `six.StringIO`, since the behavior is slightly different.

PR #743, opened 2018-03-22
Use six.string_types where there was basestring before

Kenadia:
@kdsudac ran into this since it broke the UserInput plug.

Used diff of #679 as reference. `fastboot_protocol.py` was fixed previously in #739.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/743)


kdsudac:
LGTM

PR #745, opened 2018-03-23
Fix extra exception when plug init fails.

arsharma1:
If there is an exception during plug initialization, the test record
outcome is never set for the console printing code in
_excecute_test_teardown to be called properly.

This required a few other changes:
* In Phase Executor, don't set the stopping event if there is no
current phase running.
* In Test State, make sure _finalize always sets the test record
outcome.
* Add unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/745)

### DIFF
--- a/openhtf/core/test_executor.py
+++ b/openhtf/core/test_executor.py
@@ -150,10 +150,21 @@ class TestExecutor(threads.KillableThread):
       # Full plug initialization happens _after_ the start trigger, as close to
       # test execution as possible, for the best chance of test equipment being
       # in a known-good state at the start of test execution.
-      self.test_state.plug_manager.initialize_plugs()
+      if not self._initialize_plugs():
+        return
 
       # Everything is set, set status and begin test execution.
       self._execute_test_phases(phase_exec)
+
+  def _initialize_plugs(self, plug_types=None):
+    try:
+      self.test_state.plug_manager.initialize_plugs(plug_types=plug_types)
+      return True
+    except Exception:  # pylint: disable=broad-except
+      # Exit early if plug initializaion failed.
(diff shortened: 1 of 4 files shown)


openhtf/core/test_executor.py
```diff
@@ -150,7 +150,13 @@ def _thread_proc(self):
...
       # test execution as possible, for the best chance of test equipment being
       # in a known-good state at the start of test execution.
-      self.test_state.plug_manager.initialize_plugs()
+      try:
+        self.test_state.plug_manager.initialize_plugs()
+      except Exception:  # pylint: disable=broad-except
+        # Exit early if plug initialization failed.
+        self.test_state.finalize_on_failed_plug_initialization(
+            phase_executor.ExceptionInfo(*sys.exc_info()))
+        return
```
Kenadia: Looks good, can we apply the same thing when running the test start?

arsharma1: Done.

openhtf/core/test_state.py
```diff
@@ -380,13 +391,14 @@ def _finalize(self, test_outcome):
...
-      raise BlankDutIdError(
-          'Blank or missing DUT ID, HTF requires a non-blank ID.')
 
     self.test_record.outcome = test_outcome
 
+    # Sanity check to make sure we have a DUT ID by the end of the test.
+    #if not self.test_record.dut_id:
+    #  raise BlankDutIdError(
+    #      'Blank or missing DUT ID, HTF requires a non-blank ID.')
+
```
Kenadia: un-comment

arsharma1: Ah, yes, trying to figure out why unit tests didn't work.

PR #746, opened 2018-03-26
MacAddressLogFilter use record.getMessage() if not string

kdsudac:
record.msg is not necessarily a string and python logging api does support arbitrary objects (as long as they have __str__ implemented) being passed into logging.  Calling record.getMessage() 

https://docs.python.org/2/library/logging.html#logging.LogRecord.getMessage

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/746)

### DIFF
--- a/openhtf/util/logs.py
+++ b/openhtf/util/logs.py
@@ -152,11 +152,14 @@ class MacAddressLogFilter(logging.Filter):
   def filter(self, record):
     if self.MAC_REPLACE_RE.search(record.getMessage()):
       # Update all the things to have no mac address in them
-      record.msg = self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, record.msg)
-      record.args = tuple([
-          self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, str(arg))
-          if isinstance(arg, six.string_types)
-          else arg for arg in record.args])
+      if isinstance(record.msg, six.string_types:
+        record.msg = self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, record.msg)
+        record.args = tuple([
+            self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, str(arg))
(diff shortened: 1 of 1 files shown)


openhtf/util/logs.py
```diff
@@ -152,11 +152,14 @@ def __init__(self):
...
-          if isinstance(arg, six.string_types)
-          else arg for arg in record.args])
+      if isinstance(record.msg, six.string_types):
+        record.msg = self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, record.msg)
+        record.args = tuple([
+            self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, str(arg))
+            if isinstance(arg, six.string_types)
+            else arg for arg in record.args])
+      else:
+        record.msg = self.MAC_REPLACE_RE.sub(self.MAC_REPLACEMENT, record.getMessage())
```
Kenadia: Works under the assumption that record.args is empty if record.msg is not a string type. Strictly speaking not necessarily true, but should be close enough.

kdsudac: record.getMessage() does the substituting of the args if they exist, that's why on line 153 it will catch cases regardless of whether the mac-address is in record.msg or record.args.

Kenadia: It's pretty good, but consider this (somewhat pathological) edge-case that lets a MAC address through:

```
class StringHolder(object):
  def __init__(self, s):
    self.s = s
  def __str__(self):
    return self.s

leaked_mac = '00:14:22:01:23:45'
logging.warn(StringHolder('%%s%s'), leaked_mac)
```

I don't care too much about this since there are pre-existing, easier ways for MAC addresses to sneak through, e.g.:

```
logging.warn('%s', StringHolder(leaked_mac))
```

Kenadia: I guess what's more relevant is I thought this would break cases like

```python
logging.warn(StringHolder('%s'), 'hi')
```

But it looks like that's not the case

PR #747, opened 2018-03-30
Fix Phase Executor stop race condition.

arsharma1:
The Phase Executor has a race condition where a stop() call issued after
execute_phase() checks the _stopping event, but before the internal
functions actually set the current phase thread.  When this occurs, the
phase will start unexpectedly.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/747)

### DIFF
--- a/openhtf/core/phase_executor.py
+++ b/openhtf/core/phase_executor.py
@@ -193,6 +193,9 @@ class PhaseExecutor(object):
 
   def __init__(self, test_state):
     self.test_state = test_state
+    # This lock exists to prevent stop() calls from being ignored if called when
+    # _execute_phase_once is setting up the next phase thread.
+    self._current_phase_thread_lock = threading.Lock()
     self._current_phase_thread = None
     self._stopping = threading.Event()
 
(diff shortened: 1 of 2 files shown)


kdsudac:
Want to make sure I understand existing behavior:  If stop() is called when the self._current_phase_thread is not set the stop is effectively a no-op.  Right?

It looks like you recently moved the self._stopping.set() down a few lines in stop() to below the return.  Was that necessary, can we just move the self._stopping.set() back up?

kdsudac:
Some comments would be really useful.  The code in our core execution is pretty confusing.

I'm not clear on exactly what you are trying to do with self._current_phase_thread_lock.  Best I can tell you are just using it to make sure the self._current_phase_thread isn't tampered with while you are stopping.  Is that right?

If so, we should move the self._stopping.is_set(): return None out from the locking block.

arsharma1:
This is not a noop in the following condition:
  Different Thread:
    ...
    ->  .execute_phase()
         !! Checked ._stopping !!
         $MARK1
         -> ._execute_phase_once
           phase_thread = ...
           phase_thread.start()
           $MARK2
           self._current_phase_thread = phase_thread

Where .stop() is executed in between $MARK1 and $MARK2.

The recent change I made to move self._stopping.set() down was to allow the PhaseExecutor instance to be reused for TestExecutor's .execute_test_teardown.  The current change avoids that issue by calling .reset_stop().

The reason the _stopping check is inside the locking block in _execute_phase_once is to make sure to gate checking to when we aren't about to change the _current_phase_thread.

I agree about it being confusing; I've still got a task to try to refactor this code at some point.

arsharma1:
By adding the locking, my internal testing errors went from 30% to 0%.  I am also confident that the stop logic is less likely to have other race conditions.

PR #750, opened 2018-04-02
standardizing unit test format for copybara

kdsudac:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/750)


test/output/callbacks/callbacks_test.py
```diff
@@ -20,6 +20,7 @@
 
 import sys
 from io import BytesIO, StringIO
+import unittest
```
Kenadia: Would prefer a different way to fix this, but we don't enforce lint rules strictly in openhtf so okay for now

PR #756, opened 2018-04-10
Majorly refactor logging and error handling.

jettisonjoe:
This change re-tools OpenHTF's logging to clean up and simplify it.
Rather than having handlers attached at different levels of the
logging hierarchy, they are all now controllable via the top-level
'openhtf' logger. In addition:

  * CLI logging is turned off by default.
  * CLI verbosity uses the familiar -vvv flag format.
  * CLI logging format is cleaner and more minimal.
  * Framework logs to to all test records rather than a separate file.
  * Full stack traces during test execution are logged but not printed.
  * Errors during execution are printed last and printed in red.
  * Action/status pairs interleave better with logging.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/756)


arsharma1:
@jettisonjoe 
Re: Framework logs to to all test records rather than a separate file.
How does this handle multiple openhtf.Test instances running under the same process?

openhtf/util/logs.py
```diff
@@ -77,54 +78,34 @@ def MyPhase(test, helper):
...
-    '--quiet', action=argv.StoreInModule, target='%s.QUIET' % __name__,
-    proxy=argparse._StoreTrueAction, help="Don't output logs to stderr.")
-ARG_PARSER.add_argument(
-    '--log-file', action=argv.StoreInModule, target='%s.LOGFILE' % __name__,
-    help='Filename to output logs to, if any.')
-ARG_PARSER.add_argument(
-    '--log-file-level', default=DEFAULT_LOGFILE_LEVEL, choices=LEVEL_CHOICES,
-    action=argv.StoreInModule, target='%s.DEFAULT_LOGFILE_LEVEL' % __name__,
-    help='Logging verbosity level for log file output.')
+ARG_PARSER.add_argument('-v', action='append_const', const='v')
```
arsharma1: Removing the argv.StoreInModule action makes it hard to change the defaults from an external module.

openhtf/util/logs.py
```diff
@@ -162,19 +143,39 @@ def filter(self, record):
...
+    self.test_uid = test_uid
+
+  def filter(self, record):
+    # Only filter logs that from individual test instances. Framework logs go
+    # to every test instance's test record to make station debugging easier.
+    if record.name.startswith(RECORD_LOGGER_PREFIX):
+      _, suffix = record.name.split(RECORD_LOGGER_PREFIX)
+      if suffix:
+        test_uid = suffix.split('.')[0]
+        return True if test_uid == self.test_uid else False
```
arsharma1: return test_uid == self.test_uid

openhtf/util/logs.py
arsharma1: Removing the argv.StoreInModule action makes it hard to change the defaults from an external module.

openhtf/util/logs.py
```diff
@@ -199,40 +200,48 @@ def emit(self, record):
...
+      # If nested under a test record, show where the message came from.
+      if len(loggers) > 1:
+        terse_name = '<{type}: {identifier}>'.format(
+            type=loggers[1], # First thing after UUID should be the logger type.
+            identifier=loggers[-1]) # Last thing should uniquely id the logger.
+      else:
+        # Fall back to using the last five characters of the test UUID.
+        terse_name = '<test %s>' % loggers[0][-5:]
+    return '{lvl} {time} {logger} - {msg}'.format(lvl=terse_level,
+                                                     time=terse_time,
```
arsharma1: style nit: additional lines should line up with the open parenthesis.

openhtf/util/logs.py
```diff
@@ -77,54 +78,34 @@ def MyPhase(test, helper):
...
 
 
 def get_record_logger_for(test_uid):
-  return logging.getLogger('.'.join((RECORD_LOGGER, test_uid)))
+  return logging.getLogger('.'.join((RECORD_LOGGER_PREFIX, test_uid)))
 
 
 def initialize_record_logger(test_uid, test_record, notify_update):
+  htf_logger = logging.getLogger(LOGGER_PREFIX)
+  htf_logger.propagate = False
```
arsharma1: Should this also occur when the logger gets the CLI handler attached in configure_cli_logging?

openhtf/util/logs.py
jettisonjoe: Oops! Had adjusted the template string and forgot to fix the lines below.

openhtf/util/logs.py
jettisonjoe: Ah, that's a good idea.

openhtf/util/logs.py
```diff
@@ -199,40 +201,48 @@ def emit(self, record):
...
+    super(CliFormatter, self).format(record)
+    localized_time = datetime.datetime.fromtimestamp(record.created)
+    terse_time = localized_time.strftime(u'%H:%M:%S')
+    terse_level = record.levelname[0]
+    terse_name = record.name.split('.')[-1]
+    if record.name.startswith(RECORD_LOGGER_PREFIX):
+      # The [1:] split used here is to discard the extra leading dot.
+      loggers = record.name.split(RECORD_LOGGER_PREFIX)[1][1:].split('.')
+      # If nested under a test record, show where the message came from.
+      if len(loggers) > 1:
```
grybmadsci: this is making some assumptions about how users have set up their logger hierarchy, might want to explicitly check for 'plug' and 'phase'?

openhtf/util/threads.py
```diff
@@ -60,7 +60,7 @@ def run(self):
       self._thread_proc()
     except Exception:  # pylint: disable=broad-except
       if not self._thread_exception(*sys.exc_info()):
-        _LOG.exception('Thread raised an exception: %s', self.name)
+        _LOG.critical('Thread raised an exception: %s', self.name)
```
grybmadsci: similar to the other exception -> critical, consider not giving up the stack trace.. when you want it there, you *really* want it there.

openhtf/util/logs.py
```diff
@@ -199,40 +201,48 @@ def emit(self, record):
...
-  if not QUIET:
-    console_handler = logging.StreamHandler(stream=sys.stderr)
-    console_handler.setFormatter(formatter)
-    console_handler.setLevel(DEFAULT_LEVEL.upper())
-    console_handler.addFilter(MAC_FILTER)
-    logger.addHandler(console_handler)
+def configure_cli_logging():
+  """Configure OpenHTF to log to the CLI based on verbosity arg."""
+  args, _ = ARG_PARSER.parse_known_args()
+  if not args.v: # The default behavior is not to log anything to the CLI.
```
Kenadia: When I run a test that raises an error (e.g. add `raise RuntimeError` to hello_world.py) I get a stack trace in the console (without -v). Same if I abort before DUT ID is set. Seems unintended?

I'm also surprised that without -v, I don't see in the console whether the test passed or failed (again using hello_world.py). Is that intended?

openhtf/util/logs.py
jettisonjoe: Okay, cleaned this up a bunch using regexes too.

openhtf/util/threads.py
jettisonjoe: Using log.exception doesn't make sense when we're about to re-raise on the next line, and as Arun mentions, ends up resulting in the same stack trace showing up multiple times. The documentation for logging.exception says it should only be used in exception handlers. Since we're re-raising we should leave it up to the configured exception handler whether/how to express the stack trace. For OpenHTF's usage we're overriding this to propagate exception info up via phase result as discussed in the other place this comes up.

openhtf/util/logs.py
jettisonjoe: I must be missing something with stack traces. I'll take another look. The end test result thing I could go either way on. I can add a (colorized!) cli_print of it at the end of the test.

jettisonjoe:
Okay, I think I've addressed all the feedback so far and unit tests are passing. PTAL!

openhtf/core/test_state.py
```diff
@@ -370,14 +370,14 @@ def finalize_normally(self):
       # Otherwise, the test run was successful.
       self._finalize(test_record.Outcome.PASS)
 
-    self.logger.info('Finishing test execution normally with outcome %s.',
+    self.logger.debug('Finishing test execution normally with outcome %s.',
                      self.test_record.outcome.name)
```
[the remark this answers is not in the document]
jettisonjoe: Done.

openhtf/core/test_executor.py
```diff
@@ -51,7 +51,7 @@ class TestExecutor(threads.KillableThread):
   """Encompasses the execution of a single test."""
   daemon = True
 
-  def __init__(self, test_descriptor, execution_uid, test_start,
+  def __init__(self, test_descriptor, execution_uid, test_start, default_dut_id,
```
kdsudac: Is there a reason we aren't just passing in a TestOptions here instead of unpacking all the values to pass them in one at a time?

jettisonjoe: That seems like a good change we should make.

PR #759, opened 2018-04-13
Serial Collection Plug

maddychan:
- Upstreaming serial collection plug

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/759)

### DIFF
--- a/openhtf/plugs/generic/serial_collection.py
+++ b/openhtf/plugs/generic/serial_collection.py
@@ -1,0 +1,87 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""OpenHTF plug for serial port.
+
+Allows for writing out to a serial port.
+"""
+
+from contextlib import contextmanager
+import threading
+import serial
+
+import openhtf
+from openhtf.util import conf
+
+
+conf.declare(
+    'serial_collection_port',
+    description='Port on which to collect serial data.',
+    default_value='/dev/ttyACM0')
+conf.declare(
+    'serial_collection_baud',
+    description='Baud rate for serial data collection.',
+    default_value=115200)
+
+
+class SerialCollectionPlug(openhtf.plugs.BasePlug):
+  """Plug that collects data from a serial port."""
+  SERIAL_EXCEPTIONS = (serial.SerialException,
+                       serial.SerialTimeoutException,
+                       ValueError)
+
+  @conf.inject_positional_args
+  def __init__(self, serial_collection_port, serial_collection_baud):
+    super(SerialCollectionPlug, self).__init__()
+    # Instantiate the port with no name, then add the name, so it won't be
+    # opened until the collection context is entered.
+    self._serial = serial.Serial(port=None,
+                                 baudrate=serial_collection_baud,
+                                 timeout=1)
+    self._serial.port = serial_collection_port
+
+  @contextmanager
+  def save_data(self, dest):
+    """Save serial data to the named destination while in context.
+
+    Spawns a thread that will open the configured serial port, continously
+    poll the port for data, and write the data to the destination file as it is
+    received. If any serial errors are encountered during the lifetime of the
+    polling thread, data collection stops and an error message is logged.
+    Otherwise, data collection stops and the serial port is closed when the
+    context is exited.
+    """
+    collect = True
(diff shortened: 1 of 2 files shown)


openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,87 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""OpenHTF plug for serial port.
+
+Allows for writing out to a serial port.
+"""
+
+from contextlib import contextmanager
```
Kenadia: style: prefer `import contextlib`

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,87 @@
...
+    self._serial = serial.Serial(port=None,
+                                 baudrate=serial_collection_baud,
+                                 timeout=1)
+    self._serial.port = serial_collection_port
+
+  @contextmanager
+  def save_data(self, dest):
+    """Save serial data to the named destination while in context.
+
+    Spawns a thread that will open the configured serial port, continously
```
Kenadia: `continuously`

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,87 @@
...
+      with open(dest, 'w+') as outfile:
+        while collect:
+          data = self._serial.readline().decode()
+          outfile.write(data)
+
+    collection_thread = threading.Thread(target=_poll)
+    collection_thread.daemon = True
+    try:
+      self.logger.debug(
+          'Starting serial data collection on %s.' % self._serial.port)
```
Kenadia: nit - `...collection on port %s.`

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,87 @@
...
+      self._serial.open()
+      collection_thread.start()
+      yield
+    except self.SERIAL_EXCEPTIONS:
+      self.logger.error('Serial port error. Stopping data collection.',
+                        exc_info=True)
```
Kenadia: I'm suspicious of this `except` clause. It will catch certain errors, including any ValueErrors, that occur in the code managed by this context. But it won't catch errors that occur in the thread. Is that the intention? If so, could you give an example of how this plug is used, and what kind of ValueErrors you're trying to catch? Catching all ValueErrors in the managed block doesn't seem right

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+"""OpenHTF plug for serial port.
+
+Allows for writing out to a serial port.
+"""
+
+from contextlib import contextmanager
+import threading
+import serial
```
Kenadia: Consider adding something like

```python
try:
  import serial
except ImportError:
  logging.error('Failed to import pyserial. Please install the `serial_collection_plug` extra, '
                'e.g. via `pip install openhtf[serial_collection_plug]`.')
  raise
```

I notice we do this for the usb plugs.

openhtf/plugs/generic/serial_collection.py
maddychan: Thanks for your input! Upon further inspection we decided to retool save_data to start_collection() and stop_collection(), which doesn't use the context manager anymore.

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,98 @@
...
+
+  Spawns a thread that will open the configured serial port, continuously
+  poll the port for data, and write the data to the destination file as it is
+  received. If any serial errors are encountered during the lifetime of the
+  polling thread, data collection stops and an error message is logged.
+  Otherwise, data collection stops and the serial port is closed when
+  stop_collection() is called.
+  """
+  SERIAL_EXCEPTIONS = (serial.SerialException,
+                       serial.SerialTimeoutException,
```
Kenadia: `serial.SerialTimeoutException` is a subclass of `serial.SerialException` so it's not needed here

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,98 @@
...
+                       ValueError)
```
Kenadia: Could use a comment here explaining what this is intended to catch

openhtf/plugs/generic/serial_collection.py
```diff
@@ -0,0 +1,98 @@
...
+        'Starting serial data collection on port %s.' % self._serial.port)
+    self._serial.open()
+    self._collection_thread.start()
+
+  def stop_collection(self):
+    if (self._collection_thread is None or
+        not self._collection_thread.is_alive()):
+      self.logger.warning(
+          'Data collection was not running, cannot be stopped.')
+      return
```
Kenadia: Just some minor observations here. I wonder what an example usage of this API looks like. It looks like the code using this plug can't tell if the thread is still running, so it can't know whether it's appropriate to call `stop_collection()` or not. Because of this I imagine usage looks like:

```python
try:
  serial_plug.start_collection()
finally:
  serial_plug.stop_collection()
```

This makes me think we should log an error only if `start_collection()` was never called, but not if the thread was started and errored out. (The warning would be sort of redundant with the error.) It also makes me wonder why a context manager is no longer used.

PR #762, opened 2018-04-19
open proto file with read permissions and binary

wallacbe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/762)


wallacbe:
seeing unicode error without this

PR #763, opened 2018-04-20
Add a --quiet flag that suppresses all CLI output.

jettisonjoe:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/763)

### DIFF
--- a/openhtf/util/console_output.py
+++ b/openhtf/util/console_output.py
@@ -26,15 +26,30 @@ import os
 import re
 import string
 import sys
+import textwrap
 import time
 
 import colorama
 import contextlib2 as contextlib
 
+from openhtf.util import argv
+
 # Colorama module has to be initialized before use.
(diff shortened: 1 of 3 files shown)


openhtf/util/console_output.py
```diff
@@ -77,6 +92,8 @@ def banner_print(msg, color='', width=60, file=sys.stdout, logger=_LOG):
     ======================== Foo Bar Baz =======================
 
   """
+  if CLI_QUIET:
+    return
   logger.debug(ANSI_ESC_RE.sub('', msg))
```
Kenadia: The behavior wrt logging seems inconsistent. In most places in this file, we continue to log messages if `CLI_QUIET` is False. Here, in `banner_print`, we don't log if `CLI_QUIET` is False. In `bracket_print` there is no logging at all.

PR #764, opened 2018-04-22
Make Equals.expected, InRange.minimum and InRange.maximum consistent with the value used for valiation

shijunliu-google:
`Equals.expected`, `InRange.minimum` and `InRange.maximum` should return the same value that has been used in validation, even when a customized `type` has been specifed. This helps the third party test recording tool such as [steam_engine](http://google3/third_party/car/hw/testing/output/steam_engine.py) to correctly interpret the validator's expectations.

### DIFF
--- a/openhtf/util/validators.py
+++ b/openhtf/util/validators.py
@@ -118,16 +118,18 @@ class InRange(RangeValidatorBase):
 
   @property
   def minimum(self):
-    return self._minimum
+    converter = self._type or _identity
+    return converter(self._minimum)
 
   @property
   def maximum(self):
-    return self._maximum
+    converter = self._type or _identity
+    return converter(self._maximum)
 
   def with_args(self, **kwargs):
     return type(self)(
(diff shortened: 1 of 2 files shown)


openhtf/util/validators.py
```diff
@@ -118,16 +118,18 @@ def __init__(self, minimum=None, maximum=None, type=None):
 
   @property
   def minimum(self):
-    return self._minimum
+    converter = self._type or _identity
```
jettisonjoe: As elegant as the boolean short-circuiting trick is here, I think it's better to explicitly check for `None` instead of relying on truthiness as its proxy. For example, it's possible (though unlikely, I admit) that we'll get an instance (with a `__call__` override) for type that has `__bool__` and `__nonzero__` overrides.

shijunliu-google: Fixed, PTAL.

shijunliu-google:
PTAL

PR #769, opened 2018-05-01
OpenHTF data module's convert_to_base_types() now actually leaves uni…

kdsudac:
…code types alone like its doc string suggests.

I ran into this issue when trying to reprocess a test run that had unicode characters in a measurement's unit's "suffix" field.

RELNOTES:
PiperOrigin-RevId: 194878170

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/769)


kdsudac:
Looks like this restores history.py we'll have to clean that up.

Kenadia:
Re CLA: the code is authored by Googlers

PR #770, opened 2018-05-02
Handle dimensioned measurement validation errors

arsharma1:
Exceptions in dimensioned measurement validations occur in the Test
Executor thread, which measn they are not caught by the normal flow of
tests.  When those occur, the test will end and will be reported as a
PASS.

This commit changes that.  Now, an exception during a dimensioned
validation measurement validator function will set the phase result to
the raise exception if the phase result is not terminal.  The existing
result will be left in place if it is terminal; in this case, the
validator exception is logged with its stack dump to ensure that there
is still a record.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/770)

### DIFF
--- a/openhtf/core/phase_executor.py
+++ b/openhtf/core/phase_executor.py
@@ -57,6 +57,7 @@ _LOG = logging.getLogger(__name__)
 
 class ExceptionInfo(collections.namedtuple(
     'ExceptionInfo', ['exc_type', 'exc_val', 'exc_tb'])):
+  """Wrap the description of a raised exception and its traceback."""
 
   def _asdict(self):
     return {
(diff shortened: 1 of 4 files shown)


openhtf/core/phase_executor.py
```diff
@@ -256,13 +258,18 @@ def _execute_phase_once(self, phase_desc, is_last_repeat):
...
       if phase_state.result.is_repeat and is_last_repeat:
         _LOG.error('Phase returned REPEAT, exceeding repeat_limit.')
         phase_state.hit_repeat_limit = True
-        result = PhaseExecutionOutcome(openhtf.PhaseResult.STOP)
+        override_result = PhaseExecutionOutcome(openhtf.PhaseResult.STOP)
       self._current_phase_thread = None
 
+    # Refresh the result in case a validation for a partially set measurement
+    # raised an exception.
+    result = phase_state.result
```
fahhem: `result = override_result or phase_state.result`

arsharma1: Done.

PR #777, opened 2018-05-05
Let users control which filters are added to a test.

### LINKED ISSUE #656: logger displays string as <REDACTED>
test.logger.info('FF:FF:FF:FF:FF:FF') is displayed as
FF:FF:FF:<REDACTED>

Any way around this?

jettisonjoe:
Fixes #656

### DIFF
--- a/openhtf/util/logs.py
+++ b/openhtf/util/logs.py
@@ -154,9 +154,6 @@ class MacAddressLogFilter(logging.Filter):
             self.MAC_REPLACEMENT, record.getMessage())
     return True
 
-# We use one shared instance of this, it has no internal state.
-MAC_FILTER = MacAddressLogFilter()
-
 
 class TestUidFilter(logging.Filter):
   """Only allow logs to pass whose logger source matches the given uid."""
@@ -173,6 +170,7 @@ class TestUidFilter(logging.Filter):
 
 class RecordHandler(logging.Handler):
   """A handler to save logs to an HTF TestRecord."""
+  FILTERS = list()  # List of filters to add to every RecordHandler
 
   def __init__(self, test_uid, test_record, notify_update):
     # Record handlers get the DEBUG logging level since we don't want any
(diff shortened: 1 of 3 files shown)


jettisonjoe:
Couple things for reviewers to look at, plx!
  * I stash the list of filter classes on the `RecordHandler` class. It's a little weird? Open to better ideas if anyone has them.
  * Had to remove `@call_once` from `configure_cli_logging()` because it has args now. I added some checking and logged a warning if it looks like logging might already be configured, but this doesn't 100% prevent duplicate logging like it used to.
  * Because `configure_cli_logging()` happens in execute() now stuff that may have logged between `configure()` and `execute()` won't work anymore. In practice I don't know of any use cases affected by this, but what do folks think?

Kenadia:
Sorry for the long overdue review. A couple comments (mostly following up on the concerns you've already mentioned):

1. Without the `call_once` decorator, we trigger the warning `“calling configure_cli_logging more than once”` by simply running a test more than once in a process. I think that's a normal use case so I think this is a problem.

2. Given all the logging changes I've seen recently in OpenHTF and my team's own code, I've come to believe that logging initialization should not be test-dependent.<br>
As you mentioned, we miss logs between `configure()` and `execute()`, but I think the situation is even worse for users who do additional logging configuration. If their configuration has to happen after `execute()` then it probably has to happen in a phase—my team has been there and I'd rather not go back to it. I think it's moving in the wrong direction with regard to #647, where Dan points out that this affects logs from plug initialization.

Kenadia:
Trying to think of others ways to handle this, here's a (hopefully not *too* creative) proof of concept https://github.com/google/openhtf/compare/master...Kenadia:logs-filtering?expand=1

openhtf/util/logs.py
```diff
@@ -246,15 +245,24 @@ def configure_cli_logging():
...
-  cli_handler.addFilter(MAC_FILTER)
+  for filter_class in filters:
+    cli_handler.addFilter(filter_class())
   # Make sure we also suppress CLI logging (in addition to anything that would
   # have been printed through the printing helper functions) if the user has set
   # the --quiet flag in the console_output module.
   cli_handler.addFilter(console_output.CliQuietFilter())
   htf_logger = logging.getLogger(LOGGER_PREFIX)
+  if htf_logger.handlers:
+    _LOG.warning(
```
Kenadia: _LOG is not defined

openhtf/util/logs.py
```diff
@@ -61,6 +61,7 @@ def MyPhase(test, helper):
 import argparse
 import collections
 import datetime
+import functools
```
Kenadia: unused

PR #782, opened 2018-05-14
Fix more KillableThread related race conditions.

arsharma1:
These conditions include:
* Killing a thread before the _thread_proc function runs, which prevents the
  killed state from being handled properly.
* Killing a thread after the _thread_proc function ran, which prevents proper
  cleanup.  This can also cause errors with garbage collection.
* Killing threads while logging occurs can cause deadlocks if the log handler
  lock is left in a bad state.
* Killing threads while in threading.getLogger can cause deadlocks if logging's
  module-level RLock is left in a bad state.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/782)

### DIFF
--- a/openhtf/core/phase_descriptor.py
+++ b/openhtf/core/phase_descriptor.py
@@ -31,6 +31,7 @@ from openhtf.core import test_record
 import openhtf.plugs
 from openhtf.util import data
 from openhtf.util import logs
+from openhtf.util import threads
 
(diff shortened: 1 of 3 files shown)


kdsudac:
Remove core/history.py

arsharma1:
Removed core/history.py and core/station_api.py.

PR #784, opened 2018-05-16
Make small fixes to console_output.py

Kenadia:
- Fix bug where `bracket_print` would print the message twice in the case where the action failed. 
(Compare output of `py -m openhtf.util.console_output` before and after.)
- Allow passing `logger=None` to disable logs
- Minor refactoring and cleanup

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/784)

### DIFF
--- a/openhtf/util/console_output.py
+++ b/openhtf/util/console_output.py
@@ -15,9 +15,14 @@
 
 """Console output utilities for OpenHTF.
 
-This module provides a few conviences to help formatting output for the CLI, as
-well as for forking output to both a logger (for use with test record loggers)
-and to the CLI directly (through sys.stdout).
+This module provides convenience methods to format output for the CLI, along
+with the ability to fork output to both a logger (e.g. a test record logger)
+and to the CLI directly (e.g. sys.stdout).
+
+Under the default configuration, messages printed with these utilities will be
+saved to the test record logs of all running tests. To change this behavior,
+the `logger` parameter should be overridden, e.g. by passing in the test record
+logger for the current test.
 """
 
 import logging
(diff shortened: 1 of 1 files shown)


jettisonjoe:
Oops, I left some comments on #785 that may actually be better applied here.

openhtf/util/console_output.py
```diff
@@ -92,14 +97,16 @@ def banner_print(msg, color='', width=60, file=sys.stdout, logger=_LOG):
...
   """
-  logger.debug(ANSI_ESC_RE.sub('', msg))
-  if not CLI_QUIET:
-    lpad = int(math.ceil((width - _printed_len(msg) - 2) / 2.0)) * '='
-    rpad = int(math.floor((width - _printed_len(msg) - 2) / 2.0)) * '='
-    file.write('{sep}{color}{lpad} {msg} {rpad}{reset}{sep}{sep}'.format(
-        sep=_linesep_for_file(file), color=color, lpad=lpad, msg=msg, rpad=rpad,
-        reset=colorama.Style.RESET_ALL))
-    file.flush()
+  if logger:
```
Kenadia: > If the intention is to use `None` to disable the logging aspect, I would kinda prefer we use explicit `is not None` checks rather than implicit truthiness checks. Someone could conceivably (though weirdly) pass in a legit logger that has a falsey `__bool__` override. (@jettisonjoe)

Although I'm flexible on this, the implicit check feels right given the following:
- The Google style guide suggests: “Use the "implicit" false if at all possible.”
- The Python logging source relies strongly on logger objects being truthy: see [`Logger.getEffectiveLevel()`](https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1562)

> On a related note, effect of passing in `None` should probably be mentioned in the arg docstrings throughout. (@jettisonjoe)

Done.

Kenadia:
Responded to review, PTAL

openhtf/util/console_output.py
jettisonjoe: >   * The Google style guide suggests: “Use the "implicit" false if at all possible.”

That section of the style guide explicitly calls out this case as a caveat to avoid: "Beware of writing `if x:` when you really mean `if x is not None:`"!

>   * The Python logging source relies strongly on logger objects being truthy: see Logger.getEffectiveLevel()

I was previously unaware of this, and I find it convincing. You may proceed. :smiley:

PR #789, opened 2018-05-22
Map TIMEOUT and ABORTED test outcome to their respective status code

clchiou:
openhtf used to mapping both TIMEOUT and ABORTED test outcome to ERROR test run status code. This change maps them to their respective test run status code.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/789)

### DIFF
--- a/test/output/callbacks/callbacks_test.py
+++ b/test/output/callbacks/callbacks_test.py
@@ -27,6 +27,7 @@ import openhtf as htf
 from openhtf.output.callbacks import console_summary
 from openhtf.output.callbacks import json_factory
 from openhtf.output.callbacks import mfg_inspector
+from openhtf.output.proto import test_runs_pb2
 from openhtf import util
 from openhtf.util import test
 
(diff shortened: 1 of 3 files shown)


test/output/callbacks/callbacks_test.py
```diff
@@ -65,6 +66,21 @@ def test_testrun(self, user_mock):
     testrun_output = BytesIO()
     mfg_inspector.OutputToTestRunProto(testrun_output)(record)
 
+    testrun = test_runs_pb2.TestRun()
+    testrun.ParseFromString(testrun_output.getvalue())
+    self.assertEqual(test_runs_pb2.FAIL, testrun.test_status)
```
clchiou: Hmmm... funny thing. In Python 2, test status is FAIL but in Python 3, it's ERROR. I guess this should be a bug, right? (In both Python 2 and 3 we expect to have the same test status.)

fahhem: What's the raised error in Python 3? It might be a 2 vs 3 incompatibility

clchiou: In Python 2, `testrun.test_status` is `FAIL`, but in Python 3, it's `ERROR`.

clchiou: I found why. In `examples.all_the_things.attachments`, the `test_attachment.data` is `bytes`-typed in Python 3, and thus `assert test_attachment.data == 'This is test attachment data.'` failed.

clchiou:
Unit test is fixed. PTAL.

Kenadia:
Fundamentally, a hardware test run can have one of three results:
- Pass: Indicates that the DUT meets spec.
- Fail: Indicates that the DUT does not meet spec.
- Error: Indicates that something went wrong and the pass-worthiness of the DUT is unknown.

This is, in a sense, the “minimum” information that a hardware test must report, and it therefore makes sense to me that there is a field on the output proto that represents this minimum result.

OpenHTF clearly marks aborted and timed-out runs using “outcome details,” so I believe this info is already readily available. The mere fact that ABORTED/TIMEOUT exist in the Status enum (along with eight other unused values) doesn't seem a good enough reason to use them.

For those reasons, I hesitate to approve what otherwise looks like a sensible change. If you're still interested in making this change I'd like to get input from someone who has been on this project since the earlier days. This is a breaking change for systems that consume test runs—I know mfg-inspector.com is affected and I would want to get input from other users in case other systems are affected.

PR #798, opened 2018-06-15
Refactor user_input

Kenadia:
Summary of changes:
1. Add `last_response` attribute to provide a way to retrieve the response when using prompts asynchronously.
2. Only use the hex form of the prompt UUID. Seems there's no benefit to storing the whole UUID object, and the difference between `prompt_id` and `prompt_id.hex` has caused a couple mixups already.
3. Merge `_create_prompt()` into `start_prompt()`.
4. Add unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/798)


test/plugs/user_input_test.py
```diff
@@ -0,0 +1,75 @@
...
+    thread.start()
+    return_value = self.plug.prompt('Test blocking prompt.')
+
+    self.assertIs(self.plug._asdict(), None)
+    self.assertEqual(return_value, 'Mock response.')
+
+  def test_respond_to_non_blocking_prompt(self):
+    prompt_id = self.plug.start_prompt('Test non-blocking prompt.')
+
+    self.assertIsNot(self.plug._asdict(), None)
```
arsharma1: self.assertIsNotNone

test/plugs/user_input_test.py
```diff
@@ -0,0 +1,75 @@
...
+      while not as_dict:
+        time.sleep(0.05)
+        as_dict = self.plug._asdict()
+      self.plug.respond(as_dict['id'], 'Mock response.')
+
+    thread = threading.Thread(target=_respond_to_prompt)
+    thread.start()
+    return_value = self.plug.prompt('Test blocking prompt.')
+
+    self.assertIs(self.plug._asdict(), None)
```
arsharma1: self.assertIsNone

PR #799, opened 2018-06-15
Refactor logs.py

Kenadia:
Summary:
- Address #787
- Rewrite obsolete docstrings.
- Use `logger.getChild()` when possible to access child loggers.
- Don't call `setLevel()` unnecessarily. Loggers and handlers default to the `NOTSET` level. Only “top-level” loggers with `propagate=False` need to have their level set.
- Other minor cleanup. See comments inline.

--

<sub>Note: Originally I was trying to address #788, and did so by defining a second `openthf.output.servers` tree separate from the main `openhtf` logging tree, so that server logs including multicast logs could avoid being sent to the test record. @jettisonjoe pointed out that the server logs are generally more helpful than annoying, as long as they're kept at the debug level. So we need another way of addressing #788—probably by removing that log line entirely. FWIW here's what my original diff looked like: https://github.com/google/openhtf/compare/master...Kenadia:server-logs?expand=1</sub>

Issue #788, opened 2018-05-17
multicast.py spams the test record with logs


Kenadia:
With the recent logging overhaul, framework logs (i.e. logs emitted by loggers descending from `openhtf` except for the “test record loggers”) are sent to all test records.

This includes the `openhtf.util.multicast` debug logs, which by default, are emitted every 3 seconds if there is a dashboard server around. (The dashboard server is new in #727 but the old Station API behavior was the same.)

This clogs the logs and has been observed to cause performance issues in the frontend if too many logs (e.g. 100,000) accumulate.

Kenadia:
### Repro steps:

In a tab run:

> python -m openhtf.output.servers.dashboard_server

In another tab run:

> python -m examples.frontend_example -vv

Result:

![screen shot 2018-06-15 at 10 27 25 am](https://user-images.githubusercontent.com/3301218/41481487-0b69a4c8-7087-11e8-8723-472b1f56089a.png)

The “Received multicast message” logs repeat every three seconds. This causes a problem because these are saved to the test record.

PR #800, opened 2018-06-15
Silence multicast logs by default

Kenadia:
See #788 for repro steps.

**Before:**

screen shot 2018-06-15 at 10 27 25 am

**After:** No multicast log spam.

See also #799.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/800)

### DIFF
--- a/openhtf/util/multicast.py
+++ b/openhtf/util/multicast.py
@@ -29,6 +29,10 @@ from six.moves import queue
 
 _LOG = logging.getLogger(__name__)
 
+# The multicast logs can occur frequently, e.g. every couple seconds. Prevent
+# most logs from ending up in the test record by default.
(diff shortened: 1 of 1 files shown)


openhtf/util/multicast.py
```diff
@@ -29,6 +29,10 @@
 
 _LOG = logging.getLogger(__name__)
 
+# The multicast logs can occur frequently, e.g. every couple seconds. Prevent
+# most logs from ending up in the test record by default.
+_LOG.setLevel(logging.WARNING)
+
```
Kenadia: This also could have been added into the one-time logs setup in `openhtf.util.logs`. I figured it makes sense for each module to govern its own default logging behavior.

Issue #793, opened 2018-06-13
Output callbacks are difficult to debug


Kenadia:
### Repro steps

In `examples/hello_world.py` add the following lines right after `test = htf.Test(hello_world)`:

```python
  def raises():
    raise RuntimeError
  test.add_output_callbacks(raises)
```

### Issue 1

Run the example as normal with `py -m examples.hello_world`. There is no indication that the output callback failed.

In the recent logging changes, we tried to hide tracebacks by default to provide a cleaner UI experience in production. But I question whether serious errors like an output callback failing should be silenced entirely. Shouldn't there be at least a single line to indicate something went wrong?

### Issue 2

Run the example verbosely with `py -m examples.hello_world -vvv`. We learn that an output callback raised, but there is no way to get a traceback. I'd expect there to be some way to get a traceback.

PR #803, opened 2018-06-22
If a callback fails you should print the traceback

kdsudac:
If a callback fails you can continue but should print the traceback, not suppress.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/803)


Kenadia:
Duplicates #793. I suggest that in non-verbose mode, there should be a single line indicating a callback wailed, and in verbose mode, we should get a traceback.

Kenadia:
Addresses #793

PR #804, opened 2018-06-22
Adding uses_base_tear_down functionality and having thread name refle…

kdsudac:
…ct if the tearDown is the BasePlug no-op.  The thread name is logged via KillableThread so this should give test author clear indicator that the tearDown run for that plug is a no-op.

PiperOrigin-RevId: 201715393

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/804)

### DIFF
--- a/openhtf/plugs/__init__.py
+++ b/openhtf/plugs/__init__.py
@@ -198,7 +198,7 @@ class BasePlug(object):
     """Checks whether the tearDown method is the BasePlug implementation."""
     this_tear_down = getattr(self, 'tearDown')
     base_tear_down = getattr(BasePlug, 'tearDown')
-    return this_tear_down.__func__ is base_tear_down.__func__
+    return this_tear_down.__code__ is base_tear_down.__code__
 
 
 class FrontendAwareBasePlug(BasePlug, util.SubscribableStateMixin):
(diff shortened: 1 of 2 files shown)


openhtf/plugs/__init__.py
```diff
@@ -450,8 +458,11 @@ def tear_down_plugs(self):
     """
     _LOG.debug('Tearing down all plugs.')
     for plug_type, plug_instance in six.iteritems(self._plugs_by_type):
-      thread = _PlugTearDownThread(plug_instance,
-                                   name='<PlugTearDownThread: %s>' % plug_type)
+      if plug_instance.uses_base_tear_down():
+        name = '<PlugTearDownThread: BasePlug No-Op for %s>' % plug_type
```
Kenadia: I wonder, if this is really a no-op, why even run the thread?

By the way, could you provide a little info about what motivated this change?

kdsudac: Yeah, I'm on the fence about why even running the thread.  On one hand it's somewhat nice to see in the logs that all the teardowns are running as expected, on the other hand it is needless overhead.

Motivation for change:  It's somewhat easy to implement a tearDown and it not actually be run (e.g. a typo on casing like TearDown) but appear in the logs as if it is run because the KillableThread shows it as being run.  This problem was more acute when I used device_wrapper.DeviceWrappingPlug, because the tearDown is not proxied to the underlying device (although all other methods are).  

I had a change to have the underlying device's tearDown be run if it exists, but Joe felt somewhat strongly that was not the correct solution.  So compromise was to at least give better indicator to test author and/or end user when the tearDown is a no-op.

PR #812, opened 2018-07-13
🚧Changes needed for the dashboard to work with Pyhton 3

adavidzh:
While trying to get up and running with OpenHTF, we got stuck with the dashboard and the station servers not being able to communicate. This was traced back to some formatting issues with the message contents `b''` string literals, which is probably related to differences between Python 2 and 3. This solution is a kludge in the sense that no thought was put into wether it is the best, but since some `.encode`s and `.decode`s were already in the code base, we thought this could not hurt.

⚠️This is foremost a warning to others using OpenHTF under Python 3.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/812)


openhtf/util/multicast.py
```diff
@@ -166,6 +168,7 @@ def _handle_responses():
     while True:
       try:
         data, address = sock.recvfrom(MAX_MESSAGE_BYTES)
+        data = data.decode("utf-8")
```
Kenadia: Single quotes please, for consistency with the codebase style

Kenadia:
LGTM but I think we need a CLA in order to merge this

adavidzh:
Regarding the CLA, I signed it!

PR #816, opened 2018-08-28
adding simple checkpoint to openhtf.util

kdsudac:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/816)

### DIFF
--- a/openhtf/util/checkpoints.py
+++ b/openhtf/util/checkpoints.py
@@ -15,13 +15,17 @@
 
 """A simple utility to check whether all previous phases have passed.
 
-In general test execution stops on a raise Exception but will continue if a
+In general test execution stops on a raised Exception but will continue if a
 phase has a failed measurement.  This is desirable behavior in many cases
 because you want to acquire additional information before the test stops.
 
 However, in some cases it's less desirable because downstream long-running
 phases will be executed only to FAIL at test completion due to the failing
 measurements.  This can have negative implications for tact time.
+
+Checkpoints allow test authors to have their cake and eat it too.  A checkpoint
+checks the result of all previous phases and will stop test execution if a
+previous phase has failed.
 """
 
 from openhtf.core import phase_descriptor
--- a/openhtf/util/checkpoints.py
+++ b/openhtf/util/checkpoints.py
@@ -1,0 +1,46 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


openhtf/util/checkpoints.py
```diff
@@ -0,0 +1,50 @@
...
+
+Checkpoints allow test authors to have their cake and eat it too.  A checkpoint
+checks the result of all previous phases and will stop test execution if a
+previous phase has failed.
+"""
+
+from openhtf.core import phase_descriptor
+from openhtf.core import test_record
+
+def Checkpoint(checkpoint_name=None):
```
fahhem: that's an odd name? I'd expect a Checkpoint to save my progress and let me restart a test from a checkpoint (like in a Mario or Sonic game), which is something I know was previously requested by one of the internal teams.

kdsudac: You have a pretty narrow definition of checkpoint, from a couple video games. :)

https://en.wikipedia.org/wiki/Checkpoint

I'm using it more in the sense of a place where you are stopped and a check is run to verify certain conditions are meet before you are allowed to proceed.

Open to other naming suggestions.

Kenadia: The name makes sense to me but is a little nonspecific and could be annoying if we end up implementing the other type of checkpoint down the line.

Looks like internally we called this `CheckValidOrDie`. Another straightforward option that I like is `StopTestIfFailed` which I think should make things very clear to test operators and others viewing the test results.

fahhem:
I was going to suggest something along the lines of `StopTestIfFailed`

openhtf/util/checkpoints.py
kdsudac: I'm inclined to go with checkpoints.AlreadyFailing().  I have a bit of a pet peeve against function names being too verbose to the point where they are replacing short docstrings.

fahhem: Don't forget that phase lists are normally read directly, and you want to make sure that the phase name makes sense. `AlreadyFailing` doesn't say what the response is to it already failing, is it setting a measurement to True if it's already failing? `StopIfFailing` isn't any longer (I think it's actually shorter) and is clearer in its meaning

kdsudac: Yes, that's why we have a docstring and a logging statement.

I'm also still of opinion that checkpoints as a concept should capture the idea of 'stop' if conditions aren't met.

What I'm trying to avoid is this module turning into a bunch of flow control phases.  e.g. XifY.  I know there are some requests for that, but I'm of the opinion that is less is more.  Start with the simplest case and hope that it captures 95% of the the use cases.  We can re-examine the remaining 5% ad consider whether they should be added to the framework or are better handled in one-off fashion.

Kenadia: On the other hand if this phase is already doing flow control, it might be better to admit it than try to hide it. I'm mostly thinking of operators on the line for whom this could be confusing. They don't have docstrings and in general a single log line passes by very quickly and won't be seen (limitations of the UI design).

Anyway I don't like `AlreadyFailing` but if you prefer `Checkpoint` to `StopIfFailing` I'm fine getting this in and seeing how it goes.

PR #822, opened 2018-10-02
Fix for issue 820

kwmlodozeniec:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/822)


kdsudac:
This doesn't actually work.

kdsudac:
`from openhtf.output.web_gui import web_launcher` is the import you want.

Running simple fixes like this manually is a good habit to get into because our unit test coverage isn't 100%.  In this case just running `python examples/frontend_example.py`.

Thanks for fixing.

kwmlodozeniec:
The import you suggested doesn't work.

kdsudac:
Doesn't work as in you get a traceback on an import error?

or doesn't work as in your don't see the frontend pop-up?

kwmlodozeniec:
Import error.

arsharma1:
This pull is not needed for main, where it has already been fixed.  It's only an issue in the 1.3.0 release, which means we need a new one.

kwmlodozeniec:
So, can this PR be closed?

PR #828, opened 2018-10-17
logging of exception in TestExecutor, import cleanup

kdsudac:
Adding logging of exception inside TestExecutor.
Removing some unused (or should be unused) imports of built-ins and common modules into openhtf module namespace.

PiperOrigin-RevId: 217422914

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/828)


kdsudac:
@arsharma1  Ping, PTAL.

PR #831, opened 2018-10-20
Remove unused imports

sam-bristow:
I've left the internal openhtf imports for now. They can be cleaned up
in a separate patch once I've convinced myself it won't break anything
unexpected.

All tests pass on Python 2.7.15 and 3.6.6

Should be no functional changes as a result of this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/831)

### DIFF
--- a/examples/example_plugs.py
+++ b/examples/example_plugs.py
@@ -14,8 +14,6 @@
 
 """Example plugs for OpenHTF."""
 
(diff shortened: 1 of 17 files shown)


arsharma1:
Thanks for the cleanup!

PR #835, opened 2018-12-04
Python 3 compatibility xrange(10) --> range(10)

cclauss:
__xrange()__ was removed in Python 3 in favor of __range()__.  For unknown values we would __from six.moves import xrange__ but for the low value of __10__, there is no practical difference __range()__ and __xrange()__ performance on Python 2.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/835)


kdsudac:
Thanks for the fix. 

This change looks identical to https://github.com/google/openhtf/pull/843.  Perhaps both were generated using automated tools?

cclauss:
Yes on the identical.  No on the automated changes.

PR #841, opened 2018-12-19
feat: make test execution return code usable in CI

Minipada:
# What I did
In the current software, we get True if tests passed, False otherwise.
With the change we get 0 if it succeeded and the outcome enumeration value if failed. The goal is to get a 0 error code at the end of the test execution when it succeeds, to do

    sys.exit(test.execute(foo))

# Why I did it
This is useful in CI context for example where you need the error code to determine if a job failed. In the current situation, we get True if tests passed (1) and False (0) otherwise, which is not usable.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/841)

### DIFF
--- a/openhtf/core/test_descriptor.py
+++ b/openhtf/core/test_descriptor.py
@@ -351,8 +351,10 @@ class Test(object):
         del self.TEST_INSTANCES[self.uid]
         self._executor = None
 
-    return final_state.test_record.outcome == test_record.Outcome.PASS
-
+    if final_state.test_record.outcome == test_record.Outcome.PASS:
+      return 0
+    else:
+      return final_state.test_record.outcome.value
 
 # TODO(arsharma): Deprecate the teardown_function in favor of PhaseGroups.
 class TestOptions(mutablerecords.Record('TestOptions', [], {


kdsudac:
In the past people have used the return code for control flow from command
line.  e.g.  loop this test until it fails or run these 3 tests in series
as long as previous tests passed.

Minipada:
I could also not change anything and use `sys.exit(!test.execute(foo))` but this is counter intuitive to use non zero value for success. What would you suggest?

kdsudac:
I misunderstood exactly what you were trying to do.  My team internally has some code that wraps OpenHTF that does something similar to what you are doing:

```
success = test.execute(test_start=test_start)
# ... lots of other random stuff
if not success:
  sys.exit(1)
```

Minipada:
That could work indeed. Now that I think of it, I definitely see some use for the outcome values. Still in CI context, it would be interesting to know if a test failed or it an error occurred. Those different outcome could automatically alert different teams.

jettisonjoe:
I agree that being able to differentiate error runs from fail runs via the outcome of `test.execute()` would be really nice, but this would be an API-breaking change and thus we should be careful to push it only in a major version bump rather than a patch.

Minipada:
@jettisonjoe I'd like to help, I don't see where it could break anything though. Could you give some directions?

dbhatman:
> I agree that being able to differentiate error runs from fail runs via the outcome of `test.execute()` would be really nice, but this would be an API-breaking change and thus we should be careful to push it only in a major version bump rather than a patch.

I agree with Joe's sentiment here regarding this being an API breaking change. Maybe we could instead consider returning an object that would act truthy (override __bool__) but also has an attribute to hold the status or even the whole test record could be useful.

Either way I think this PR as is cannot go in so I'll close it out.

Issue #734, opened 2018-03-05
examples/all_the_things.py broken


kdsudac:
examples/all_the_things.py appears to be returning FAIL due to this test phase:

@htf.measures(
    htf.Measurement('replaced_min_only').in_range('{min}', 5, type=int),
    htf.Measurement('replaced_max_only').in_range(0, '{max}', type=int),
    htf.Measurement('replaced_min_max').in_range('{min}', '{max}', type=int),
)
def measures_with_args(test, min, max):
  test.measurements.replaced_min_only = 1
  test.measurements.replaced_max_only = 1
  test.measurements.replaced_min_max = 1

The logging output to stdout/stderr indicates a FAIL, but no indication of what failed.  Not sure if expected behavior is to also include reason for failure.

karlp:
This has been fixed and should be closed.

PR #851, opened 2019-01-14
all_the_things.py PASS by default, console_summary.py handle UNSET

kdsudac:
Make all_the_things.py PASS by default.

Make console_summary do a better job of handling UNSET dimensions (previously did not print out this as reason for test failure).

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/851)


kdsudac:
This addresses: https://github.com/google/openhtf/issues/734

PR #861, opened 2019-03-05
Improve clean up code after running a Test

arsharma1:
Improve clean up code after running a Test by not relying on the `__del__` function being called.  Instead, rename to `cleanup` and call it explicitly when needed.

Also made cleanup more explicit for unit  tests.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/861)

### DIFF
--- a/openhtf/core/test_executor.py
+++ b/openhtf/core/test_executor.py
@@ -68,6 +68,7 @@ class TestExecutor(threads.KillableThread):
     This function is defined instead of a __del__ function because Python calls
     the __del__ function unreliably.
     """
+    self.wait()
     self.test_state.close()
 
   def abort(self):
--- a/openhtf/core/test_executor.py
+++ b/openhtf/core/test_executor.py
@@ -60,6 +60,16 @@ class TestExecutor(threads.KillableThread):
     self._full_abort = threading.Event()
     self._teardown_phases_lock = threading.Lock()
 
+  def close(self):
+    """Close and remove any global registrations.
+
+    Always call this function when finished with this instance.
+
+    This function is defined instead of a __del__ function because Python calls
+    the __del__ function unreliably.
+    """
+    self.test_state.close()
+
   def abort(self):
     """Abort this test."""
     if self._abort.is_set():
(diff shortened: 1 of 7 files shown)


BilalMAddam:
I've been extensively testing your fix. I've ran the company's framework which is using openhtf as the core tester module > 2000 iterations (real DUT and test equipment). Your fix seems to reduce the growth of the memory usage drastically. However, as u said, there is still some minor leak which keeps the memory usage increasing at a small rate.

*** Attached is a snapshot of the process's memory growth at runtime (pid rss comm)
[mmleak_fix_861.log](https://github.com/google/openhtf/files/2930554/mmleak_fix_861.log)

I've been monitoring the memory usage using the following tools:
** htop
** system monitor
** while sleep 1; do ps -ho pid,rss,comm PID; done

It's a great improvement though!
Besides, I'll be running tests over and over and dumping the memory from time to time to identify the small leak that is growing slowly over the time.

openhtf/core/test_executor.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, test_descriptor, execution_uid, test_start, test_options):
     self._full_abort = threading.Event()
     self._teardown_phases_lock = threading.Lock()
 
+  def cleanup(self):
```
kdsudac: nit, docstring describing when/why to use cleanup vs other things like abort, etc.

openhtf/core/test_state.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, test_desc, execution_uid, test_options):
     self.execution_uid = execution_uid
     self.test_options = test_options
 
-  def __del__(self):
+  def cleanup(self):
```
kdsudac: So the idea is instead of deleting the test state, you call cleanup which removes the logs record handler... is this effectively dereferencing the last reference to test state, so it will be garbage collected?

I'd say worth a comment or docstring here.

openhtf/core/test_executor.py
arsharma1: Done.

openhtf/core/test_state.py
arsharma1: Done.

PR #869, opened 2019-04-10
User Input: No console prompt thread for Non-TTY

arsharma1:
Do not run the console prompt thread when stdin is not a TTY.  This will
prevent a thread from being created that almost immediately sleeps for a
week.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/869)

### DIFF
--- a/openhtf/plugs/user_input.py
+++ b/openhtf/plugs/user_input.py
@@ -29,7 +29,6 @@ import platform
 import select
 import sys
 import threading
-import time
 import uuid
 
 from openhtf import PhaseOptions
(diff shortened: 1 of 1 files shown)


openhtf/plugs/user_input.py
```diff
@@ -38,7 +37,7 @@
 from six.moves import input
 
 if platform.system() != 'Windows':
-  import termios
+  import termios  # pylint: disable=g-import-not-at-top
```
cricdecyan: I am not sure this comments belongs in the github source

arsharma1: It's common in protobuf code and has been used in other OpenHTF files.

PR #883, opened 2019-05-21
Add stop_on_failure to PhaseOptions

kunaljdoshi:
If stop_on_failure Phase Option is set to True
PhaseResult is set to STOP instead of CONTINUE
when the PhaseOutcome is FAIL

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/883)

### DIFF
--- a/openhtf/core/test_state.py
+++ b/openhtf/core/test_state.py
@@ -644,6 +644,17 @@ class PhaseState(
                  else test_record.PhaseOutcome.FAIL)
     self.phase_record.outcome = outcome
 
+  def _update_phase_result_from_phase_outcome(self):
+    """Update PhaseResult based on PhaseOutcome.
+
+    This method updates PhaseResult to STOP in case of PhaseOutcome is FAIL.
+    """
+    if self.phase_record.outcome is test_record.PhaseOutcome.FAIL:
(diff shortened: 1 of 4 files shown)


openhtf/core/test_state.py
```diff
@@ -644,6 +644,17 @@ def _set_phase_outcome(self):
                  else test_record.PhaseOutcome.FAIL)
     self.phase_record.outcome = outcome
 
+  def _update_phase_result_from_phase_outcome(self):
```
cricdecyan: How about calling this _update_phase_result_from_failed_phase_outcome()  since it only updates the result for failed phases?

dbhatman:
A stop_on_measurement_failure was added in https://github.com/google/openhtf/pull/1004 that implements a very similar feature, so I am going to close this out.

If you are still interested in this feature, please reopen and update this change with the latest version from master as quite a bit of the code structure has changed.

In addition I think this may break some assumptions around FAIL_AND_CONTINUE because test_record.PhaseOutcome.FAIL is still set then but the user would expect the test to continue. Similarly, the comments say this is specific to measurements but again I think FAIL_AND_CONTINUE may break that logic.

PR #886, opened 2019-06-05
Added docs for building protobuf on windows

Hfusi0n:
Added docs for building protobuf on windows. 
The following documentation should work as a workaround to windows pip installation error.


JamesMTSloan:
This was useful for me when installing with Windows and conda.

PR #905, opened 2019-12-18
Added the ability to show an image with user.prompt. …

Spudmn:
e.g. “user.prompt('Image example',text_input=False,image_url = "http://localhost:5555/lights.jpg")
user.prompt now has the optional parameter image_url.
At the moment the image must be served from a web server.
I don’t know how to get Angular to display images from a file system.
The workaround is to use python SimpleHTTPServer to serve a
folder of images “e.g. python -m SimpleHTTPServer 5555”

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/905)

### DIFF
--- a/openhtf/output/web_gui/src/app/plugs/user-input-plug.component.ts
+++ b/openhtf/output/web_gui/src/app/plugs/user-input-plug.component.ts
@@ -22,7 +22,6 @@ export declare interface UserInputPlugState {
   id: string;
   message: string;
   'text-input': string;
-  'show-image': string;
   'image-url': string;
 }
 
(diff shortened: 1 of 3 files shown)


Spudmn:
@googlebot I fixed it.

openhtf/output/web_gui/src/app/plugs/user-input-plug.component.ts
```diff
@@ -65,6 +67,15 @@ export class UserInputPlugComponent extends BasePlug {
     return this.getPlugState()['text-input'];
   }
 
+  hasImage() {
+    return this.getPlugState()['show-image'];
```
arsharma1: Can you just check if 'image-url' is set rather than adding another parameter?

Spudmn: Fixed in a new commit

PR #926, opened 2020-02-29
Fix validators documentation

tcm0116:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/926)

### DIFF
--- a/openhtf/util/validators.py
+++ b/openhtf/util/validators.py
@@ -5,7 +5,7 @@ method.  They can then be accessed directly as attributes on the validators
 module, and will typically be a type, instances of which are callable:
 
   from openhtf.util import validators
-  from openhtf.util import measurements
+  from openhtf.util.validators import ValidatorBase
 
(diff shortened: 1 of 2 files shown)


tcm0116:
@googlebot I signed it!

openhtf/util/validators.py
```diff
@@ -5,7 +5,7 @@
 module, and will typically be a type, instances of which are callable:
 
   from openhtf.util import validators
-  from openhtf.util import measurements
+  from openhtf.util.validators import ValidatorBase
```
arsharma1: Our style guide does not recommend importing classes and functions from modules.

openhtf/util/validators.py
```diff
@@ -5,7 +5,7 @@
...
 
   class MyLessThanValidator(ValidatorBase):
```
arsharma1: validators.ValidatorBase

PR #931, opened 2020-04-28
Updated Dimension class equality method

telday:
This is useful if you have a list of objects that may not be of type `Dimension` and you need to check if a `Dimension` object is in the list. Previously it would throw an error.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/931)


telday:
@googlebot I signed it!

PR #938, opened 2020-08-19
Fix mfg test record conversions

jdve:
MfgInspector was caching the conversion results.  When using
MfgInspector with output callbacks and the StationServer, this was
causing the saved outputs to always be the results from the first test
run.  Stop caching conversion results to fix this problem.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/938)

### DIFF
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -10,6 +10,8 @@ import zlib
 import httplib2
 import oauth2client.client
 
+from collections import OrderedDict
+
 from openhtf.output import callbacks
 from openhtf.output.proto import guzzle_pb2
 from openhtf.output.proto import test_runs_converter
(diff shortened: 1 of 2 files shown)


arsharma1:
Hi,
Please accept the CLA before I can look at these pull requests.

Thanks!

jdve:
@googlebot I signed it!

jdve:
This approach doesn't work.  The better solution is to simply create a new instance of `MfgInspector` and only use it once:

```python
    def _save_to_disk(test_record):
        inspector = mfg_inspector.MfgInspector()
        inspector.set_converter(mfg_event_converter.mfg_event_from_test_record)
        inspector.save_to_disk('./history/mfg_event_{dut_id}_{start_time_millis}.pb')(test_record)

    test.add_output_callbacks(_save_to_disk)
```

PR #939, opened 2020-08-19
Only send last message if test still running

jdve:
Fix the "Unknown test UID" message that appears in the web gui after any
test has completed. This is caused by the last test record containing
information about a test that has completed. Tests that complete are no
longer in the TEST_INSTANCES list, so the web gui can't retrieve any
more information about it.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/939)


jdve:
@googlebot I signed it!

jdve:
@googlebot I signed it!

PR #940, opened 2020-08-19
Ensure output directory exists

jdve:
If needed, create the directory for the intended output file. This
allows outputs to be grouped into subdirectories instead of only in the
main history directory.

### DIFF
--- a/openhtf/output/callbacks/__init__.py
+++ b/openhtf/output/callbacks/__init__.py
@@ -22,6 +22,7 @@ examples.
 
 import collections
 import contextlib
+import os
 import shutil
 import tempfile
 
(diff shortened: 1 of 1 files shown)


jdve:
@googlebot I signed it!

jdve:
@googlebot I signed it!

openhtf/output/callbacks/__init__.py
```diff
@@ -82,6 +83,10 @@ def serialize_test_record(test_record):
   @staticmethod
   def open_file(filename):
     """Override method to alter file open behavior or file types."""
+    basepath = os.path.dirname(filename)
+
```
arsharma1, line 87: Please remove extra blank lines

openhtf/output/callbacks/__init__.py
```diff
@@ -82,6 +83,10 @@ def serialize_test_record(test_record):
...
+    os.makedirs(basepath, exist_ok=True)
```
arsharma1, line 88: note that `exists_ok` will bring PY2 compatibility, though I plan on doing that shortly anyway.

PR #942, opened 2020-08-19
Fix encoding json with Attachments

jdve:
When using Attachments, we need to properly encode them into json to
share them outside. OutputToJSON handled this correctly, but both
MfgInspector and StationServer did not. Move TestRecordEncoder into
openhtf.util.json, so that it can be shared by all three.  For
station_server, patch the tornado json encoder, so that all values being
sent to the client through tornado are properly encoded.


jdve:
@googlebot I signed it!

jdve:
@googlebot I signed it!

openhtf/output/proto/mfg_event_converter.py
```diff
@@ -23,6 +23,7 @@
 from openhtf.util import data as htf_data
 from openhtf.util import units
 from openhtf.util import validators
+from openhtf.util.json_encoder import TestRecordEncoder
```
arsharma1, line 26: Style: Do not import classes (or functions) from libraries.  Use the library namespace.

from openhtf.util import json_encoder

openhtf/output/proto/mfg_event_converter.py
```diff
@@ -172,7 +173,7 @@ def _convert_object_to_json(obj):
   # Since there will be parts of this that may have unicode, either as
   # measurement or in the logs, we have to be careful and convert everything
   # to unicode, merge, then encode to UTF-8 to put it into the proto.
-  json_encoder = json.JSONEncoder(sort_keys=True, indent=2, ensure_ascii=False)
+  json_encoder = TestRecordEncoder(sort_keys=True, indent=2, ensure_ascii=False)
```
arsharma1, line 176: Use the library namespace, though that will require using a different variable for the encoder.

openhtf/output/servers/station_server.py
```diff
@@ -29,6 +29,7 @@
 from openhtf.util import logs
 from openhtf.util import multicast
 from openhtf.util import timeouts
+from openhtf.util.json_encoder import TestRecordEncoder
```
arsharma1, line 32: Same here.

openhtf/util/json_encoder.py
```diff
@@ -0,0 +1,28 @@
...
+
+import base64
+import json
+
+from openhtf.core import test_record
+
+
+class TestRecordEncoder(json.JSONEncoder):
+  """JSON encoder that supports Attachments."""
+  def default(self, obj):
```
arsharma1, line 23: Add a blank line after the class docstring.

jdve: Done.

jdve:
Thanks @arsharma1 for the feedback!  I've updated the PR as your requested.

cclauss:
git conflicts

PR #943, opened 2020-09-01
Partially modernize build system

pganssle-google:
This adds a `pyproject.toml`, which allows you to declare the (python) requirements for your build system.

This allows us to drop at least one compatibility anti-feature: the conditional dependency on `enum34` (which won't work properly with wheels).

I also fixed an issue in `find_packages` where you were shipping a top-level package called `pylint_plugins`, which I assume was unintentional. See [the commit message](https://github.com/google/openhtf/commit/082c552b8f8b0ddd0f07cc805e95639969347f64) for more details.

I've dropped the maximum version pin on `pandas` in the `examples` extra, because `python_requires` should handle that, but I didn't drop it in `tests_require`, because you seem to be using the deprecated `setup.py test` mechanism for executing your tests, with a `tests_require` — I believe this may install via the deprecated and long-unmaintained `easy_install` mechanism in `setuptools`, which I think doesn't support `python_requires`.

I think the ideal situation would be to drop all the custom `setup.py` commands and do a direct `pytest` invocation in `tox.ini`. I also strongly recommend dropping all references to `-e .` and `usedevelop=True` and instead [test the package as installed](https://blog.ganssle.io/articles/2019/08/test-as-installed.html) — either by switching to the `src/` layout or using one of the `cd` tricks.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/943)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=40.8.0", "wheel>=0.29.0"]
+build-backend = "setuptools.build_meta"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,2 @@
-[wheel]
+[bdist_wheel]
 universal = 1
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,7 @@ setup(
     author_email='madsci@google.com',
     maintainer='Joe Ethier',
     maintainer_email='jethier@google.com',
-    packages=find_packages(exclude='examples'),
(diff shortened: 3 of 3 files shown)


wallacbe:
Looks like this change prevents wheels from being built and installed properly
    packages=find_packages(include=['openhtf.*']),

$ python -c "from openhtf import conf"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'conf' from 'openhtf' (unknown location)

This works:
packages=find_packages(),

PR #950, opened 2020-09-11
update collections iterable import

wallacbe:
@kdsudac @arsharma1

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/950)

### DIFF
--- a/openhtf/output/callbacks/__init__.py
+++ b/openhtf/output/callbacks/__init__.py
@@ -20,7 +20,6 @@ alternative serialization schemes, see json_factory.py and mfg_inspector.py for
 examples.
 """
 
-import collections
 import contextlib
 import shutil
 import tempfile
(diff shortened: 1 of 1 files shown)


openhtf/output/callbacks/__init__.py
```diff
@@ -21,6 +21,12 @@
 """
 
 import collections
+
+try:
+  collectionsAbc = collections.abc
```
arsharma1, line 26: Please use the six import for now.

from six.moves import collections_abc

...
collections_abc.Iterable

wallacbe: Made update, PTAL

wallacbe: Updated import order

arsharma1 (approved):

PR #981, opened 2021-04-23
Add PEP-561 py.typed file

lalten:
This commit adds a py.typed file, which marks the package as having type information available.
For more information, see  https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages


cclauss:
git conflicts.

PR #1005, opened 2022-05-24
Added support for Python version 3.9.

dieppedalle:
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1005)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -216,9 +216,8 @@ setup(
     tests_require=[
         'absl-py>=0.10.0',
         'mock>=2.0.0',
-        # Remove max version here after we drop Python 2 support.
-        'pandas>=0.22.0,<0.25.0',
-        'numpy<1.17.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -216,9 +216,8 @@ def run_tests(self):
     tests_require=[
         'absl-py>=0.10.0',
         'mock>=2.0.0',
```
glados-verma, line 218: We can remove this now - this is the backported library for python2, right?

dieppedalle: We could remove "mock>=2.0.0" in a future commit. We still have dependencies on "import mock" (https://github.com/google/openhtf/search?q=%22import+mock%22) instead of "from unittest import mock". Let's make a subsequent commit to remove PY2 support (i.e. remove all six dependencies).

glados-verma (approved):
Looks good, let's fix that in follow-ups.

PR #1009, opened 2022-05-26
Removed support for PY2.

dieppedalle:
# What I did
I'm removing all the third party libraries used to add compatibility for PY2.

# Why I did it
To reduce tech debt as PY2 is deprecated (https://www.python.org/doc/sunset-python-2/).

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1009)

### DIFF
--- a/openhtf/core/diagnoses_lib.py
+++ b/openhtf/core/diagnoses_lib.py
@@ -327,7 +327,7 @@ class _BaseDiagnoser(object):
     pass
 
 
-class BasePhaseDiagnoser(_BaseDiagnoser, metaclass=abc.ABCMeta):
+class BasePhaseDiagnoser(_BaseDiagnoser, abc.ABC):
   """Base class for using an object to define a Phase diagnoser."""
 
   __slots__ = ()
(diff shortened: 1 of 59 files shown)


openhtf/core/diagnoses_lib.py
```diff
@@ -204,8 +203,8 @@ def _convert_result(self,
       return
     elif isinstance(diagnosis_or_diagnoses, Diagnosis):
       yield self._verify_and_fix_diagnosis(diagnosis_or_diagnoses, diagnoser)
-    elif (isinstance(diagnosis_or_diagnoses, six.string_types) or
-          not isinstance(diagnosis_or_diagnoses, collections_abc.Iterable)):
+    elif (isinstance(diagnosis_or_diagnoses, str) or
+          not isinstance(diagnosis_or_diagnoses, collections.abc.Iterable)):
```
glados-verma, line 207: Consider importing this as 
from collections.abc import Iterable

...the same way as you'd import CamelCased types from the typing module.

dieppedalle: Done.

glados-verma (changes requested):

glados-verma (approved):

PR #1011, opened 2022-05-27
Fix regressions occuring during PY2 deprecation.

dieppedalle:
# What I did
I fixed a few regressions that occurred because of #1009.

# Why I did it
* `encode()` cannot be called on a set of bytes so we check the type before converting to bytes.
* `dummy_thread` is removed in PY3.9: https://bugs.python.org/issue37312.

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1011)

### DIFF
--- a/openhtf/output/callbacks/mfg_inspector.py
+++ b/openhtf/output/callbacks/mfg_inspector.py
@@ -148,8 +148,8 @@ class MfgInspector(object):
     if user and keydata:
       self.credentials = oauth2client.client.SignedJwtAssertionCredentials(
           service_account_name=self.user,
-          private_key=self.keydata.encode()
-          if isinstance(self.keydata, str) else self.keydata,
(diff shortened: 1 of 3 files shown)


openhtf/output/callbacks/mfg_inspector.py
```diff
@@ -148,7 +148,8 @@ def __init__(self,
     if user and keydata:
       self.credentials = oauth2client.client.SignedJwtAssertionCredentials(
           service_account_name=self.user,
-          private_key=self.keydata.encode(),
+          private_key=self.keydata.encode()
```
glados-verma, line 151: Surround with parentheses and reformat so that the statement starts on the same column across lines.

glados-verma (approved):

glados-verma (approved):

PR #1016, opened 2022-07-29
Fix `reportPrivateImportUsage` errors

lalten:
Fixes `reportPrivateImportUsage` errors in pyright (which is used in Pylance) by switching `__init__.py` exports to a "from p import x as x" pattern

See also https://github.com/Textualize/rich/pull/1596

---
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1016)

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -16,47 +16,47 @@
 import signal
 import typing
 
-from openhtf import plugs
-from openhtf.core import phase_executor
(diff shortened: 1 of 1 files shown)


dbhatman:
Hi, can you check if https://github.com/google/openhtf/pull/1046 resolves this the errors for you instead?

lalten:
Superseded by #1046

PR #1018, opened 2022-08-15
Update dependencies and drop support for Python 3.6 and Python 3.7

jonasclaes:
This PR contains updated dependency constraints. It also drops support for Python 3.6 and Python 3.7, due to these changes. I think it is about time that we drop these versions since they are quite old already.
We're talking 4 years old for the release date of Python 3.7.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1018)

PR #1027, opened 2022-09-09
Drop Python 3.6 support and update some dependencies

jonasclaes:
This PR is similar to #1018.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1027)


setup.py
```diff
@@ -135,12 +135,12 @@ def run(self):
     'dataclasses;python_version<"3.7"',
```
lalten, line 135: Should this also be removed?

jonasclaes: Yes that seems only logical. Looks like I overlooked that one.

jonasclaes:
Looks like I need to fix the mfg_inspector OAuth client.

jonasclaes:
Can we try running the CI again? It should work now.

lalten (approved):

jonasclaes:
LGTM!

dbhatman:
https://github.com/google/openhtf/pull/1028 indicates your change would break the web gui due to tornado api changes.

jonasclaes:
I have just reverted the upgrade of Tornado. Can we do another CI run?

dbhatman:
I *think* the newest tornado version is required for python3.10 (https://github.com/tornadoweb/tornado/pull/2946). But sadly CI isn't catching it because of a lack of tests? We may want to wait to add py3.10 unfortunately :(

jonasclaes:
Oh, hmm. Let's remove Python 3.10 for now then, I'll get started behind the scenes to convert this to be compatible with the new Tornado version.

cricdecyan (approved):

PR #1028, opened 2022-09-09
Loosen dependencies to allow newer packages

nehalkpatel:
Many dependencies are very outdated and when openhtf is combined with or used with other projects, version conflicts become very challenging to unwind.

NOTE: tornado is left with a '<5.0' dependency as there have been some major API changes that require more extensive refactors to accommodate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1028)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,16 +130,16 @@ build.sub_commands.insert(0, ('build_proto', None))
 
 INSTALL_REQUIRES = [
     'attrs>=19.3.0',
-    'colorama>=0.3.9,<1.0',
-    'contextlib2>=0.5.1,<1.0',
-    'google-auth>=1.34.0',
+    'colorama>=0.3.9',
+    'contextlib2>=0.5.1',
     'inflection',
-    'mutablerecords>=0.4.1,<2.0',
+    'google-auth>=1.34.0',
+    'mutablerecords>=0.4.1',
     'protobuf>=3.6.0,<4.0',
     'PyYAML>=3.13',
-    'pyOpenSSL>=17.1.0,<23',
+    'pyOpenSSL>=17.1.0',
     'requests>=2.27.1',
-    'sockjs-tornado>=1.0.3,<2.0',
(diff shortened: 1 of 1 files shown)


dbhatman:
This change breaks in CI due to the oauth2client version update. I suggest we focus our efforts on https://github.com/google/openhtf/pull/1027 and try to get that in first as it bumps a few of these packages (though it does need to resolve its tornado upgrade issue).

nehalkpatel:
yeah, I have made similar fixes locally. Let me find a patch to share.

nehalkpatel:
Locally, to use the newer oath2client, I had to make changes like this:
```
- import oauth2client
+ from oauth2client import service_account

# Code that reads from json file and assigns to self.user, self.keydata

-    self.credentials = oauth2client.client.SignedJwtAssertionCredentials(
-        service_account_name=self.user, private_key=self.keydata, scope=SCOPE_CODE_URI)
+    self.credentials = service_account.ServiceAccountCredentials.from_json_keyfile_dict(
+        data, scopes=[SCOPE_CODE_URI])
```

jonasclaes:
> 

This has just been performed. Once tests pass it can be merged and then this PR should be based on the new Python versions.

dbhatman:
Hi! Can you update your branch with the latest changes?

cricdecyan:
We are in the middle of replacing oauth2client with google.oauth2, so I think it would be best to punt this PR for a few days.

nehalkpatel:
> We are in the middle of replacing oauth2client with google.oauth2, so I think it would be best to punt this PR for a few days.

That's perfectly fine; my main interest is removing the `<x.y` restrictions on package dependencies, as it makes it _really_ hard to use openhtf in the same python environment with other tools we may be building/using

cricdecyan:
@nehalkpatel We are done with our oauth2 migration, if you rebase the PR I can review it.

cricdecyan (approved):

PR #1030, opened 2022-09-14
Changes deprecated oauth2client for google-auth.

tan01:
Oauth2client was previously used in the mfg-inspector output callback for authentication, but the library is now deprecated and replaced by the google-auth library.

This removes oauth2client and httplib2 from openhtf's output callbacks and replaces them with google-auth's oauth2 implementation using the Requests transport.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1030)


dbhatman:
@tan01 looks like the test is failing: https://github.com/google/openhtf/actions/runs/3055466480/jobs/4928650854

PR #1035, opened 2022-09-16
Fix web_gui build

lalten:
The web interface is currently broken on master and won't build.

This PR:
* Fixes the missing `SafeHtml` type
* Reverts the (partial?) update to Angular 10
* Reverts a change in a spec file to be compatible with Angular 4 again
* Replaces optimization in webpack with Terser (was: built-in UglifyJS)
* Adds a new build of the web_gui (this commit could be dropped)

Fixes https://github.com/google/openhtf/issues/1039

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1035)

PR #1036, opened 2022-09-16
Json stringify `object` measurements

lalten:
This change will e.g. make the values of `typing.NamedTuple`s visible in the Web UI. Before this change they were displayed as `[object Object]`. I believe the measurements are guaranteed to be json-serializable because they are coming in via the websocket transport, which uses json already.

Note that the web interface is currently broken on master. You need to apply this change to an older commit to be able to actually build the web_gui.


dbhatman:
Hey you mentioned an older commit, what commit are you using to build and test this out? We can consider rolling the web gui backwards to that point if its really necessary.

I hesitate to take this and https://github.com/google/openhtf/pull/1035 until the changes (as in the web_gui with and without the changes) can be built at current master.

lalten:
That would have been 341aa8610e94899fc651d7c781772487a0c504e8 I believe

lalten:
@glados-verma @dbhatman Now that the web gui is buildable again it would be great if you could consider this one as well.

glados-verma (dismissed):

PR #1042, opened 2022-09-29
Correct TypeVar usage for Validators

dbhatman:
The TypeVar usage was indicating that ALL validators available in _VALIDATORS would have the same type. This update only indicates that these validators are an instance of the ValidatorBase.

Pytype is currently erroring on master and this resolves those errors.

### DIFF
--- a/openhtf/util/validators.py
+++ b/openhtf/util/validators.py
@@ -70,8 +70,7 @@ class ValidatorBase(abc.ABC):
     """Should validate value, returning a boolean result."""
 
 
-_ValidatorT = TypeVar("_ValidatorT", bound=ValidatorBase)
-_ValidatorFactoryT = Union[Type[_ValidatorT], Callable[..., _ValidatorT]]
+_ValidatorFactoryT = Union[Callable[..., ValidatorBase]]
 _VALIDATORS: Dict[str, _ValidatorFactoryT] = {}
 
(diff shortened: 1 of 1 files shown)


cricdecyan (approved):

openhtf/util/validators.py
```diff
@@ -70,8 +70,7 @@ def __call__(self, value) -> bool:
     """Should validate value, returning a boolean result."""
 
 
-_ValidatorT = TypeVar("_ValidatorT", bound=ValidatorBase)
-_ValidatorFactoryT = Union[Type[_ValidatorT], Callable[..., _ValidatorT]]
+_ValidatorFactoryT = Union[Callable[..., ValidatorBase]]
```
lalten, line 73: Doesn't this change also mean that e.g. the `Equals` class (the class itself, not instances of it) can not be used anymore in _VALIDATORS?
I think at least Mypy does not agree that a class type doubles as a `Callable[..., Classtype]` even if you could argue that the class `__init__` makes it behave that way.

dbhatman: Mypy does agree that Type[C] is a Callable[..., C]: https://github.com/python/mypy/issues/1670

Also interestingly enough ABCs also match: https://github.com/python/mypy/issues/12361

PR #1046, opened 2022-10-04
Expose public API with __all__ and minor formatting tweaks

dbhatman:
This replaces the need for https://github.com/google/openhtf/pull/1016

PiperOrigin-RevId: 478889967

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1046)

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -19,38 +19,17 @@ import typing
 from openhtf import plugs
 from openhtf.core import phase_executor
 from openhtf.core import test_record
-from openhtf.core.base_plugs import BasePlug
-from openhtf.core.diagnoses_lib import DiagnosesStore
-from openhtf.core.diagnoses_lib import Diagnosis
-from openhtf.core.diagnoses_lib import DiagnosisComponent
-from openhtf.core.diagnoses_lib import DiagPriority
-from openhtf.core.diagnoses_lib import DiagResultEnum
-from openhtf.core.diagnoses_lib import PhaseDiagnoser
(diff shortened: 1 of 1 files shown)


lalten:
Thanks for taking a stab at this @dbhatman. This does indeed fix the issue with Pylance described in #1016, but it adds another problem: Go-to-definition (in VS Code) will now always jump to `__init__.py` instead of the actual definition. The solution in #1016 does not show this behavior, it jumps right into the actual source file.

dbhatman:
> Thanks for taking a stab at this @dbhatman. This does indeed fix the issue with Pylance described in #1016, but it adds another problem: Go-to-definition (in VS Code) will now always jump to `__init__.py` instead of the actual definition. The solution in #1016 does not show this behavior, it jumps right into the actual source file.

Thanks for following up @lalten 

The behavior (ie requires the second click of go-to) you are seeing is pretty typical in projects following google style guides because of this style rule: https://google.github.io/styleguide/pyguide.html#22-imports. As part of cleaning up some things on our side so we can more closely align with the style guide, I am going to suggest we go forward with this but I'll keep this issue in mind and see if I can come up with an alternate solution that still complies.

lalten:
Lgtm then. If you're in the process of cleaning up it would be great to give the pylintrc a refresh as that's currently broken (references deleted plugins and such).

dbhatman:
Ah was just looking at the pylintrc and scratching my head at the plugins! 

(and cleaning that up broke this PR 😆)

glados-verma (approved):

PR #1049, opened 2022-10-08
Fix and add CI checks for `autoflake`, `isort`

lalten:
[autoflake](https://github.com/PyCQA/autoflake) cleans up unused imports and useless pass statements.
[isort](https://github.com/PyCQA/isort) manages import sorting.

I configured the profiles that would have the smallest diff and seemed most reasonable.

The really interesting part (IMO) is the addition of these as a GitHub Actions CI job that should prevent non-conforming code from being added and hence keep a more consistent code style in the repo.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1049)

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -49,6 +49,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - run: |
-        pip install -U autoflake
-        autoflake .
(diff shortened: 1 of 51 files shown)


glados-verma:
Thanks for the PR, I agree these cleanups are important. I am enabling internal tooling to look for these and additional lint violations, and closing this PR. Also #1198 cleans up existing unused imports and I'll make another PR to fix the import ordering.

PR #1057, opened 2022-10-12
Use bundled sockjs

lalten:
Currently OpenHTF loads SockJS from a CDN over public internet. This is an issue when running in a production environment where public internet connection is not available or desired.

This PR bundles the currently used version of SockJS so the CDN won't be hit anymore.

Probably it would be better to describe this dependency via npm, maybe that could be done in a follow-up PR.


Teslafly:
Is https://github.com/google/openhtf/pull/1164 a potentially better solution for this?

PR #1063, opened 2022-10-14
Branch build_and_deploy to continuous_integration.yml and remove deployment action (to be added in separate workflow)

glados-verma:
Create continuous integration using GitHub starter workflow. Copy over current build step without checking if we still need all dependencies (swig?). Add web_ui jobs.Skip PyPI publish, that will be a separate workflow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1063)


glados-verma:
Not sure why the coverage went down slightly.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,68 @@
...
+name: Continuous Integration
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
```
dbhatman, line 27: maybe build_and_test?

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,68 @@
+# Copyright 2022 Google LLC
```
dbhatman, line 1: Can you fork this file from the build and deploy yml? Most of the contents are the same right?

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Continuous Integration
+
+on:
+  push:
+    branches: [ "master" ]
```
dbhatman, line 22: nice, good call!

dbhatman (changes requested):
Should we do the deletion in this change instead? If not we will run 2x the number of stuff in the mean time (not really a big deal)

.github/workflows/continuous_integration.yml
dbhatman: Actually this one is kinda weird, because it wont rerun these on pushes to pull-request branches!

glados-verma:
> Should we do the deletion in this change instead? If not we will run 2x the number of stuff in the mean time (not really a big deal)

I disabled that workflow manually from the Actions menu; once the corresponding file is deleted I imagine it will just "go away".

.github/workflows/continuous_integration.yml
glados-verma: Done.

.github/workflows/continuous_integration.yml
glados-verma: Done.

.github/workflows/continuous_integration.yml
glados-verma: Reverted this to how it used to be. If we run into quota issues (unlikely since non-private repos should get unlimited minutes, but the rules might have changed/might change) we can reevaluate.

dbhatman (approved):

dbhatman:
I suspect you need to update Settings to use build_and_test instead of build for "required" checks?

glados-verma:
> 

Thanks for the pointer, done.

PR #1093, opened 2023-04-08
Python3.10 compatibility

BananaLoaf:
When trying to run WebUI in Python3.10, I get an error `AttributeError: module 'collections' has no attribute 'MutableMapping'`. It happens because `collections.MutableMapping` got moved to `collections.abc.MutableMapping` in Python3.10. 

Ideally, I'd have to bump tornado package, but I haven't had any success doing so because of the new async policies, that tornado>=5 started following. I could not make async loop to run correctly and even if I did, I'd not be sure that something else has not broken, hence this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1093)

### DIFF
--- a/openhtf/__init__.py
+++ b/openhtf/__init__.py
@@ -14,6 +14,7 @@
 """The main OpenHTF entry point."""
 
 import signal
+import sys
 import typing
 
 from openhtf import plugs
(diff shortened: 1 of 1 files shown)


Teslafly:
Would this go well with this sockjs removal PR?
https://github.com/google/openhtf/pull/1164/files

I would also like to use a newer version of python. If only to get switch-case in python 3.10.

glados-verma:
Should no longer be needed after #1181

PR #1131, opened 2024-01-04
Use coveralls github action instead of manual coveralls

dbhatman:
Coveralls is failing on recent runs so lets migrate to the current latest way to use it.

Example failure: https://github.com/google/openhtf/actions/runs/7403069556/job/20142167751

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1131)


mitchgu (approved):
LGTM

mitchgu (approved):

PR #1157, opened 2024-06-18
Add brackets to ImageURL in user-input-plug.component.html

dieser-niko:
The paramenter `image_url` in `openhtf.plugs.user_input.UserInput.promt()` doesn't seem to have any effect. After adding these brackets, the given image shows up just fine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1157)


dieser-niko:
I don't want to sign in with a google account, so closing again.

PR #1173, opened 2024-10-04
Fix continuous integration (again)

glados-verma:
Update Python versions tested against, and stop using deprecated setuptools test feature.

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1173)


tan01 (approved):
Think the broken integration checks are unrelated.

PR #1185, opened 2024-11-01
Continuous Integration: python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]

cclauss:
Test on current versions of Python.
* https://devguide.python.org/versions

- - -
This change is [Reviewable](https://reviewable.io/reviews/google/openhtf/1185)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v4
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,8 @@ python =
     3.9: py39
     3.10: py310
     3.11: py311
+    3.12: py312
+    3.13: py313
 
 [testenv]
 deps =


glados-verma:
A little bit of work is need to support 3.12 and newer -  I made partial [progress here](https://github.com/google/openhtf/compare/master...glados-verma:openhtf:py312). I think the only thing still needing work is making the built-in test profiling work correctly - for some reason the generated profile files seem to be empty.

glados-verma:
> ```python
>       profile_tempfile = tempfile.NamedTemporaryFile(delete=False)
> ```
> 
> The new optional parameter in Python 3.12 is called `delete_on_close` which is different than `delete`.
> 
> * https://docs.python.org/3/whatsnew/3.12.html#tempfile

Yes, but: `If delete is false, the value of delete_on_close is ignored` from [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)

.github/workflows/continuous_integration.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.11", "3.13"]
```
glados-verma, line 29: Please add 3.10 and 3.12 here as well.

glados-verma:
I'm looking into the profiling breakage under Python3.12+. The error message might be misleading.

cclauss:
https://docs.python.org/3/whatsnew/3.12.html#distutils

glados-verma:
The distutils related issue is already fixed in #1183 . The problem right now is that the profiling unit test fails. I have found [the cause for the breakage](https://github.com/python/cpython/issues/110770). It looks like the profiling implementation in OpenHTF isn't quite right, so I'm looking into that now.

.github/workflows/continuous_integration.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         python -m build
     - name: Test with tox
       run: |
-        python -m pip install tox tox-gh-actions
+        python -m pip install tox tox-gh-actions setuptools
```
glados-verma, line 48: setuptools installation is no longer needed after #1183 , you can revert this line.

tox.ini
```diff
@@ -14,6 +16,7 @@ deps =
     numpy
     pytest>=2.9.2
     pytest-cov>=2.2.1
+    setuptools
```
glados-verma, line 19: Also not needed

glados-verma (changes requested):

.github/workflows/continuous_integration.yml
cclauss:
```suggestion
        python -m pip install tox tox-gh-actions
```

tox.ini
cclauss:
```suggestion
```

glados-verma (approved):

glados-verma:
Thanks for the PR!

Issue #884, opened 2019-05-21
default timeout in prompts causing OverFlowError on windows


wallacbe:
with self._cond:
      if self._prompt:
        if timeout_s is None:
          self._cond.wait(3600 * 24 * 365)

arsharma1:
Check out threading.TIMEOUT_MAX on the platform.  The wait may be too long.

PR #1260, opened 2025-11-11
Use the max timeout for unlimited timeout

fahhem:
On Windows, `threading.TIMEOUT_MAX` is around 24 days (signed 32-bit int of milliseconds), while on Linux it's around 292 years (signed 64-bit int of nanoseconds).

This also fixes https://github.com/google/openhtf/issues/884 and the discord post https://discord.com/channels/1077920440569299074/1418256119658184795


glados-verma (approved):

glados-verma:
I'm curious, what is that Discord group?

PR #1288, opened 2026-05-22
Add custom unit fields to test_run.proto

Sayter99:
This change is to enable generated test_run to support custom units.


openhtf/output/proto/test_runs.proto
```diff
@@ -144,6 +144,10 @@ message TestParameter {
   // Created for visualization by UIs that don't support certain fancy
   // parameters. UIs that do support them should hide these parameters.
   optional string associated_attachment = 21;
+
+  // If using a unit not in the Units.UnitCode enum:
+  optional string custom_unit_code = 22;
+  optional string custom_unit_suffix = 23;
```
glados-verma: The next tag is 28, per comment below. Can you "reserve" these two tags? They should have been reserved in this commit: https://github.com/google/openhtf/commit/c497890b861cb9cb8a53b53962118b3921f47945

Please use tag 28 and 29 and increment the next tag comment

Sayter99: thanks for pointing this out, done

glados-verma (approved):

PR #1292, opened 2026-06-11
Fix error when capture_source is enabled and test_start is None

### LINKED ISSUE #1279: Error when `capture_source` is enabled and `test_start` is `None`
Hi, ran into and issue when enabling `capture_source`.

https://github.com/google/openhtf/blob/618b91420ed896a691d9ab0b2fd40b2de60c4012/openhtf/core/test_descriptor.py#L326-L327

When `capture_source` is enabled and no `test_start` is given, it tries to get the `func` attribute on a `None` value.

harrish243:
What does this PR do?
This PR fixes an AttributeError that occurs when the capture_source configuration is enabled but no test_start trigger is provided.

Why is this change needed?
In openhtf/core/test_descriptor.py, if test_start is passed as None, the trigger variable also evaluates to None. When CONF.capture_source is enabled, the code previously attempted to unconditionally access trigger.func, resulting in an error because it tried to get the func attribute on a None value.

How does this PR fix the problem?
I added a simple safety check (if CONF.capture_source and trigger:) so that it only attempts to capture the source code info if a valid trigger function actually exists.

Fixes #1279

### DIFF
--- a/openhtf/core/test_descriptor.py
+++ b/openhtf/core/test_descriptor.py
@@ -323,7 +323,7 @@ class Test(object):
       else:
         trigger = test_start
 
-      if CONF.capture_source:
+      if CONF.capture_source and trigger:
         trigger.code_info = htf_test_record.CodeInfo.for_function(trigger.func)
 
       self._executor = test_executor.TestExecutor(


glados-verma:
Looks good, but can you resolve the merge conflicts please?

glados-verma:
Superseded by #1307

PR #1294, opened 2026-06-24
Fix poor test isolation in plugs_test.py (issue #753)

### LINKED ISSUE #753: plugs_test.py poorly isolated (relies upon socket)
plugs_test.py is poorly isolated.  It relies upon opening a socket.

I tried simply mocking out xmlrpcutil.SimpleThreadedXmlRpcServer which is used via the test plug which is a front end aware plug.  However, still faced failures because I believe plug manager expects to be able to communicate with the front end aware plug over the socket.

harrish243:
Fixes #753

plugs_test.py had several class-level state leaks that made tests order-dependent and relied (in older versions) on a real socket being opened by FrontendAwareBasePlug via xmlrpcutil.SimpleThreadedXmlRpcServer.

Changes:
- Add module-level docstring explaining the hermetic design contract.
- Reset AdderPlug.LAST_INSTANCE in setUp() so each test starts with a clean slate (previously only INSTANCE_COUNT was reset, causing the LAST_INSTANCE reference from one test to bleed into the next).
- Reset TearDownRaisesPlug1.TORN_DOWN and TearDownRaisesPlug2.TORN_DOWN in setUp() so test_tear_down_raises cannot influence other tests via shared class-level boolean state.
- Add class-level and method-level docstrings that document why state is reset and confirm that no network socket is opened.
- Add a docstring to test_plug_updates() making it explicit that the update-notification path uses only in-process threading.Event objects (via SubscribableStateMixin) and no socket is involved.

All 9 existing tests continue to pass.

### DIFF
--- a/test/plugs/plugs_test.py
+++ b/test/plugs/plugs_test.py
@@ -11,6 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Tests for openhtf.plugs.
+
+This test module is designed to be fully hermetic: it must not open any
+network sockets or rely on any external services.  Specifically, even though
+FrontendAwareBasePlug is tested here, the PlugManager is exercised without
+starting the StationServer (or any other server), so no XML-RPC or WebSocket
+connection is ever established.  All plug-update signalling relies only on
+in-process threading.Event objects exposed by SubscribableStateMixin.
+
+See https://github.com/google/openhtf/issues/753.
+"""
 
 import threading
 import time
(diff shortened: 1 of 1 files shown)


test/plugs/plugs_test.py
```diff
@@ -11,6 +11,17 @@
...
+"""Tests for openhtf.plugs.
+
+This test module is designed to be fully hermetic: it must not open any
+network sockets or rely on any external services.  Specifically, even though
+FrontendAwareBasePlug is tested here, the PlugManager is exercised without
+starting the StationServer (or any other server), so no XML-RPC or WebSocket
+connection is ever established.  All plug-update signalling relies only on
+in-process threading.Event objects exposed by SubscribableStateMixin.
+
+See https://github.com/google/openhtf/issues/753.
```
glados-verma, line 23: Can remove the tracker link

test/plugs/plugs_test.py
```diff
@@ -157,6 +196,12 @@ def test_tear_down_raises(self):
     self.assertTrue(TearDownRaisesPlug2.TORN_DOWN)
 
   def test_plug_updates(self):
+    """Test update-notification path of FrontendAwareBasePlug.
+
+    This exercises SubscribableStateMixin.asdict_with_event() /
+    notify_update() entirely in-process via threading.Event objects.
+    No network socket is opened.
```
glados-verma, line 203: Condense this into a one-line docstring.

glados-verma (changes requested):
Thanks for the PR, looks good overall. Couple of minor changes requested.

PR #1314, opened 2026-07-25
fix: give alternate unit suffixes their own keys

### LINKED ISSUE #1250: PERCENT double declaration
## Issue
In openhtf.units L3648C1-L3651C26, PERCENT is defined [twice](
https://github.com/google/openhtf/blob/22ef3a04774c157ea1c2ba5712e83e71e33d266e/openhtf/util/units.py#L3648C1-L3651C26):

PERCENT = UnitDescriptor('percent', 'P1', '%')
ALL_UNITS.append(PERCENT)
PERCENT = UnitDescriptor('percent', 'P1', 'pct')
ALL_UNITS.append(PERCENT)

Because the second assignment overwrites the first, units.PERCENT always resolves to "pct", and the "%" unit is inaccessible.

## Solution
Rename PERCENT "pct" --> PERCENT_PCT

Sanjays2402:
Fixes #1250

When a UNECE symbol column lists aliases (`% or pct`), the generator emitted both descriptors under the *same* Python key, so the second overwrote the first — `units.PERCENT` resolved to `pct` and the `%` descriptor was unreachable as a module attribute. `KILOGRAM_PER_LITRE`, `DECITONNE` and `RACK_UNIT` had the same collision.

`units_from_xls.py` now keeps the canonical key for the first suffix and derives a distinct key for each alternate (`PERCENT_PCT`, etc.); `units.py` is regenerated to match. Lookup via `units.Unit('%')` / `units.Unit('pct')` is unchanged. New `test/util/units_test.py` fails on master and passes with the fix.


Sanjays2402:
@googlebot I signed it!

Sanjays2402:
@googlebot I signed it!

Sanjays2402:
I've signed the Google Individual Contributor License Agreement as an individual contributor, and the cla/google check is now passing.

This should be ready for review whenever a maintainer has a moment. Happy to rebase or adjust anything if needed. Thanks!

glados-verma:
There are a few problems with this approach, including:

1. The alias names have "jitter" e.g. `KILOGRAM_PER_LITRE_KG_PER_L`
2. The approach doesn't care for which one should be considered the "primary" alias

Given the small number of units that have aliases, we might as well create a lookup such units.

Sanjays2402:
Both fair, and the jitter one was the bigger problem — `KILOGRAM_PER_LITRE_KG_PER_L` is not a name anyone would type.

Reworked in b81fdd5 along the lines you suggested. The generator now keeps the canonical key for the *first* symbol listed in the sheet (that's the primary one) and appends alternates to `ALL_UNITS` anonymously, so they stay reachable via `units.Unit('kg/L')` / `Unit('dtn')` / `Unit('RU')` but get no module-level name. A small explicit `ALIAS_UNIT_KEYS` table names the one alias I think is genuinely worth exposing, `PERCENT_PCT`, since `pct` is common in test configs — it's a two-line dict so adding another later is trivial.

Net effect on the generated module: `PERCENT` is `%` again, `KILOGRAM_PER_LITRE` is `kg/l`, `DECITONNE` is `dt`, `RACK_UNIT` is `U`, and none of the four are overwritten by their alias. Diff to units.py is now 4 hunks instead of the previous key-renaming spread.

Tests updated to match (primary keeps its symbol / alternates resolve by lookup only). 4 pass here; they fail 3/4 against current master.

Happy to drop `PERCENT_PCT` too and make the rule purely "only the primary symbol gets a name" if you'd rather have zero exceptions.

